PR #32, opened 2022-06-21
Feature: Exclusion tables

joeflack4:
```
commit be60334abcc55cbee3308d34b1e87cbf8fcacf95
    Feature: Exclusion tables
    - Update: TODOs
    - Update: Utilizing `base_prefix_map` now instead of `prefix_map` from ontology metadata `.yml` files.
    - Update: get-terms-children.sparql.jinja2: Always using rdfs:subClassOf* instead of a + at the end. Updated Python code to reflect.
    - Update: Makefile target: Now explicitly passes params again.
    - Added: New exclusion files: ORDO, DO. Still pending NCIT after resolving Java heap space issue.
    - Update: Robot template output: 'AI rdfs:seeAlso' --> 'AI MONDOREL:has_exclusion_reason'
```

```
commit 6df216a9c319194f4e4faaf571d77fce8ecdfa2b
    Feature: Exclusion tables
    - Added new command: 'python-install-dependencies'
    - Updated makefile commands to use new dependency 'python-install-dependencies'
    General
    - Some changes to makefile and a docs file related to running 'update_repo'
```

```
commit 4ffa8a00b2eaf53f760fa36a310a0ced62c4a5af
    Feature: Exclusion tables
    - Optionalized params: Now any path params will get populated using the required --onto-name param should they be absent.
    - CodeStyle: PEP8 compliance update
    - Script location: Simplified
    - Ontology config .yml updates: Added missing prefix_map items
    - Script now utilizes config.yml files for each ontology in order to filter out relevant prefixes.
    - doid_relevant_signature.sparql: added missing prefix
    - added warning for excluded terms that appear in relevant_signature file
    - Now includes non-inclusions, utilizing mirror_signature and component_signat
ure.
```

```
commit 3f9edcb97816215b0e7924345442d09ccad21af0
    Feature: Exclusion tables
    - Add: ORDO exclusions
    - Update: refactor (i) comments, (ii) TODOs
    - Update: SPARQL: (i) removed the returning of term labels, (ii) utilized rdfs:label instead of skos:prefLabel, (iii) removed unnecessary prefixes.
    - Update: Used tmp/ dir for cache.

    Feature: ORDO Mapping annotations
    - Update: Use tmp/ dir for cache.
```

```
commit a22d82f60f8975760c517e72ce5b67032cb75192
    Feature: Exclusion tables
    - Update: Make goal: Fixed .owl file dependency using 'relevant signature' file.
    - Update: ICD10CM exclusion table TSV: Replaced arbitrarily expanded list back to original intensional list given by curators.
    - Update: Script now uses --relevant-signature param
    - Update: Script now reports raises an error when conflict in `exclude_children` between parent and child classes.
    - Bugfix: Trimmed whitespace programmatically from input exclusion file.
    - Bugfix: Trimmed whitespace manually in input exclusion file.
    - Comments/todo's: (i) added a todo for possible edge case issue w/ exclude_children
    - Update: Added param `--onto-name` to script. This is good for future proofing. Previously, I was
extracting the ontology name from one of the file names that was passed as a param. However, as our
 pipeline changes, those file names are also subject to change. As such, it is more stable to pass
the ontology name explicitly.
```

```
commit ae1c3950f909a8134cde331bda5694a44dc6929d
    Feature: Exclusion tables
    - Updates to existing exclusion query tables: Added header, including missing columns, but did not yet populate data in those columns.
    - Added new exclusion query tables for the rest of the ontologies.
    - Added exclusion results for ICD10CM.
    - Added Python script to work w/ these tables: term_range_expansion.py
    - Added SPARQL Jinja2 scripts to be called by Python script.
    - Added make goal to call Python script.
    - requirements: Updated Python requirements files to include OAK.
```


src/ontology/mondo-ingest.Makefile
```diff
@@ -160,15 +160,17 @@ mappings: sssom $(ALL_MAPPINGS)
...
-
+# Documentation for `report-mapping-annotations` and `update-jinja-sparql-queries`: `docs/developer/ordo.md`
 report-mapping-annotations:
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/report_mapping_annotations.py
 
 update-jinja-sparql-queries:
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/create_sparql__ordo_replace_annotation_based_mappings.py
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/create_sparql__ordo_mapping_annotations_violation.py
 
+config/%_term_exclusions.txt: config/%_term_exclusions.tsv mirror/%.owl
```
joeflack4, line 171: I'm assuming this is the correct place in the makefile to add this make goal.

### Issues
Make goal seems to fail because of absence of these goals:
1. `config/%_term_exclusions.tsv`
2. `mirror/%.owl`

src/ontology/mondo-ingest.Makefile
```diff
@@ -160,15 +160,17 @@ mappings: sssom $(ALL_MAPPINGS)
...
+	python3 ../scripts/term_range_expansion.py --exclusions-path $(word 1,$^) --onto-path $(word 2,$^)
```
joeflack4, line 172: I changed the `$^` syntax for more explicit named parameterization.

src/scripts/term_range_expansion.py
```diff
@@ -0,0 +1,207 @@
...
+sparql_child_exclusion_path = os.path.join(PROJECT_DIR, 'sparql', 'get-terms.sparql.jinja2')
+input_exclusion_term_ranges_csv_path = os.path.join(
+    PROJECT_DIR, 'preparation', 'icd10cm_excluded_term_ranges_intensional.csv')
+input_icd10cm_ttl_path = os.path.join(PROJECT_DIR, 'preparation', 'icd10cm.ttl')
+TERM_RANGE_MAX_VAL = 99
+
+
+# Functions
+# todo: pd.errors.EmptyDataError: I think this happens because Python process needs to stop before results saved?
+def sparql_jinja2_file_query__via_robot(
```
joeflack4, line 41: I have copy/pasted and repurposed this function many times in several repositories. I'm guessing that it would be good to standardize this and move it into some library. Maybe OAK?

src/scripts/term_range_expansion.py
```diff
@@ -0,0 +1,207 @@
...
+    try:
+        df = pd.read_csv(results_path).fillna('')
+    except pd.errors.EmptyDataError as err:
+        os.remove(results_path)  # done so that it doesn't read this from cache
+        raise err
+
+    return df
+
+
+def run(onto_path: str, exclusions_path: str) -> pd.DataFrame:
```
joeflack4, line 87: This is still a work in progress. I'm repurposing this from the previous ICD10CM work.

src/scripts/term_range_expansion.py
```diff
@@ -0,0 +1,207 @@
...
+
+def cli():
+    """Command line interface."""
+    parser = get_parser()
+    kwargs = parser.parse_args()
+    kwargs_dict: Dict = vars(kwargs)
+
+
+    # TODO: temporary hack; need to get .owl versions of these eventually
+    onto_file_remappings = {
```
joeflack4, line 195: Temporary hack. I have / have been constructing these as RDF `.ttl`. Maybe it's OK to keep using that? But we probably want to standardize everything to the same file extension. So maybe there is some way to get these into `.owl` isntead?

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
```diff
@@ -0,0 +1,272 @@
...
+    #  Should match all labels then end in the word phenotype and then exclude all their children.
+    #  First here, I should filter original df to have only regex rows
+    pass
+
+    # TODO: JOIN non-regex and regex results
+    pass
+
+    # todo: save
+    #  2. a simple two column file with the term_id and the exclusion reason
+    #    - Fields: `Term ID`, `Exclusion Reason` || `term_id`, `exclusion_reason`
```
joeflack4, line 193: @matentzn You wrote in the GH issue to name these columns `Term ID` and `Exclusion Reason`, but I'm wondering if you'd rather leave them as `term_id` and `exclusion_reason` for consistency / ease of downstream processing?

src/ontology/mondo-ingest.Makefile
matentzn: YES! Awesome work I have not looked at everything yet but super excited so far!

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
matentzn: I think keep as `term_id` for downstream processing related reasons.. Thank you!

src/sparql/get-terms-children.sparql.jinja2
```diff
@@ -0,0 +1,9 @@
+prefix ICD10CM: <http://purl.bioontology.org/ontology/ICD10CM/>
+prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+prefix skos: <http://www.w3.org/2004/02/skos/core#>
+
+select ?term_id ?child_id ?child_label where {
+  VALUES ?term_id { {{values}} }
+  ?child_id rdfs:subClassOf{{property_path_qualifier}} ?term_id ;
```
matentzn, line 7: lol smart. Chris woul want to see us using oak here moving forward. For the next PR, can we try to use oak rather than sparql? No need now to waste hours. Thanks!

src/sparql/get-terms.sparql.jinja2
```diff
@@ -0,0 +1,10 @@
+prefix ICD10CM: <http://purl.bioontology.org/ontology/ICD10CM/>
+prefix owl:  <http://www.w3.org/2002/07/owl#>
+prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+prefix skos: <http://www.w3.org/2004/02/skos/core#>
+
+select ?term_id ?term_label where {
```
matentzn, line 6: Why is this needed, if you already supply you with the term tables?

joeflack4: Ah, good point, thanks Nico. I probably would have realized this when I got further on w/ my script, but now that you bring it up, I realize that this particular query is not necessary at all. Just need to the other query to get children. Don't need this query, as I already have `term_id` and `term_label` for the parent classes excluded.

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
```diff
@@ -0,0 +1,277 @@
...
+    #         new_df = new_df.drop([col + '_x', col + '_y'], axis=1)
+    #
+    #     return new_df
+
+    # df['term_info_source'] = ''
+    # results_df_excluded_kids['term_info_source'] = 'icd10cm.ttl'
+    # results_df_included_kids['term_info_source'] = 'icd10cm.ttl'
+    # # Determine which factors were not in source
+    term_in_icd10cm_ttl = set(results_df['term_id'])
+    df['term_in_source_icd10cm.ttl'] = df['term_id'].apply(
```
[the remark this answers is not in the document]
joeflack4: Yep, it is / will be. I still haven't finished repurposing the script from previous ICD work yet.

src/ontology/config/icd10cm_exclusions.tsv
```diff
@@ -0,0 +1,725 @@
+term_id	term_label	exclusion_reason	exclude_children
```
joeflack4, line 1: There are actually some problems with this file. What I have here had to be massaged from our previous work that we had done and put in Google Sheets. But there were actually a number of problems with those results, namely that the input list of intensional terms had some problems. I'll bring up what those were in the next QC meeting.

For now, I created a GitHub issue about the issues:
https://github.com/monarch-initiative/mondo-analysis/issues/38

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
```diff
@@ -0,0 +1,299 @@
+"""Takes in the full ontology and the exclusions tsv and extracts a simple list of terms from it.
+Outputs two files:
+1. a simple list of terms (config/ONTO_NAME_term_exclusions.txt)
+2. a simple two column file with the term_id and the exclusion reason
+
+# Resources
+- GitHub issue: https://github.com/monarch-initiative/mondo-ingest/issues/22
+
+TODO's
+  - todo: Do we need to worry about edge case where exclusion table is empty? (e.g. we include such a table for all
```
joeflack4, line 10: `todo: Do we need to worry about edge case where exclusion table is empty? (e.g. we include such a table for all ontologies, but there exists 1 where we don't exclude anything.`
I can add this edge-case handling if you want.

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
```diff
@@ -0,0 +1,299 @@
...
+     ontologies, but there exists 1 where we don't exclude anything.
+  - todo: later: see below: #x1: regex results: would be good to filter out by IRI or prefix that is not from source
```
joeflack4, line 12: `todo: later: see below: #x1: regex results: would be good to filter out by IRI or prefix that is not from source ontology, but how to determine the appropriate prefix just based on the 2 file params, but could have edge cases.`
Basically, it is possible that the REGEX matches might return terms that are in appear in one of our target ontologies, but are xrefs to another ontology. So I'm wondering if we might need to write some logic (and possibly pass an explicit param, e.g. the prefix to look for) to handle those edge cases.

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
```diff
@@ -0,0 +1,299 @@
...
+     ontology, but how to determine the appropriate prefix just based on the 2 file params, but could have edge cases.
+  - todo: later: see below: #x2
+  - todo: later: see below: #x3
+  - todo: later: see below: #x4: Switch from SPARQL to OAK.
+  - todo: later: When constructing exclusion TSV: bug: ICD10CM:C7A-C7A reported as term, but should be ICD10CM:C7A
```
joeflack4, line 17: This is actually a bug I'll need to address in the ICD10CM expansion work done in `mondo-analysis`: https://github.com/monarch-initiative/mondo-analysis/issues/37

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
```diff
@@ -0,0 +1,299 @@
...
+    except pd.errors.EmptyDataError:
+        # remove: so that it doesn't read this from cache, though it could be that there were really no results.
+        os.remove(results_path)
+        # could also do `except pd.errors.EmptyDataError as err`, `raise err` or give option for this as a param to func
+        df = pd.DataFrame()  # could also return None
+
+    return df
+
+
+def expand_ontological_exclusions(
```
joeflack4, line 129: FYI: Since the REGEX-based queries and the explicit term queries share a lot of logic, I abstracted a lot of this logic into this function.

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
```diff
@@ -0,0 +1,299 @@
...
+    onto_path: str, exclusions_df: pd.DataFrame, cache_suffix: str = None,
+    export_fields=['term_id', 'exclusion_reason'], uri_fields=['term_id', 'child_id']
+) -> pd.DataFrame:
+    """Using an ontology file and an exclusions file, query and get return an extensional list of exclusions.
+    cache_suffix: Used by and explained in sparql_jinja2_robot_query()."""
+    # Variable names: kids0=Children excluded; kids1=Children included
+    df_kids1 = exclusions_df[exclusions_df['exclude_children'] == False]
+    df_kids0 = exclusions_df[exclusions_df['exclude_children'] == True]
+    terms_kids1: List[str] = list(set(df_kids1['term_id']))  # QA: set() removes any duplicates
+    df_kids1_results: pd.DataFrame = sparql_jinja2_robot_query(
```
joeflack4, line 139: FYI: For cases where `exclude_children=FALSE`, we need to query. But when `TRUE`, don't need to query. So I split rows, query for the ones where `FALSE`, and then I join the results together at the end.

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
```diff
@@ -0,0 +1,299 @@
...
+    exclusion_fields = ['exclusion_reason', 'exclude_children']
+    sep = '\t' if exclusions_path.endswith('.tsv') else ',' if exclusions_path.endswith('.csv') else ''
+    onto_name = os.path.basename(onto_path).split('.')[0]
+    df_results_path = os.path.join(CONFIG_DIR, f'{onto_name}_exclusion_reasons.tsv')
+    df_results_simple_path = os.path.join(CONFIG_DIR, f'{onto_name}_term_exclusions.txt')
+
+    # Prepare data
+    df = pd.read_csv(exclusions_path, sep=sep).fillna('')
+    # - Split into two: 1 with regexp patterns, and 1 with explicit term_id
+    df_regex = df[df['term_label'].str.startswith(CONFIG['regex_prefix'])]
```
joeflack4, line 182: FYI: I split the REGEX rows and the explicit term rows, do queries, and then join the results.

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
```diff
@@ -0,0 +1,299 @@
...
+    df_explicit = df.filter(items=set(df.index) - set(df_regex.index), axis=0)
+
+    # Query 1: on explicit term_id
+    if len(df_explicit) > 0:
+        df_explicit_results = expand_ontological_exclusions(onto_path, df_explicit, cache_suffix='1')
+
+    # Query 2: on regex pattern
+    # todo: later: #x4: use OAK when `.ttl` is supported: https://github.com/INCATools/ontology-access-kit/issues/134
```
joeflack4, line 190: This block of code looks a little messy, but it does the job. I use the Jinja template to do 1 query per REGEX pattern, then combine the results.

I tried to use OAK, but `.ttl` doesn't seem to be supported yet, so I made a ticket: https://github.com/INCATools/ontology-access-kit/issues/134

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
```diff
@@ -0,0 +1,299 @@
...
+from argparse import ArgumentParser
+from typing import Dict, List
+
+from jinja2 import Template
+import pandas as pd
+
+
+# Vars
+# # Config
+CONFIG = {
```
joeflack4, line 30: I left some explicit config options for fine-tuning / having a place to go to quickly change some things as needed. Most of these are just helpful for development.

src/sparql/get-terms-by-regex.sparql.jinja2
```diff
@@ -0,0 +1,9 @@
+prefix owl:  <http://www.w3.org/2002/07/owl#>
+prefix skos: <http://www.w3.org/2004/02/skos/core#>
+
+select ?term_id ?term_label where {
+  ?term_id a owl:Class;
+    skos:prefLabel ?term_label .
+  FILTER (REGEX(?term_label, "{{ regexp }}"))
+  # FILTER(isIRI(?term_id) && (STRSTARTS(str(?term_id), "{{ iri }}")))  # leaving this out and doing in python for now
```
joeflack4, line 8: I created a new SPARQL Jinja template for REGEX queries.

In a different review comment, I mentioned that there could be a problematic edge case where sometimes terms that are xrefs from other ontologies might be caught and returned by the regexp pattern. I'm not doing this currently, though.

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
matentzn: yes, lets just handle this case and do nothing.

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
matentzn: It is, just not with pronto. Responded on issue for recommended procedure.

src/ontology/config/icd10cm_exclusion_reasons.tsv
```diff
@@ -0,0 +1,770 @@
+term_id	exclusion_reason
```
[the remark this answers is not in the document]
matentzn: Can we rename this file to src/ontology/config/icd10cm_exclusion_reasons_.robot.template.tsv (and the other ones too)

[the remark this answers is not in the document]
joeflack4: Yeah, sure, np. Did you mean to include the `_` at the end of `icd10cm_exclusion_reasons_`? I'm guessing I should leave that out, but correct me if I'm wrong.

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
joeflack4: I hear you, but I don't know off the top of my head how to write a SPARQL query (or use one of our libraries, for that matter) to avoid this edge case. Maybe you have some idea?

Here's the SPARQL I have now for this query now:
```sparql
prefix skos: <http://www.w3.org/2004/02/skos/core#>

  ?term_id a owl:Class;
    skos:prefLabel ?term_label .
  FILTER (REGEX(?term_label, "{{ regexp }}"))
  # FILTER(isIRI(?term_id) && (STRSTARTS(str(?term_id), "{{ iri }}")))  # Would be good to filter out xrefs from other ontologies. This could be done here or in Python.
}
```

The edge case I'm worried about would be caused by `where {?term_id a owl:Class; ...`

So actually, you are correct to be confused about the edge case catching xrefs. This wouldn't capture xrefs. But it still could capture instances where a class from a separate ontology is listed. I don't know why this would occur, but it still could theoretically happen, right?

Could avoid this by checking if the class is a descendant of the root node of the target ontology, but that would require this script to be more complex and we'd have to store the root node somewhere and pass it as a param.

We could also just not worry about this if we think it won't happen; up to you.

src/ontology/mondo-ingest.Makefile
joeflack4: figured this one out. going to try on my own and if theres an issue ill report back

src/scripts/term_range_expansion.py
joeflack4: will verify that the new make goal works and if so, i can drop this hacky logic

matentzn:
Looked at this today, @joeflack4 still working on adding the relevant signature logic discussed above.

matentzn:
- [ ] Merge #34 before moving on here.

src/ontology/mondo-ingest.Makefile
```diff
@@ -160,15 +160,20 @@ mappings: sssom $(ALL_MAPPINGS)
...
-
+# Documentation for `report-mapping-annotations` and `update-jinja-sparql-queries`: `docs/developer/ordo.md`
 report-mapping-annotations:
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/report_mapping_annotations.py
 
 update-jinja-sparql-queries:
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/create_sparql__ordo_replace_annotation_based_mappings.py
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/create_sparql__ordo_mapping_annotations_violation.py
 
+config/%_term_exclusions.txt: config/%_exclusions.tsv component-download-%.owl
```
joeflack4, line 171: @matentzn Slight issue I'm having with `component-download-%.owl` requirement here.

I don't know if this is just for ICD10CM or all ontologies, but this always downloads, even if the file is always there. I realize that this is probably a good practice in production, but it slows development down a bit. I'm wondering if it's possible to improve these parts of our pipelines by using file hashes (e.g. MD5)? What I mean is, whenever we download or upload a file, we store the hash somewhere, either alongside the file itself, or somewhere within the file name. That way we can check if the hash of the file online is the same as the hash currently on disk. If so, we can skip the download step. What do you think?

For reference, here's the make goal for `component-download-icd10cm.owl`:
```make
component-download-icd10cm.owl: | $(TMPDIR)
	if [ $(MIR) = true ]; then wget $(ICD10CM_URL) -O $(TMPDIR)/icd10cm.tmp.owl && $(ROBOT) remove -i 
```

src/ontology/mondo-ingest.Makefile
```diff
@@ -160,15 +160,20 @@ mappings: sssom $(ALL_MAPPINGS)
...
+	python3 ../scripts/exclusion_term_expansion/exclusion_term_expansion.py \
+	--exclusions-path $(word 1,$^) \
+	--onto-path $(TMPDIR)/$@.owl \
```
joeflack4, line 174: @matentzn 
## Summary
I think you had a bug in one of the recommended changes to this make command. But I fixed it.

## Full bugfix description
(a) `--onto-path $(TMPDIR)/$@.owl` vs (b) `--onto-path $(word 2,$^)` vs (c) (see below)

(b) is how I originally had things in my last update, but you recommended a change to (a). I think this is a mistake.
- (a) expands to: `--onto-path tmp/config/icd10cm_term_exclusions.txt.owl`
- (b) expands to: `--onto-path component-download-icd10cm.owl

I looked at `tmp` and it looks like the file that I need to be referencing is: `component-download-icd10cm.owl.owl`

Therefore, I think I fixed this bug by changing to (c): `--onto-path $(TMPDIR)/$(word 2,$^).owl`
...which expands to: `tmp/component-download-icd10cm.owl.owl`

joeflack4, line 174: @matentzn 
## `.owl.owl` issue
Also, I think we spoke briefly about this before, but what's with the `.owl.owl` (e.g. `component-download-icd10cm.owl.owl`). Is this just a bug that needs to be fixed? That is, do we need to update several of our make commands such that we only produce/reference files with `.owl` rather than `.owl.owl`?

joeflack4: As further indication that there is something strange with the `component-download-%.owl` recipe, notice that this is the only case where below I have to also list the directory (`$(TMPDIR)/`) and the additional file extension `.owl`. For the other dependencies, the name/path of the make recipe is the same name/path as the file generated.

```make
config/%_term_exclusions.txt: config/%_exclusions.tsv component-download-%.owl $(TMPDIR)/%_relevant_signature.txt
	python3 ../scripts/exclusion_term_expansion/exclusion_term_expansion.py \
	--exclusions-path $(word 1,$^) \
	--onto-path $(TMPDIR)/$(word 2,$^).owl \  # <---- the odd one
	--relevant-signature-path $(word 3,$^)
```

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
```diff
@@ -0,0 +1,316 @@
...
+TODO's
+  - todo: Do we need to worry about edge case where exclusion table is empty? (e.g. we include such a table for all
+     ontologies, but there exists 1 where we don't exclude anything.
+  - todo: later: see below: #x1: regex results: would be good to filter out by IRI or prefix that is not from source
+     ontology, but how to determine the appropriate prefix just based on the 2 file params, but could have edge cases.
+  - todo: later: see below: #x2
+  - todo: later: see below: #x3
+  - todo: later: see below: #x4: Switch from SPARQL to OAK.
+  - todo: later: When constructing exclusion TSV: bug: ICD10CM:C7A-C7A reported as term, but should be ICD10CM:C7A
+  - todo: later: QA: possible conflicts in icd10cm_exclusions.tsv: Add a check to raise an error in event of a parent
```
joeflack4, line 18: @matentzn W/ the holiday this week and a sick day, I'm a bit low on time. I think this edge case is a low probability, so I have just added a todo comment for now. I don't think this is something I can do in like ~10 minutes. Probably would take me more than 30. So I think this is a low priority right now w/ the other stuff on my plate.

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
```diff
@@ -0,0 +1,316 @@
...
+    df_regex_results = None
+    exclusion_fields = ['exclusion_reason', 'exclude_children']
+    sep = '\t' if exclusions_path.endswith('.tsv') else ',' if exclusions_path.endswith('.csv') else ''
+    onto_name = os.path.basename(onto_path).split('.')[0]
+    df_results_path = os.path.join(CONFIG_DIR, f'{onto_name}_exclusion_reasons.robot.template.tsv')
+    df_results_simple_path = os.path.join(CONFIG_DIR, f'{onto_name}_term_exclusions.txt')
+
+    # Prepare data
+    df = pd.read_csv(exclusions_path, sep=sep).fillna('')
+    df['term_id'] = df['term_id'].apply(lambda x: x.strip())
```
joeflack4, line 184: @matentzn FYI: A bug had slipped in. I had whitespace around some of the term IDs in my exclusions file. This is one of those edge cases I sometimes forget to handle. But it is handled now. I was able to notice this bug after I examined any terms that were filtered out by `relevant_signature.txt`. After fixing the whitespace issue, no terms are filtered out anymore.

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
```diff
@@ -0,0 +1,316 @@
...
+        df_regex_results = expand_ontological_exclusions(onto_path, df_regex_pre_results, cache_suffix='2')
+
+    # Combine & massage
+    # - CONCAT: Combine results of all queries
+    df_results = pd.concat([x for x in [df_explicit_results, df_regex_results] if x is not None])
+    # - Drop duplicates: As there is likely to be overlap between (a) explicitly listed terms, and (b) Regex-selected
+    # terms. It is also theoretically possible that there may be some cases where there is a different exclusion_reason
+    # for a term that is within both (a) and (b). In that case, duplicate terms will remain, showing both reasons.
+    df_results = df_results.drop_duplicates()
+    # - Filter: terms not in relevant_signature.txt, e.g. terms actually in ICD / that we care about
```
joeflack4, line 229: @matentzn FYI: Utilization of `relevant_signature.txt`

src/ontology/mondo-ingest.Makefile
```diff
@@ -160,15 +160,21 @@ mappings: sssom $(ALL_MAPPINGS)
...
 update-jinja-sparql-queries:
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/create_sparql__ordo_replace_annotation_based_mappings.py
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/create_sparql__ordo_mapping_annotations_violation.py
 
+config/%_term_exclusions.txt: config/%_exclusions.tsv component-download-%.owl $(TMPDIR)/%_relevant_signature.txt
+	python3 ../scripts/exclusion_term_expansion/exclusion_term_expansion.py \
+	--exclusions-path $(word 1,$^) \
+	--onto-path $(TMPDIR)/$(word 2,$^).owl \
+	--relevant-signature-path $(word 3,$^) \
+	--onto-name $*
```
joeflack4, line 176: @matentzn FYI: Added param `--onto-name` to script. This is good for future proofing. Previously, I was extracting the ontology name from one of the file names that was passed as a param. However, as our pipeline changes, those file names are also subject to change. As such, it is more stable to pass the ontology name explicitly.

joeflack4:
I have updated the makefile recipe and python script to utilize other recipes which provide ICD10CM in `.owl` format, and I have verified that it all works.

joeflack4:
@matentzn This is ready to merge but if you could look through and resolve items, that would be nice if you have time.

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
matentzn: Oh, the expansion  process should _only_ use rdfs:subClassOf and owl:equivalentClass, no other relation (in particular not xref). Is this what you are asking?

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
matentzn: Was this fixed?

src/ontology/config/icd10cm_exclusion_reasons.robot.template.tsv
```diff
@@ -0,0 +1,173 @@
+term_id	exclusion_reason
+ID	AI rdfs:seeAlso
```
matentzn, line 2: instead of seeAlso, we should use https://github.com/monarch-initiative/mondo/issues/5177 when its done

src/ontology/config/icd10cm_exclusion_reasons.robot.template.tsv
```diff
@@ -0,0 +1,173 @@
...
+ICD10CM:B95-B97	disease causing agents
```
matentzn, line 3: There should use only _coded_ exclusion reasons, like we do for ORDO. (`MONDO:excludeRare`). Can you work with @sabrinatoro to make sure the correct codes are used here, not the "labels" of the exclusion reasons?

src/ontology/config/icd10cm_term_exclusions.txt
```diff
@@ -0,0 +1,171 @@
+ICD10CM:B95-B97
```
matentzn, line 1: These seem very few: Given that src/ontology/config/icd10cm_exclusions.tsv has 115 rows, and most of them exclude all children, I expected this list to bein the 1000+ range, perhaps even 10K!

src/sparql/get-terms-children.sparql.jinja2
```diff
@@ -0,0 +1,9 @@
...
+    skos:prefLabel ?child_label .
```
matentzn, line 8: almost none of the ontologies we care about use skos:prefLabel. This should either be optional or not here at all.. Why is it needed?

src/sparql/get-terms-by-regex.sparql.jinja2
```diff
@@ -0,0 +1,9 @@
+prefix owl:  <http://www.w3.org/2002/07/owl#>
+prefix skos: <http://www.w3.org/2004/02/skos/core#>
+
+select ?term_id ?term_label where {
+  ?term_id a owl:Class;
+    skos:prefLabel ?term_label .
```
matentzn, line 6: Almost none of the ontologies we care about use skos:prefLabel. most use `rdfs:label`. If you want to obtain the label this way you will have to use VALUES { rdfs:label skos:prefLabel }

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
```diff
@@ -0,0 +1,312 @@
...
+CONFIG = {
+    'use_cache': False,
+    'include_superclass': False,  # is this the correct name for what this does: * property path qualifier
+    'save': True,  # save dataframes to disk
+    'regex_prefix': 'REGEX:'  # used to know which rows in exclusion TSVs are regex pattern searches
+}
+
+# # Static
+THIS_SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
+CACHE_DIR = os.path.join(THIS_SCRIPT_DIR, 'cache')
```
matentzn, line 43: The universal "cache" dir in all ODK repos is:

```
src/ontology/tmp
```

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
```diff
@@ -0,0 +1,312 @@
...
+PROJECT_DIR = os.path.join(SCRIPTS_DIR, '..')
+ONTOLOGY_DIR = os.path.join(PROJECT_DIR, 'ontology')
+CONFIG_DIR = os.path.join(ONTOLOGY_DIR, 'config')
+SPARQL_DIR = os.path.join(PROJECT_DIR, 'sparql')
+SPARQL_CHILD_INCLUSION_PATH = os.path.join(SPARQL_DIR, 'get-terms-children.sparql.jinja2')
+SPARQL_TERM_REGEX_PATH = os.path.join(SPARQL_DIR, 'get-terms-by-regex.sparql.jinja2')
+
+
+# Functions
+def uri_to_curie(uri: str) -> str:
```
matentzn, line 54: Not this time, but lets try to use this next time:

https://bioregistry.readthedocs.io/en/latest/api/bioregistry.curie_from_iri.html?highlight=get_curie#bioregistry.curie_from_iri

matentzn (commented):
Looks good, left some comments. 

We need to figure out some way to make smaller pull requests.. If you have an idea, let me know. I  did not review much of the python code - maybe we can agree that rather than looking in much detail "how" you do stuff, I will focus on the IO (files going into the process and produced by the process). What do you think?

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
joeflack4: I was just asking about how to filter out so only classes that belong to that ontology are included in the results. But you actually recommended a good solution for this earlier; you had me use the `signature-%` files. So I believe this is resolved.

joeflack4:
_[@matentzn wrote](https://github.com/monarch-initiative/mondo-ingest/pull/32#pullrequestreview-1038562485):_
> We need to figure out some way to make smaller pull requests.. If you have an idea, let me know. I did not review much of the python code - maybe we can agree that rather than looking in much detail "how" you do stuff, I will focus on the IO (files going into the process and produced by the process). What do you think?

Yep, that's fine with me. I'll move to doing FYIs only if they seem pretty important, otherwise will comment mostly where I need assistance or want to point out a bug/problem that is on my radar but haven't fixed yet.

src/ontology/config/icd10cm_exclusion_reasons.robot.template.tsv
joeflack4: Sounds good! I added a TODO above the appropriate line for now and referenced that issue.

src/ontology/config/icd10cm_exclusion_reasons.robot.template.tsv
sabrinatoro: @joeflack4 We have this: https://mondo.readthedocs.io/en/latest/editors-guide/exclusion-reasons/ (also [yaml here](https://github.com/monarch-initiative/mondo/blob/master/src/schema/mondo.yaml). Is this what you are asking?

src/ontology/config/icd10cm_term_exclusions.txt
joeflack4: @matentzn Hmm... Potentially what I could do is, rather than trying to analyze why this is low, I could use OAK instead of SPARQL and see if that changes anything.

FYI, the sparql code that generates this is:
`{{property_path_qualifier}}` below expands to `+` by default, but can also expand to `*` depending on an additional flag I can pass to the Python script. However, I'm not sure if this has anything to do with the issue.
```sparql
select ?term_id ?child_id ?child_label where {
  ?child_id rdfs:subClassOf{{property_path_qualifier}} ?term_id ;
    skos:prefLabel ?child_label .
}
```

src/sparql/get-terms-children.sparql.jinja2
joeflack4: I think there probably was not a reason; probably just slipped my mind. Removed!

src/sparql/get-terms-by-regex.sparql.jinja2
joeflack4: Hmmm, I don't understand the need for `VALUES` here but no need to worry about explaining if you're short on time. Idk why I included return of labels here. For this one, I might have thought it was necessary to select the label in order for the regex to work. But I'm removing it now and I imagine it'll run fine.

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
joeflack4: Done! Also changed for my other script.

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
joeflack4: sounds good! will add a TODO

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
```diff
@@ -0,0 +1,312 @@
...
+    if '_' in end_bit:
+        curie = end_bit.replace('_', ':')
+    else:
+        curie = uri_list[-2] + ':' + end_bit
+    return curie
+
+
+# todo: later #x2: terms vs regexp: this would be better as a class or done some other way probably
+# todo: later #x3: could be nice to add param to auto-do uri_to_curie()
+def sparql_jinja2_robot_query(
```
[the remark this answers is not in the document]
joeflack4: Hmm, I agree this function is not 100% general. It is almost general enough to be given that name, but it does have a couple specifics. I know that the Jinja is out of place a little, but I think it is actually more in line with how we do things in our pipeline (flat files of all different kinds) rather than constructing in runtime. The only thing that'd make it more in line with our pipeline is if I wrote make goals to instantiate the Jinja.

src/ontology/mondo-ingest.Makefile
joeflack4: @matentzn Per requested reduction of FYI comments I'm going to mark this one resolved.

src/ontology/config/icd10cm_exclusion_reasons.robot.template.tsv
joeflack4: Sounds good! I added a TODO for now.

src/ontology/config/icd10cm_term_exclusions.txt
joeflack4: This is really weird and annoying. I responded to like 5 of your comments and my responses disappeared from GitHub.

I think you are meant to refer to `icd10cm_exclusion_reasons.robot.template.tsv`, which has ~173 rows. 

I wonder if there's an issue w/ the query? `property_path_qualifier` is `+` by default, but can also be `*` depending on a flag. However I'm not sure if that is the issue.
```sparql
select ?term_id ?child_id where {
  ?child_id rdfs:subClassOf{{property_path_qualifier}} ?term_id .
}
```

My plan is to migrate to OAK and see if that changes anything.

src/ontology/config/icd10cm_exclusion_reasons.robot.template.tsv
joeflack4: Twice I responded to this and my comments (including like 5-10 others) were deleted! I just saw a red banner at the top of the page I must not have noticed earlier, ugh.

Anyway sure thing. I made a TODO for now.

joeflack4: ok third time ive added a comment and its not showing up. wth is happening

src/ontology/config/icd10cm_term_exclusions.txt
joeflack4: ok now my comment showed up again...

src/ontology/config/icd10cm_exclusion_reasons.robot.template.tsv
joeflack4: @sabrinatoro I think that .yml will do it!

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
joeflack4: resolving for fyi reduction as requested

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
joeflack4: resolving for fyi reduction as requested

src/ontology/metadata/icd10cm.yml
```diff
@@ -4,6 +4,7 @@ description: >
 homepage: https://bioportal.bioontology.org/ontologies/ICD10CM
 prefix_map:
   skos: "http://www.w3.org/2004/02/skos/core#"
```
joeflack4, line 6: @matentzn Using these `metadata/*.yml` files is still probably the best way to get the list of prefixes that I need to filter for to look for terms to add to exclusions list. However, I noticed that `skos` appears here. We don't want those terms in ICD10CM exclusions table. For now, I've got a list of prefixes in my Python script that I will excluded should they appear here (e.g. `skos`, `rdfs`, etc). However it's important that you know about this.

**My request:** Can we update this yml to make it clear which prefixes 'belong' to the ontology?
a. A new top-level key, such as `prefixes_owned_by_ontology`
b. Convert `prefix_map` from a list of key value pairs, to a more complex object, e.g. 
```yml
prefix_map:
    ICD10CM: 
        uri: https://icd.codes/icd10cm/
        owned_by_ontology: true
```

I prefer a.

src/ontology/mondo-ingest.Makefile
```diff
@@ -160,15 +160,17 @@ mappings: sssom $(ALL_MAPPINGS)
...
-
+# Documentation for `repo   rt-mapping-annotations` and `update-jinja-sparql-queries`: `docs/developer/ordo.md`
 report-mapping-annotations:
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/report_mapping_annotations.py
 
 update-jinja-sparql-queries:
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/create_sparql__ordo_replace_annotation_based_mappings.py
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/create_sparql__ordo_mapping_annotations_violation.py
 
+config/%_term_exclusions.txt: config/%_exclusions.tsv component-download-%.owl $(TMPDIR)/%_relevant_signature.txt
```
joeflack4, line 171: Now only `--onto-name` is needed. Paths to these files can be ascertained. However I've left them as optional CLI args.

src/ontology/metadata/icd10cm.yml
joeflack4: @matentzn Also, FYI. For some reason the `mondo.sssom.tsv` that I had listed (1) `https://icd.codes/icd10cm/` instead of (2) `http://purl.bioontology.org/ontology/ICD10CM/`. (2) is what is being used in the actual ICD10CM ontology.

src/scripts/exclusion_term_expansion.py
```diff
@@ -0,0 +1,378 @@
...
+    # TODO: print any that were intensional exclusion expansions but showed in relevant sig
+    #  ...and add as a comment fo nico
+    df_relevant_terms = pd.read_csv(relevant_signature_path).fillna('')
+    df_relevant_terms['term'] = df_relevant_terms['term'].apply(uri_to_curie)
+    relevant_terms: List[str] = list(set(df_relevant_terms['term']))
+    # todo: some redundancy in these 2 lines that could be reduced. see if i can assign ~df_results... to a variable
+    #   and re-use that for both the print statement and the df_results2 filter
+    relevant_terms_accidentally_excluded = [x for x in relevant_terms if x in set(df_results['term_id'])]
+    if relevant_terms_accidentally_excluded:
+        print(f'Warning: The extensionalization of intensional exclusion table resulted in the following terms which '
```
joeflack4, line 276: I realized that this edge case could happen, and decided to handle it and print a warning if it occurs:

```python
print(f'Warning: The extensionalization of intensional exclusion table resulted in the following terms which '
              f'would have been excluded, but are contained within {os.path.basename(relevant_signature_path)}, and'
              f' thus have been removed from exclusion: {relevant_terms_accidentally_excluded}', file=sys.stderr)
```

joeflack4: @matentzn @sabrinatoro @nicolevasilevsky Hmm, interesting. Glad I put this warning here. I was running ICD10CM and I did notice that there are terms in the `relevant_signature` file that appear in the `exclusions` file:

```
python /exclusion_term_expansion.py -n icd10cm
Warning: The extensionalization of intensional exclusion table resulted in the following terms which would have been excluded, but are contained within icd10cm_relevant_signature.txt, and thus have been removed from exclusion: ['ICD10CM:E89-E89', 'ICD10CM:P96.3', 'ICD10CM:P92', 'ICD10CM:N99-N99', 'ICD10CM:P94.8', 'ICD10CM:R00-R09', 'ICD10CM:P05-P08', 'ICD10CM:Z19-Z19', 'ICD10CM:P91.8', 'ICD10CM:R10-R19', 'ICD10CM:Z66-Z66', 'ICD10CM:P94', 'ICD10CM:D78-D78', 'ICD10CM:P94.0', 'ICD10CM:P00-P04', 'ICD10CM:O85-O92', 'ICD10CM:P96.5', 'ICD10CM:E16.3', 'ICD10CM:O60-O77', 'ICD10CM:C7B-C7B', 'ICD10CM:P52', 'ICD10CM:K40-K46', 'ICD10CM:P36', 'ICD10CM:P96.9', 'ICD10CM:Z14', 'ICD10CM:M40', 'ICD10CM:P96.8', 'ICD10CM:M96-M96', 'ICD10CM:P91.61', 'ICD10CM:P96.89', 'ICD10CM:P91', 'ICD10CM:E36-E36', 'ICD10CM:P92.0', 'ICD10CM:P92.2', 'ICD10CM:E16.0', 'ICD10CM:P94.2', 'ICD10CM:V10-V19', 'ICD10CM:P91.63', 'ICD10CM:P92.3', 'ICD10CM:V70-V79', 'ICD10CM:R47-R49', 'ICD10CM:E16.2', 'ICD10CM:T79-T79', 'ICD10CM:X52-X58', 'ICD10CM:P93', 'ICD10CM:E15', 'ICD10CM:Z40-Z53', 'ICD10CM:X10-X19', 'ICD10CM:P92.5', 'ICD10CM:P91.2', 'ICD10CM:P96.0', 'ICD10CM:Z30-Z39', 'ICD10CM:P91.62', 'ICD10CM:V80-V89', 'ICD10CM:P91.4', 'ICD10CM:Y90-Y99', 'ICD10CM:P60', 'ICD10CM:C7A-C7A', 'ICD10CM:P51', 'ICD10CM:P91.0', 'ICD10CM:P91.811', 'ICD10CM:P91.6', 'ICD10CM:X71-X83', 'ICD10CM:Z00-Z13', 'ICD10CM:Z77-Z99', 'ICD10CM:P91.821', 'ICD10CM:R30-R39', 'ICD10CM:P09-P09', 'ICD10CM:P93.8', 'ICD10CM:Y62-Y69', 'ICD10CM:R20-R23', 'ICD10CM:R25-R29', 'ICD10CM:P96', 'ICD10CM:M26', 'ICD10CM:P96.82', 'ICD10CM:P92.8', 'ICD10CM:R97-R97', 'ICD10CM:P59', 'ICD10CM:B95-B97', 'ICD10CM:V00-Y99', 'ICD10CM:P92.09', 'ICD10CM:V50-V59', 'ICD10CM:V40-V49', 'ICD10CM:P50', 'ICD10CM:Z55-Z65', 'ICD10CM:V00-V09', 'ICD10CM:P91.88', 'ICD10CM:H59-H59', 'ICD10CM:P84-P84', 'ICD10CM:P92.4', 'ICD10CM:V90-V94', 'ICD10CM:Y83-Y84', 'ICD10CM:Z69-Z76', 'ICD10CM:R70-R79', 'ICD10CM:P91.82', 'ICD10CM:P96.81', 'ICD10CM:Z67-Z67', 'ICD10CM:W00-W19', 'ICD10CM:P91.829', 'ICD10CM:P93.0', 'ICD10CM:P92.6', 'ICD10CM:P92.01', 'ICD10CM:P96.1', 'ICD10CM:R90-R94', 'ICD10CM:V30-V39', 'ICD10CM:X92-Y09', 'ICD10CM:V98-V99', 'ICD10CM:P91.60', 'ICD10CM:O09-O09', 'ICD10CM:H95-H95', 'ICD10CM:M99-M99', 'ICD10CM:P91.822', 'ICD10CM:O00-O08', 'ICD10CM:Z18-Z18', 'ICD10CM:Y21-Y33', 'ICD10CM:V60-V69', 'ICD10CM:P96.83', 'ICD10CM:P95', 'ICD10CM:Z20-Z29', 'ICD10CM:R83-R89', 'ICD10CM:T66-T78', 'ICD10CM:P96.2', 'ICD10CM:W85-W99', 'ICD10CM:R99-R99', 'ICD10CM:P58', 'ICD10CM:V20-V29', 'ICD10CM:Y35-Y38', 'ICD10CM:P53', 'ICD10CM:P91.5', 'ICD10CM:P94.9', 'ICD10CM:V95-V97', 'ICD10CM:L76-L76', 'ICD10CM:Z68-Z68', 'ICD10CM:W65-W74', 'ICD10CM:P91.81', 'ICD10CM:R50-R69', 'ICD10CM:P90', 'ICD10CM:P92.1', 'ICD10CM:O80-O82', 'ICD10CM:X50-X50', 'ICD10CM:R00-R99', 'ICD10CM:X30-X39', 'ICD10CM:Z00-Z99', 'ICD10CM:P91.1', 'ICD10CM:R80-R82', 'ICD10CM:P10-P15', 'ICD10CM:P54', 'ICD10CM:P91.9', 'ICD10CM:O30-O48', 'ICD10CM:T15-T19', 'ICD10CM:X00-X08', 'ICD10CM:P91.3', 'ICD10CM:E16.1', 'ICD10CM:Y70-Y82', 'ICD10CM:W50-W64', 'ICD10CM:P92.9', 'ICD10CM:P94.1', 'ICD10CM:Z17-Z17', 'ICD10CM:P91.819', 'ICD10CM:R40-R46', 'ICD10CM:L49', 'ICD10CM:E16.4', 'ICD10CM:P91.823', 'ICD10CM:P80-P83', 'ICD10CM:W20-W49', 'ICD10CM:J95-J95']
```

I was thinking that maybe this is not a real problem? Maybe it the case that the `relevant_signature` file is an initial selection of relevant terms, but the exclusions are meant to override `relevant_signature`? But I think not, because Nico asked me to explicitly not exclude terms that are in `relevant_signature`.

So I'm curious as to why this is happening. Perhaps it is an artefact of the complexities of multiple inheritance from polyhierarchy? Or perhaps these terms just should not have been added to the exclusion table?

src/scripts/exclusion_term_expansion.py
```diff
@@ -0,0 +1,422 @@
...
+
+    # CONCAT: Combine term child results w/ terms where children were excluded
+    df_results = pd.concat([df_kids0, df_kids1_results]).fillna('')
+
+    return df_results
+
+
+# TODO: pass prefixes to this func. need to move logic up to run() I think and pass down to both funcs
+# TODO:
+def get_non_inclusions(mirror_signature_path: str, component_signature_path: str, prefix_uris: List[str]) -> pd.DataFrame:
```
joeflack4, line 206: @nicolevasilevsky @sabrinatoro @matentzn Per Nico's instruction from [the google doc](https://docs.google.com/document/d/1-bFd6GeHcLHImjKImML_j0uAyKENIlb5NN64oGP4rwY/edit?disco=AAAAc-gOiSs), I added the `exclusion_reason` `MONDO:excludeNonDisease` for all of these exclusions (in 'mirror' but not in 'component'). However, I noticed that this reason is not listed in the docs: https://mondo.readthedocs.io/en/latest/editors-guide/exclusion-reasons/

src/scripts/exclusion_term_expansion.py
joeflack4: @matentzn Prior to last Friday, I was instructed to use the `relevant_signature` file. The idea was to look at the extensionalized list of exclusions and filter out any that were in `relevant_signature`. However, I was just now getting all the warnings that I crossed out above, and 0 'relevant' extensionalized exclusions in my results.

Then I looked at your [google doc](https://docs.google.com/document/d/1-bFd6GeHcLHImjKImML_j0uAyKENIlb5NN64oGP4rwY/edit?disco=AAAAc-gOiSs), and noticed that `relevant_signature` was dropped from the requirements. As such, I've discarded the use for `relevant_signature`. Just want to get final verification from you that this is indeed the intent; that our exclusion tables are essentially overriding `relevant_signature`.

src/ontology/mondo-ingest.Makefile
joeflack4, line 171: @matentzn Only able to get success on the mirror and component recipes for: ICD10CM, ICD10WHO, and OMIM. I saved the error messages for the other ones, but I haven't had time to look much into them yet. I know that for ORDO, I got a 404 not found.

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
matentzn: ok make an issue instead. Good prioritsation.

src/ontology/config/icd10cm_term_exclusions.txt
matentzn: Just always use * instead of +. Does this change the number of classes?

src/ontology/mondo-ingest.Makefile
matentzn: This will deal with the ORDO issue: https://github.com/monarch-initiative/mondo-ingest/pull/41 (merge, then run `sh run.sh make update_repo`)

matentzn: Identical with reports/component_signature-omim.tsv

matentzn: Not quite identical, but same purpose.

src/scripts/exclusion_term_expansion.py
matentzn: Replaced by e.g. `reports/component_signature-omim.tsv`

src/ontology/mondo-ingest.Makefile
matentzn: Ok for now, but generally in cases like this, we always want to see the parameters passed in the python scripts explicitly

src/ontology/metadata/icd10cm.yml
matentzn: Oh yes, we probably need to change to http://purl.bioontology.org/ontology/ICD10CM/ in mondo.config. But make sure this happens at source, could be that that file is pulled from mondo repo (check mondo-ingest.Makefile)

matentzn: https://github.com/monarch-initiative/mondo-ingest/blob/main/src/ontology/metadata/doid.yml#L11 

This is what I intended for the base prefixes! Does this not work?

src/ontology/mondo-ingest.Makefile
joeflack4: Merged. Will run that now!

Full error messages for reference:
### DO
```
rget `mondo-ingest-full.owl'
Makefile:516: warning: ignoring old commands for target `mondo-ingest-full.owl'
if [ true  = true ] && [ true  = true ]; then robot --catalog catalog-v001.xml merge -I http://purl.obolibrary.org/obo/doid.owl \
	annotate --ontology-iri http://purl.obolibrary.org/obo/mondo-ingest/component-download-doid.owl annotate -V http://purl.obolibrary.org/obo/mondo-ingest/releases/2022-07-21/component-download-doid.owl --annotation owl:versionInfo 2022-07-21 -o tmp/component-download-doid.owl.owl; fi
if [ true  = true ]; then robot --catalog catalog-v001.xml query -i "tmp/component-download-doid.owl.owl" -q "../sparql/doid-relevant-signature.sparql" tmp/doid_relevant_signature.txt; fi
QUERY PARSE ERROR query cannot be parsed:
Line 7, column 12: Unresolved prefixed name: rdfs:subClassOf
For details see: http://robot.obolibrary.org/query#query-parse-error
Use the -vvv option to show the stack trace.
Use the --help option to see usage information.
make: *** [tmp/doid_relevant_signature.txt] Error 1
```

### ORDO
Got a 404:
```
/mondo-ingest/component-download-ordo.owl annotate -V http://purl.obolibrary.org/obo/mondo-ingest/releases/2022-07-21/component-download-ordo.owl --annotation owl:versionInfo 2022-07-21 -o tmp/component-download-ordo.owl.owl; fi
org.semanticweb.owlapi.io.OWLOntologyCreationIOException: http://www.orphadata.org/data/ORDO/ORDO_en_4.0.owl
Use the -vvv option to show the stack trace.
Use the --help option to see usage information.
make: *** [component-download-ordo.owl] Error 1
```

### NCIT
(i think i need to run again; some of my terminal history got erased for some reason)
[ncit-errors.txt](https://github.com/monarch-initiative/mondo-ingest/files/9169251/ncit-errors.txt)

src/ontology/Makefile
```diff
@@ -41,7 +41,7 @@ REPORT_PROFILE_OPTS =       --profile $(ROBOT_PROFILE)
 OBO_FORMAT_OPTIONS =        
 SPARQL_VALIDATION_CHECKS =  properties-as-annotation-and-object ordo-mapping-annotations 
 SPARQL_EXPORTS =            basic-report 
-ODK_VERSION_MAKEFILE =      v1.3.1
+ODK_VERSION_MAKEFILE =      v1.3.0
```
matentzn, line 44: you have to run: `docker pull obolibrary/odkfull` before running update_repo - you are downgrading the odk here!

joeflack4: Whoops. Can we add that docker command as a dependency or first line in `update_repo`? I wouldn't be a surprised if I (or possibly someone else) accidentally did this again at some point

matentzn: Maybe we could add a -U option to the run.sh command to automatically upgrade, but doing it by default has been decided against because of the possible breaking changes..

src/ontology/metadata/icd10cm.yml
joeflack4: That's great, thanks Nico. I see you've already pushed those changes. I'll update this PR to utilize `base_prefix_map`

Regarding the `mondo.sssom.congi`, I went ahead and updated the ICD10CM prefix: https://github.com/monarch-initiative/mondo-ingest/pull/42

src/ontology/mondo-ingest.Makefile
joeflack4: Ok, that makes sense as an alternative generally and for consistency in the `makefile`s. I added a `todo` about it.

src/scripts/exclusion_term_expansion.py
joeflack4: Alright, thanks for that confirmation. Removing everything related to `relevant_signature` from the PR now.

src/scripts/exclusion_term_expansion.py
joeflack4: I see an issue has been made for this. Marking resolved!

src/scripts/exclusion_term_expansion/exclusion_term_expansion.py
joeflack4: This one was a rare exception where this was actually a legit term: https://www.icd10data.com/ICD10CM/Codes/C00-D49/C7A-C7A

Removing the todo

src/ontology/config/icd10cm_term_exclusions.txt
joeflack4: I haven't checked but I will make the change now.

src/ontology/Makefile
joeflack4: Hmmm. I see what you're saying. I will try to remember, then.

src/ontology/mondo-ingest.Makefile
joeflack4: These are all resolved.

Indeed, the NCIT Java heap space issue still remains. I will make a comment specifically about that.

joeflack4:
As of just now, I have went through and confirmed all open review threads have been resolved.

---

### NCIT Java heap space issue
There is a new issue, but it is not related to the code in this PR. As @matentzn mentioned, I might need to increase my Java memory to resolve this issue:
> `Exception in thread "main" java.lang.OutOfMemoryError: Java heap space`
```
make config/ncit_term_exclusions.txt
if [ true  = true ]; then robot --catalog catalog-v001.xml merge -I http://purl.obolibrary.org/obo/ncit.owl \
	annotate --ontology-iri http://purl.obolibrary.org/obo/mondo-ingest/component-download-ncit.owl annotate -V http://purl.obolibrary.org/obo/mondo-ingest/releases/2022-07-25/component-download-ncit.owl --annotation owl:versionInfo 2022-07-25 -o tmp/component-download-ncit.owl.owl; fi
Exception in thread "main" java.lang.OutOfMemoryError: Java heap space
make: *** [component-download-ncit.owl] Error 1
```

What I tried:
I looked through the `mondo-ingest.makefile` code to try to find anywhere to add something like [`java -Xmx2048m`](https://stackoverflow.com/a/2381865/5258518) (or higher), but I couldn't find anywhere to do that. I think it will be more efficient if @matentzn can instruct me on this.

src/scripts/exclusion_term_expansion.py
```diff
@@ -0,0 +1,398 @@
...
+    df_results = pd.concat([expanded_intensional_exclusions_df, non_inclusions_df])
+    if len(df_results) == 0:
+        df_results = pd.DataFrame()
+        df_results['term_id'] = ''
+
+    # Add simplified version
+    # - df_results_simple: Simpler dataframe which is easier to use downstream for other purposes
+    df_results_simple = df_results[['term_id']]
+    # - df_results: Add special meta rows
+    # todo: Utilize `has_exclusion_reason`, pending mondo#5177 completion:
```
joeflack4, line 299: @matentzn I see that Nicole has finished this [`has_exclusion_reason` issue](https://github.com/monarch-initiative/mondo/pull/5187/files) ([pr](https://github.com/monarch-initiative/mondo/issues/5177)).

I'm wondering which of the following I may / may not need to do:
- [ ] 1. Update output row `{'term_id': 'ID', 'exclusion_reason': 'AI rdfs:seeAlso'}` to say `has_exclusion_reason`.
- [ ] 2. Update any inputs where I'm reading from a column called `exclusion_reason` to `has_exclusion_reason`
- [ ] 3. Update the rest of my outputs to use `has_exclusion_reason` instead of `exclusion_reason`

matentzn: 1. Great catch! {'term_id': 'ID', 'exclusion_reason': 'AI rdfs:seeAlso'} -> {'term_id': 'ID', 'exclusion_reason': 'AI MONDOREL: has_exclusion_reason'}. You don't have to concern yourself with the `MONDOREL` prefix right now.
2. No, don't worry. No other changes are required. Leave it at `exclusion_reason` or whatever else you have it. Only the ROBOT template above needs to be adjusted!
3. No need either. Leave it as it is!

PR #36, opened 2022-07-17
Slurp pipeline: Initial design

matentzn:
Updates
```
commit e5575265457308e40626870dbdee9de791cdcac0

    Slurp pipeline
    - Update: direct_owned_parents methods: Temporarily commented out, pending OAK fixes. Currently doing this in batch using SPARQL for now.
    - Add: Param --onto-exclusions-path: Now we no longer consider exluded terms to be possible slurp candidates.
    - Bugfix: Labels: Now they are successfully being fetched using OAK.
    - Add: New columns to output

    Exclusions
    - Bugfix: Sometimes, when exclude_children was left empty in input config/%_exclusions.tsv, yielded empty generated exclusion tables.
    - Update: Updated exclusion tables for: DOID, OMIM. These had problematic source files with no values for exclude_children, and needed to be re-run after bugfix.

    Misc
    - Update: run.sh: Reverted back to what it was before, undoing deletion of --rf, which was done to address pip installs not being persistent within the ODK docker container between runs.
```

```
commit 5fd00c4d3eaa5741b1d2c229166a686ac4000076

    Slurp pipeline
    - Delete: Initialized a variables for 'all relationships' and 'term parent map', but did a more OAK-reliant refactor and removed these.
    - Add: Term class
    - Update: Big refactor to utilize Term class
    - Add: Slurp output TSV files
    - Add: utils.py, which includes Term class
    - Add: New param: max_id

    Misc
    - Update: run.sh: This change will allow any Python package updates to be retained.
    - Add: SPARQL jinja query to get parents. Used by slurp pipeline.
```

```
commit d3e1c743e15f0e76a060793fc075e799041561cf

Slurp pipeline
- Delete: Initialized a variables for 'all relationships' and 'term parent map', but did a more OAK-reliant refactor and removed these.
```

```
commit f50d5809e71cb0bb02147e1568ab443bb0ee7438

Slurp pipeline
- Delete: `mondo-ingest.Makefile`: (i) `python-install-dependencies` phony target. This has been removed in favor of expediently updating the dependencies in ODK directly. (ii) Removed unused, commented out mondo component target.
- Update: `.gitignore`: Simplified to ignoring whole dir `src/ontology/components/*`
- Update: Python requirements: Removed `semsql` w/ specific version, due to some recent dependency fixes in OAK.

Term exclusions
- Update: Rename: reports/term_exclusions.txt -> reports/excluded_terms.txt
```

```
commit e5fab07c1638f8055a5f73396aa38f575675980a

Some fixes to make files

1. Never depend on a PHONY goal, because they force the whole pipeline to be rebuild - just add them to each goal, installing a few python dependencies that are already installed is fast enough.
2. PHONY goals cannot have wildcard dependencies: https://stackoverflow.com/questions/7887343/makefile-wildcard-static-rule-with-phony
3. Indentation is critical with Make targets, your `# TODO: Temporary debugging:` was wrongly intented to the level of the target name, not the body.

```

```
commit 1e7929bdef740e55223b4b32cb116ca5cbe800d4
Slurp pipeline
- Add: makefile: Missing make goals for dependencies for slurp goal.
- Add: Python: CLI
- Update: makefile: Slurp goal: (i) named keys/vals for all params, (ii) standardization in file/path params.
```

```
(before above commits)
- Add: makefile goals
- Add: initial python script
```

## Related to

- https://github.com/monarch-initiative/phenomics_first_resource/issues/24


src/ontology/mondo-ingest.Makefile
```diff
@@ -275,3 +274,23 @@ mappings/mondo-sources-all-lexical.sssom.tsv: $(SCRIPTSDIR)/match-mondo-sources-
...
 lexical_matches: mappings/mondo-sources-all-lexical.sssom.tsv
+
+#############################
+###### Slurp pipeline #######
+#############################
+slurp/:
+	mkdir -p $@
+
+# Feel free to change the signature. Min ID is the next available Mondo ID.
+slurp/%.tsv: components/%.owl tmp/mondo.sssom.tsv reports/mirror-signature-mondo.tsv | slurp/
```
joeflack4, line 285: @Nico: One thing I don't like about `make` is that it's sometimes hard to get answers to questions on Google. What does the pipe `|` do in the list of dependencies?

matentzn: Google for “make dependencies pipe” - it's a so called order only dependency, which means that its only triggered if at least one other dependency (normal dependency) was updated.

src/ontology/mondo-ingest.Makefile
```diff
@@ -275,3 +274,30 @@ mappings/mondo-sources-all-lexical.sssom.tsv: $(SCRIPTSDIR)/match-mondo-sources-
...
 
 lexical_matches: mappings/mondo-sources-all-lexical.sssom.tsv
+
+#############################
+###### Slurp pipeline #######
+#############################
+slurp/:
+	mkdir -p $@
+
+# TODO: Move this to Makefile. how, when I can't edit it here?
```
joeflack4, line 284: @matentzn: `component-download-mondo.owl`: Need to move this to Makefile. What is standard way for me to do that? I know I'm not supposed to edit it in `mondo-ingest` directly.

src/ontology/mondo-ingest.Makefile
```diff
@@ -275,3 +274,30 @@ mappings/mondo-sources-all-lexical.sssom.tsv: $(SCRIPTSDIR)/match-mondo-sources-
...
+.PHONY: component-download-mondo.owl
```
joeflack4, line 285: @Nico: What is it that makes this target `.PHONY`? I thought that was supposed to mean it was a command that did not produce a file or dir with of the target name. But in this case, it does. Is it `.PHONY` because it only shows the filename, but not the path?

src/ontology/mondo-ingest.Makefile
```diff
@@ -275,3 +274,30 @@ mappings/mondo-sources-all-lexical.sssom.tsv: $(SCRIPTSDIR)/match-mondo-sources-
...
+component-download-mondo.owl: | $(TMPDIR)
```
joeflack4, line 286: @matentzn Is this purl indeed correct?: http://purl.obolibrary.org/obo/mondo.owl . I got it from `tmp/mondo.owl`

joeflack4: I'm curious. Is http://purl.obolibrary.org/obo/mondo.owl updated at the same time as the [latest release](https://github.com/monarch-initiative/mondo/releases/latest)?

joeflack4: This url (http://purl.obolibrary.org/obo/mondo.owl) is now used by two goals: `tmp/mondo.owl` and `component-download-mondo.owl`. Should we make a variable for it and share that variable?

joeflack4: @matentzn Does my `component-download-mondo.owl` look correct? I repurposed it from `component-download-omim.owl`

src/ontology/mondo-ingest.Makefile
```diff
@@ -275,3 +274,34 @@ mappings/mondo-sources-all-lexical.sssom.tsv: $(SCRIPTSDIR)/match-mondo-sources-
...
+# min-id: the next available Mondo ID
+slurp/%.tsv: $(COMPONENTSDIR)/%.owl $(TMPDIR)/mondo.sssom.tsv $(REPORTDIR)/mirror_signature-mondo.tsv | slurp/
+	python $(SCRIPTSDIR)/migrate.py \
+	--ontology-path $(COMPONENTSDIR)/%.owl \
+	--sssom-map-path $(TMPDIR)/mondo.sssom.tsv \
+	--min-id 123000 \
+	--mondo-terms-path $(REPORTDIR)/mirror_signature-mondo.tsv \
+	--outpath $@
+
+slurp-%: slurp/%.tsv
```
joeflack4, line 303: Any idea why `make slurp-omim` fails?: ``make: *** No rule to make target `slurp-omim'.  Stop.``

src/ontology/mondo-ingest.Makefile
```diff
@@ -275,3 +274,34 @@ mappings/mondo-sources-all-lexical.sssom.tsv: $(SCRIPTSDIR)/match-mondo-sources-
...
+component-download-mondo.owl: | $(TMPDIR)
+	if [ $(MIR) = true ] && [ $(COMP) = true ]; then $(ROBOT) merge -I http://purl.obolibrary.org/obo/mondo.owl \
+	annotate --ontology-iri $(ONTBASE)/$@ $(ANNOTATE_ONTOLOGY_VERSION) -o $(TMPDIR)/$@.owl; fi
+
+# TODO: Move this to Makefile. how, when I shouldn't edit it from this repo?
+$(COMPONENTSDIR)/mondo.owl: component-download-mondo.owl
+	if [ $(COMP) = true ] ; then if cmp -s $(TMPDIR)/component-download-mondo.owl.owl $@ ; then echo "Component identical."; else echo "Component is different, updating." && cp $(TMPDIR)/component-download-mondo.owl.owl $@; fi; fi
+
+# min-id: the next available Mondo ID
+slurp/%.tsv: $(COMPONENTSDIR)/%.owl $(TMPDIR)/mondo.sssom.tsv $(REPORTDIR)/mirror_signature-mondo.tsv | slurp/
```
joeflack4, line 295: @matentzn I think I asked this before, but I'm still not sure on when to use static text, e.g. `reports/` vs the resolved variable, e.g. `$(REPORTDIR)`. The latter feels better, so I used that for these directories.

src/ontology/mondo-ingest.Makefile
joeflack4: Btw, the reason I added the goals `component-download-mondo.owl` and `$(COMPONENTSDIR)/mondo.owl` is because it looks like they were required in order to resolve the dependency `$(REPORTDIR)/mirror_signature-mondo.tsv`.

src/scripts/migrate.py
```diff
@@ -0,0 +1,93 @@
...
+from typing import Dict, List
+
+import oaklib
+import pandas as pd
+
+
+# TODO: implement this func:
+# todo: IDs should be int or str? prolly str
+def determine_next_available_mondo_id(min_id: str, mondo_termlist_df: pd.DataFrame) -> str:
+    """Starting from `min_id`, count up and check until finding the next ID."""
```
joeflack4, line 17: @matentzn I changed what you wrote earlier:
> satrting from min_id, then counting up and checking if it does not already exist.

to: 
> Starting from `min_id`, count up and check until finding the next ID.

Does it sound like I have the right idea?

src/scripts/migrate.py
```diff
@@ -0,0 +1,93 @@
...
+
+    result = pd.DataFrame(data)
+    result.to_csv(outpath, sep="\t")
+    return result
+
+
+def cli():
+    """Command line interface."""
+    package_description = \
+        'Slurp pipeline: Integrate new terms from other ontologies into Mondo.'
```
joeflack4, line 70: @Nico: If any of these explanations feel inaccurate to you, let me know what you'd like the text to be.

src/ontology/mondo-ingest.Makefile
joeflack4: Thanks. That actually came up w/ [something](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html). I'll give a read. I googled something like "make what is the pipe | at end of dependencies", but didn't have any luck.

src/ontology/mondo-ingest.Makefile
matentzn: Always use `$(REPORTDIR)`! Whenever you see the other, you can think "Nico was sloppy again".

src/ontology/mondo-ingest.Makefile
matentzn: This happens when make cannot figure out at least one dependency, transitively. "Figure out" means, there is at least one dependency somewhere in the depedency chain for which neither a file, nor a make target exists.. If you cant figure it out I can debug this for you, just say the word!

src/ontology/mondo-ingest.Makefile
matentzn: This was a silly design choice by my past self, sorry about that. Changing it now will cause too much churn. 

I may change this make pipeline a tiny bit during our next call, but for now, if it works, it looks ok.

src/ontology/mondo-ingest.Makefile
matentzn: > I'm curious. Is http://purl.obolibrary.org/obo/mondo.owl updated at the same time as the [latest release](https://github.com/monarch-initiative/mondo/releases/latest)?

Yes

> This url (http://purl.obolibrary.org/obo/mondo.owl) is now used by two goals: tmp/mondo.owl and component-download-mondo.owl. Should we make a variable for it and share that variable?

Dont worry I may refactor this. I would prefer if only `tmp/mondo.owl` existed.

src/ontology/mondo-ingest.Makefile
joeflack4: I understand how this works now and why it's there. I read: http://gnu.org/software/make/manual/html_node/Prerequisite-Types.html

I think it's saying something different than what you're saying, though. It says: 
> Occasionally, however, you have a situation where you want to impose a specific ordering on the rules to be invoked without forcing the target to be updated if one of those rules is executed. In that case, you want to define order-only prerequisites.

Also I feel like "order-only prerequisite" is a very strange name for this. Don't understand what 'order' has to do with it; I would have called the ones on the left of `|` something like "target update prerequisites" and the right "non target update prerequisites" or "target update independent prerequisites".

src/ontology/mondo-ingest.Makefile
joeflack4: Ok, that all sounds fine to me! I'll leave it up to you when to change it too.

src/ontology/mondo-ingest.Makefile
joeflack4: Thanks for these responses.

Sounds good to me. I can guess why. I'll leave the refactoring to you!

src/ontology/mondo-ingest.Makefile
joeflack4: You mentioned that you will probably refactor away `component-download-mondo.owl`. It might be good for me to know at some point how to make updates that will affect `src/ontology/Makefile` for future cases though.

src/ontology/mondo-ingest.Makefile
joeflack4: "the word!". That is a tiny pain I have though; no "debugging" in `make`, so to speak; just troubleshooting.

What I would expect though is that the resolution would look like this:
goal `slurp-omim` (`slurp-%`) -> prerequisite `slurp/omim.tsv` (`slurp/%.tsv`) -> goal `slurp/%.tsv: $(COMPONENTSDIR)/%.owl $(TMPDIR)/mondo.sssom.tsv $(REPORTDIR)/mirror_signature-mondo.tsv | slurp/`

But for some reason, this doesn't happen.

src/ontology/mondo-ingest.Makefile
joeflack4: haha. ok that sounds fine

src/ontology/mondo-ingest.Makefile
```diff
@@ -275,3 +274,32 @@ mappings/mondo-sources-all-lexical.sssom.tsv: $(SCRIPTSDIR)/match-mondo-sources-
...
+	python $(SCRIPTSDIR)/migrate.py \
+	--ontology-path $(COMPONENTSDIR)/%.owl \
+	--sssom-map-path $(TMPDIR)/mondo.sssom.tsv \
+	--min-id 123000 \
+	--mondo-terms-path $(REPORTDIR)/mirror_signature-mondo.tsv \
+	--outpath $@
+
+slurp-%: slurp/%.tsv
+
+slurp-all: slurp-omim slurp-doid slurp-ncit slurp-ordo slurp-icd10cm slurp-icd10who
```
joeflack4, line 305: Added to `.PHONY`
`slurp-all`: I appended this to `.PHONY` at the top of the file. I think this is correct, even though ultimately it does end up creating files; just none with the name "slurp-all".
`slurp-%`: I wasn't going to append this to `.PHONY`, but then I realized that he same logic for "slurp-all" should apply here.

src/ontology/mondo-ingest.Makefile
```diff
@@ -275,3 +274,32 @@ mappings/mondo-sources-all-lexical.sssom.tsv: $(SCRIPTSDIR)/match-mondo-sources-
...
+# TODO: Move this to Makefile. how, when I shouldn't edit it from this repo?
+$(COMPONENTSDIR)/mondo.owl: component-download-mondo.owl
+	if [ $(COMP) = true ] ; then if cmp -s $(TMPDIR)/component-download-mondo.owl.owl $@ ; then echo "Component identical."; else echo "Component is different, updating." && cp $(TMPDIR)/component-download-mondo.owl.owl $@; fi; fi
+
+# min-id: the next available Mondo ID
+slurp/%.tsv: $(COMPONENTSDIR)/%.owl $(TMPDIR)/mondo.sssom.tsv $(REPORTDIR)/mirror_signature-mondo.tsv | slurp/
+	python $(SCRIPTSDIR)/migrate.py \
+	--ontology-path $(COMPONENTSDIR)/%.owl \
+	--sssom-map-path $(TMPDIR)/mondo.sssom.tsv \
+	--min-id 123000 \
```
joeflack4, line 299: @matentzn How did you ascertain that `123000` was the next `min_id`. I imagine that this will be outdated soon. Won't we want some way to programmatically do this? How best to do that? 

**Possible solutions**
(a) Have the Python script figure it out, (b) have some other `make` goal that creates a file with this number in it, (c) have `update_repo` add a variable to `Makefile` for this?, (d) Update some `make` goal to insert this number into some type of Mondo config `.yml` somewhere?

I'm leaning towards (d).

src/ontology/mondo-ingest.Makefile
```diff
@@ -250,6 +253,7 @@ $(REPORTDIR)/mondo_ordo_unsupported_subclass.tsv: ../sparql/mondo-ordo-unsupport
 .PHONY: mondo-ordo-subclass
 mondo-ordo-subclass: $(REPORTDIR)/mondo_ordo_unsupported_subclass.tsv
 
+# TODO: I believe we'll need a special case/goal for `reports/mirror_signature-mondo.tsv`, which will depend on `tmp/mondo.owl` instead.
```
joeflack4, line 256: Regarding `reports/mirror_signature-%.tsv`:
> TODO: I believe we'll need a special case/goal for `reports/mirror_signature-mondo.tsv`, which will depend on `tmp/mondo.owl` instead.

src/scripts/migrate.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+# todo: temp links:
+# https://incatools.github.io/ontology-access-kit/
+# https://incatools.github.io/ontology-access-kit/intro/tutorial02.html
+def run(ontology_path: str, sssom_map_path: str, min_id: str, mondo_terms_path: str, outpath: str) -> pd.DataFrame:
+    """Run slurp pipeline for given ontology"""
+    # Read source files
+    # todo: `ontology`: If trouble, can try `SparqlImplementation`, but ~6 min to load and queries slow (cuz rdflib)
+    ontology = ProntoImplementation(OntologyResource(slug=ontology_path, local=True))
```
joeflack4, line 70: ## Success cases
I tried several other implementations first, but `ProntoImplementation`, which is what the documentation examples use, worked the best. `SparqlImplementation` mostly worked too, but it was _very_ slow (because of `rdflib`).

I experienced a lot of issues trying out the various implementations, and I created/updated some OAK GitHub issues.

## Failed cases
### `SqlImplementation`
I encountered 4 distinct issues, but they all seemed related to actually loading the ontology. I think it was related to an existing issue you opened, so I updated that: https://github.com/INCATools/ontology-access-kit/issues/90#issuecomment-1205957717

### `FunOwlImplementation`
I got no entities back when I queried for them: https://github.com/INCATools/ontology-access-kit/issues/218

## Other attempts
I didn't spend much time on these but I wanted to give them a shot.
- `ObographImplementation`: `NotImplementedError`
- `Bioportal`: `NotImplementedError`
- `BasicOntologyInterface`: `NotImplementedError` (I guess not surprising; it's an interface)

---

I looked at [the CLI](https://incatools.github.io/ontology-access-kit/cli.html) and I think it infers what implementation you need based on what you pass. Doesn't seem to be explicit. Maybe it is easier to use, but I'd prefer to use Python directly if I can.

src/scripts/migrate.py
```diff
@@ -0,0 +1,141 @@
...
+
+def _get_all_term_curies(ontology: ProntoImplementation, is_omim=False) -> List[str]:
+    """Get all terms as CURIEs
+    todo: Will we need prefix_map from our config `.yml`s? Or can OAK handle automatically?
+     - reasons we might need: (i) filter correct terms, (ii) convert URI to CURIE if get URI
+     """
+    # todo: if oak can't handle automatically, I think we'll need `prefix_map`:
+    terms = [x for x in ontology.entities()]
+    # todo: temp until I figure out if (a) we need `prefix_map`, or (b) OAK can somehow handle automatically:
+    if is_omim:
```
joeflack4, line 52: @matentzn This is temporary. I noticed that our `omim.owl` doesn't have these prefixes: #55

src/scripts/migrate.py
```diff
@@ -0,0 +1,141 @@
...
+        # print(f'  {rel} ! {ontology.get_label_by_curie(rel)}')
+        for parent in parents:
+            # print(f'    {parent} ! {ontology.get_label_by_curie(parent)}')
+            p.append(parent)
+    return p
+
+
+def _get_all_term_curies(ontology: ProntoImplementation, is_omim=False) -> List[str]:
+    """Get all terms as CURIEs
+    todo: Will we need prefix_map from our config `.yml`s? Or can OAK handle automatically?
```
joeflack4, line 46: @matentzn Do we need to add our `metadata/%.yml` files as a parameter to this script, in order to get `base_prefix_map`? That way we can get the list of terms that are 'owned' by that ontology? I think we do.

src/scripts/migrate.py
```diff
@@ -0,0 +1,142 @@
...
+    p = []
+    for rel, parents in rels.items():
+        # print(f'  {rel} ! {ontology.get_label_by_curie(rel)}')
+        for parent in parents:
+            # print(f'    {parent} ! {ontology.get_label_by_curie(parent)}')
+            p.append(parent)
+    return p
+
+
+def _get_all_term_curies(ontology: ProntoImplementation, is_omim=False) -> List[str]:
```
joeflack4, line 44: Had some issues getting terms to show when running from makefile: #56

src/ontology/mondo-ingest.Makefile
```diff
@@ -157,13 +157,13 @@ metadata/mondo.sssom.config.yml:
...
 #################
-# Utils #########
+##### Utils #####
 #################
 # Documentation for `report-mapping-annotations` and `update-jinja-sparql-queries`: `docs/developer/ordo.md`
 # TODO: When https://github.com/monarch-initiative/mondo-ingest/issues/43 is fixed, can change back to `requirements.txt`
 python-install-dependencies:
 	python3 -m pip install --upgrade pip
-	python3 -m pip install -r $(RELEASEDIR)/requirements-unlocked.txt
+	python3 -m pip install --upgrade -r $(RELEASEDIR)/requirements-unlocked.txt
```
joeflack4, line 166: @matentzn This is necessary for OAK I think. But we may want to restrict this `--upgrade` to just a few cases. See: #57

src/scripts/migrate.py
```diff
@@ -0,0 +1,142 @@
+"""Slurp migration pipeline
```
joeflack4, line 1: Just linking this `icd10cm.owl` issue to this PR for provenance: #58

src/ontology/mondo-ingest.Makefile
matentzn: sounds good, I think this is needed for sssom as well..

src/ontology/mondo-ingest.Makefile
```diff
@@ -3,7 +3,7 @@
 ## If you need to customize your Makefile, make
 ## changes here rather than in the main Makefile
 .PHONY: deploy-mondo-ingest build-mondo-ingest documentation mappings update-jinja-sparql-queries \
-report-mapping-annotations python-install-dependencies all-exclusions
+report-mapping-annotations python-install-dependencies slurp-all slurp-%
```
matentzn, line 6: ahy does this remove `all-exclusions`?

src/ontology/mondo-ingest.Makefile
```diff
@@ -239,7 +242,7 @@ tmp/mondo.sssom.tsv:
 	wget http://purl.obolibrary.org/obo/mondo/mappings/mondo.sssom.tsv -O $@
 
 tmp/mondo.sssom.ttl: tmp/mondo.sssom.tsv
-	sssom convert $< -O rdf -o $@
+	sssom convert $< -O rdf -o $@tmp/mondo.ow
```
matentzn, line 245: st wrong here

src/ontology/mondo-ingest.Makefile
matentzn: No no, this is not the _next_ available. This is the _minimum_ id. So basically, your script must take the minimum ID and increment the number until an ID is found that is not already taking but the ontology (mirror-signature)

matentzn: (this is just so that no ID smaller than `min` is chosen, even if it is free.

src/ontology/mondo-ingest.Makefile
matentzn: Shall I take care of this?

src/ontology/mondo-ingest.Makefile
matentzn: Its correct!

src/scripts/migrate.py
matentzn: Yes!

src/scripts/migrate.py
matentzn: did you sort this?

src/scripts/migrate.py
matentzn: Why dont you simply pass in `mirror-signature-mondo` which has all the terms?

src/scripts/migrate.py
matentzn: THIS IS GREAT! Thanks for doing this thorough analysis, I highly highly appreciated. This is long-game stuff and really valuable!

matentzn: Go ahead as you see fit.

src/scripts/migrate.py
matentzn: Looks great so far!

src/ontology/mondo-ingest.Makefile
joeflack4: I just rebased this `slurpdraft` branch from `main` after merging the "all/merged exclusions" PR.

The new command for this is no longer the `.PHONY` target `all-exclusions`. The target is now called `config/term_exclusions.txt config/exclusion_reasons.robot.template.tsv:`.

src/ontology/mondo-ingest.Makefile
joeflack4: Yes, please.

src/ontology/mondo-ingest.Makefile
joeflack4: Ah, ok. I misunderstood. So this is not likely to change. And if it does change, would not be very often.

src/scripts/migrate.py
joeflack4: Thanks. It did take a fair amount of time to try these out. Glad that it was useful to do so. I'm going to likely pick `ProntoImplementation` or `SqlImplementation`. I do have some questions you might be able to answer regarding loading an ontology into a SQLite DB. I have a note to talk about that at one of our meetings.

src/scripts/migrate.py
joeflack4: Ah yes, I'm pretty sure this is fixed. If not, I'll bring this up again later.

src/ontology/mondo-ingest.Makefile
joeflack4: Glad you noticed that. Must have been a weird copy/paste error. Fixed.

src/scripts/migrate.py
```diff
@@ -0,0 +1,142 @@
...
+
+TODO's:
+  -
+"""
+import os.path
+from argparse import ArgumentParser
+from typing import Dict, List
+
+import pandas as pd
+import oaklib
```
joeflack4, line 11: @matentzn Had some issues with (i) dependencies for `SqlImplementation` and also (ii) things not working while no internet connection. Opened up some issues / PRs: 
[OAK: [PR](https://github.com/INCATools/ontology-access-kit/pull/232), issues ([1](https://github.com/INCATools/ontology-access-kit/issues/234)); relation-graph: [PR](https://github.com/balhoff/relation-graph/pull/146); rdftab: [PR](https://github.com/ontodev/rdftab.rs/pull/22); prefix_commons: issues([1](https://github.com/prefixcommons/prefixcommons-py/issues/25)); [linkml-runtime issue](https://github.com/linkml/linkml-runtime/issues/198)]

.gitignore
```diff
@@ -62,6 +63,7 @@ src/ontology/imports/*_terms_combined.txt
 src/ontology/components/*.owl
 src/ontology/components/*.json
 src/ontology/components/*_terms_combined.txt
+src/ontology/components/*relation-graph*
```
joeflack4, line 66: @matentzn Should we just ignore all of `src/ontology/components/*`?

src/scripts/migrate.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def _get_all_term_curies(ontology: ProntoImplementation, is_omim=False) -> List[str]:
+    """Get all terms as CURIEs
+    todo: Will we need prefix_map from our config `.yml`s? Or can OAK handle automatically?
+     - reasons we might need: (i) filter correct terms, (ii) convert URI to CURIE if get URI.
+     ...Then rename function to something like _get_owned_term_curies()
+     """
+    # todo: if oak can't handle automatically, I think we'll need `prefix_map`:
+    terms = [x for x in ontology.entities()]
```
joeflack4, line 53: @matentzn So far `ProntoImplementation` has been the most stable. I might go back to that. I was able to finally get `SqlImplementation` working, but I got back significantly less members from it than when using `ProntoImplementation`: https://github.com/INCATools/ontology-access-kit/issues/235

src/scripts/migrate.py
joeflack4: @matentzn Sorry for my super-late response here. I'm not trying to get Mondo terms; I'm trying to get the terms from ontologies we are slurping.

Is `reports/mirror_signature-mondo.tsv` supposed to have all of the terms for all ontologies we've imported? I did not think that was the case. If it is supposed to be the case, then we have a problem. I just checked, and I don't see anything for `omim` or `ordo`, `icd10cm`, or most of the ontologies we are slurping. I do see some NCIT terms in there, though.

In addition, if the goal of slurping that we are supposed to be importing new terms that _aren't yet in Mondo_, then it doesn't make sense to be querying this file for those terms.

src/ontology/slurp/omim.tsv
```diff
@@ -0,0 +1 @@
+""
```
joeflack4, line 1: @matentzn There's nothing in `slurp/` yet, but I was wondering; do you want this dir and the slurp outputs that get put in here to be committed, or should I add this dir to the `.gitignore`?

src/scripts/migrate.py
```diff
@@ -0,0 +1,189 @@
...
+
+# TODO: implement this func
+# todo: IDs should be int or str? prolly str
+def determine_next_available_mondo_id(min_id: str, mondo_termlist_df: pd.DataFrame) -> str:
+    """Starting from `min_id`, count up and check until finding the next ID."""
+    next_id = str(0)
+    return next_id
+
+
+def _get_direct_parents(ontology: Union[ProntoImplementation, SqlImplementation], curie: str):
```
joeflack4, line 30: @matentzn Having some trouble getting both `ProntoImplementation` and `SqlImplementation` to return relationships (parents or otherwise) of a given term by its URI or CURIE. I created some issues:
- https://github.com/INCATools/ontology-access-kit/issues/237
- https://github.com/INCATools/ontology-access-kit/issues/238

Fortunately, I have a temporary solution that should work until one of these is fixed.

.gitignore
matentzn: makes sense

requirements.txt
```diff
@@ -62,7 +62,8 @@ myst-parser==0.18.0
 networkx==2.8
 numpy==1.22.3
 nxontology==0.4.1
-oaklib==0.1.34
+oaklib==0.1.36
+ontoportal-client==0.0.3
```
matentzn, line 66: why is this needed?

src/ontology/slurp/omim.tsv
matentzn: Yes, please commit them.

matentzn (commented):
I didnt look in too much depth yet, but

1. AWESOME thank you thank you for being so involved with OAK and pathing the way for a better future for all of us! I notice and highly appreciate the diligence with which you make issues everywhere. Thank you.
2. I have pushed some makefile changes to `main`. You should be using the `obolibrary/odkfull:dev` image from now on, for example: `ODK_TAG=dev sh run.sh make ...`. This is regularly updated and should have all the latest tools, unless newer than 1 week.

requirements.txt
joeflack4: @matentzn I have no idea. The requirements that I'm actually adding are in `requirements-unlocked.txt`. I just `pip -r install requirements-unlocked.txt; pip freeze > requirements.txt`. I imagine that `ontoportal-client` is a transitive dependency of a this newest updated version of OAK.

src/ontology/mondo-ingest.Makefile
joeflack4: Note to self: This sub-discussion is from 17 days ago. I can't remember if we agreed on these `.PHONY` targets before or after that, but either way, I remember that the last time we spoke about this, we agreed to have these phonies; so I need to add this back.

@matentzn I know this is not related to slurp, but I'll make this change either in this PR or directly to `main`. My question to you is: How many `.PHONY` targets do you want for the exclusion tables?

We have 3 distinct sets of artefacts that we're generating w/ the exclusion tables:
1. `$(REPORTDIR)/term_exclusions.txt $(REPORTDIR)/exclusion_reasons.robot.template.tsv`
2. `$(REPORTDIR)/term_exclusions.ttl`
3. `$(REPORTDIR)/excluded_terms_in_mondo_xrefs.tsv`

Which of these 3 do you want PHONY targets for? 1 for each? Or should I make one PHONY for (1-2) above combined, and a second one for (3)? And then perhaps an additional third target that runs (1-3)?

matentzn: i dont care so much about phony targets. A phony target is like a workflow. We only need it if we expect to run the task repeatedly. PHNOY goals should never be dependencies of other goals.

TLDR: as few phnonies as possible. But at least one for the whole workflow.

requirements.txt
matentzn: ok its probably an oak depedndency

src/ontology/mondo-ingest.Makefile
joeflack4: I understand this is a bit of a sidetrack, but I'd still like to know some things.

> "at least one for the whole workflow"
Does "the whole workflow" mean all of `term_exclusions.txt`? And then don't worry about the `.ttl`? Or does it include the `.ttl`? Does it include doing the whole problematic xrefs analysis? Or are those separate workflows? If by "whole workflow" you mean (1-3) above, that's easy enough!

> PHONY goals should never be dependencies of other goals.
Not necessarily. I know of one exception. My Python scripts need the `pip` requirements to be installed. So I have made my phony `python-install-dependencies` a dependency of all of the make goals that use Python.

matentzn: > Not necessarily. I know of one exception. My Python scripts need the pip requirements to be installed. So I have made my phony python-install-dependencies a dependency of all of the make goals that use Python.

It is absolutely critical that your dependencies are at all times added to ODK the same moment they are added here (make PRs there). The `python-install-dependencies` should only be used at development time - after that, the ODK should take over..

> Does "the whole workflow" mean...

I could tell you my opinion for this specific case (one workflow to run everything (1-3 for all ontologies with one command), one workflow per-ontology), but the point is that there is no one correct answer to this question. Anything that can be usefully defined by a single command can be considered a workflow, and you will have to apply some judgement.

src/ontology/mondo-ingest.Makefile
matentzn: This is where you do it:

https://github.com/INCATools/ontology-development-kit/blob/master/template/src/ontology/Makefile.jinja2

But I would suggest you make issues before you make PRs, just to save you time for PR revisions.

src/ontology/mondo-ingest.Makefile
matentzn: Ok, I didnt test it, but I fixed some issues I could see by sight in https://github.com/monarch-initiative/mondo-ingest/commit/72c3299b9b123097baa689d0cf9c4aa0b06988db

See commit message for details. Also read https://stackoverflow.com/questions/7887343/makefile-wildcard-static-rule-with-phony

src/ontology/mondo-ingest.Makefile
matentzn: This issue had a bug but i fixed it in the last commit

joeflack4: ahh OK, thanks

src/ontology/slurp/omim.tsv
joeflack4: will do!

src/ontology/mondo-ingest.Makefile
joeflack4: I read the commit message:

> 2. PHONY goals cannot have wildcard dependencies

I see! Now I understand why `slurp-omim` wasn't working!. Seems like poor language design UX, IMHO. Very good to know this.

@matentzn Marking this resolved. Just tagging you so you are aware that I understand.

src/ontology/mondo-ingest.Makefile
```diff
@@ -315,11 +312,50 @@ signature_reports: $(ALL_MIRROR_SIGNTAURE_REPORTS) $(ALL_COMPONENT_SIGNTAURE_REP
...
+slurp/%.tsv: $(COMPONENTSDIR)/%.owl $(TMPDIR)/mondo.sssom.tsv $(REPORTDIR)/mirror_signature-mondo.tsv | slurp/
+	python $(SCRIPTSDIR)/migrate.py \
+	--ontology-path $(COMPONENTSDIR)/$*.owl \
+	--sssom-map-path $(TMPDIR)/mondo.sssom.tsv \
+	--onto-config-path metadata/$*.yml \
+	--min-id 123000 \
+	--mondo-terms-path $(REPORTDIR)/mirror_signature-mondo.tsv \
+	--outpath $@
+
+slurp-%:
```
joeflack4, line 358: @matentzn Just wanted to respond to:
>     3. Indentation is critical with Make targets, your `# TODO: Temporary debugging:` was wrongly indented to the level of the target name, not the body.

I know this is a trivial matter; sorry if trying to close the loop on this is a bit much.

Just for reference, since I no longer needed the comment, the change you made was from this:
```make
slurp-%:
# Temporary debugging
```

to this:
```make
slurp-%:
	# Temporary debugging
```

I assume you made this change because you thought either (a) my intention was for the comment to print to the console, or (b) this would cause a syntactical issue.

Regarding (b), I didn't run into any issues with not indenting. It seems to be syntactically correct.
Regarding (a), it didn't actually matter to me if it printed to the console or not. Sometimes I do have a preference. But usually if I want a comment printed to the console, I use `@echo`.

Honestly the main, semi-trivial, reason for my lack of indent in these temporary comment cases is syntax highlighting in PyCharm. Helps me see it so that I can remove it later:
Screen Shot 2022-08-31 at 9 50 53 AM

Screen Shot 2022-08-31 at 9 51 25 AM

matentzn: Ok, just be a bit of caution, for example:

```
l1:
	# Comment
	echo "hi"

l2:
# Comment
	echo "hi 2"

l3:
	echo "hi 3.1"
# Comment
	echo "hi 3.2"
```

You are right, all three work. Scratch that, do what you like!

```

File: src/ontology/mondo-ingest.Makefile (hunk 18 above)
Comment (joeflack4) [reply src/ontology/mondo-ingest.Makefile]: #### depending on phony goals
I saw your comment in the commit message:

> 1. Never depend on a PHONY goal, because they force the whole pipeline to be rebuild - just add them to each goal, installing a few python dependencies that are already installed is fast enough.

This is very good for me to know, thanks. It makes perfect sense; should've thought of this. I'll put the running of PHONY goals within the block of execution for the target in the future; not as a prerequisite.

---

@matentzn Just so you know though, I think it is wrong that this forces the _whole pipeline_ to rebuild. What seems to happen is that it just forces the target itself to be rebuilt. I did a test:

```
.PHONY: xxx
xxx:
	@echo hi

tempy.txt: mirror/omim.owl xxx
	touch tempy.txt
```

The results of this: When I ran `make tempy.txt`, it did indeed force a re-run of the target. It did `touch tempy.txt`. However, it did not re-run `mirror/omim.owl`. It considered that up-to-date.

Comment (matentzn) [reply src/ontology/mondo-ingest.Makefile]: You need to use `obolibrary/odkfull:dev`, I will have a new version up tomorrow (but the one up is fairly up to date).

Comment (joeflack4) [reply src/ontology/mondo-ingest.Makefile]: ### Phony targets for workflows
Got it, so I can use my discretion here. Just a warning about giving me the liberty to create PHONY targets as workflow aliases. I have been known to be quite liberal in creating them for my own convenience. I can try to use my judgement to only write ones if I think they'll be useful for others, but I might be a bit biased to my preferences.

File: src/ontology/mondo-ingest.Makefile (hunk 2 above)
Comment (joeflack4) [reply src/ontology/mondo-ingest.Makefile]: Done and done!
- Issue: https://github.com/INCATools/ontology-development-kit/issues/677
- PR: https://github.com/INCATools/ontology-development-kit/pull/676

Comment (matentzn) [reply src/ontology/mondo-ingest.Makefile]: Sorry, my answer was generic: if you want to _ever_ change the `Makefile` logic, that is the file you need to make a pull request on. 

for now, keep this goals in mondo-ingest.Makefile. If we wanted to add them to Makefile, we would have to add `Mondo` as a source to the `mondo-ingest`, which may make some sense.

Comment (joeflack4) [reply src/ontology/mondo-ingest.Makefile]: OK. I think we don't need to bring this up at the meeting. Everything makes sense now.

FYI we had another, second Mondo component related target that was commented out. I think we don't need it anymore, so I'm gonna delete:
```
.PHONY: component-download-mondo.owl
component-download-mondo.owl: | $(TMPDIR)
	if [ $(MIR) = true ] && [ $(COMP) = true ]; then $(ROBOT) merge -I http://purl.obolibrary.org/obo/mondo.owl \
	annotate --ontology-iri $(ONTBASE)/$@ $(ANNOTATE_ONTOLOGY_VERSION) -o $(TMPDIR)/$@.owl; fi

# TODO: (a) Move this to Makefile, or (b) refactor this away, or (c) DELETE this goal.
# ...I think (c) is most likely, as `reports/mirror_signature-%.tsv` needs `component-download-%.owl`.
# ...`reports/component_signature-%.tsv` is what needs `components/%.owl`, but that shouldn't be needed here.
# $(COMPONENTSDIR)/mondo.owl: component-download-mondo.owl
# 	if [ $(COMP) = true ] ; then if cmp -s $(TMPDIR)/component-download-mondo.owl.owl $@ ; then echo "Component identical."; else echo "Component is different, updating." && cp $(TMPDIR)/component-download-mondo.owl.owl $@; fi; fi

.gitignore
```diff
@@ -29,7 +30,7 @@ src/ontology/reports/*
 !src/ontology/reports/README.md
 !src/ontology/reports/*excluded_terms_in_mondo_xrefs.tsv
 !src/ontology/reports/*exclusion_reasons.robot.template.tsv
-!src/ontology/reports/*term_exclusions.txt
+!src/ontology/reports/*excluded_terms.txt
```
joeflack4, line 33: @sabrinatoro @nicolevasilevsky FYI, I just did a rename of `reports/term_exclusions.txt` to `reports/excluded_terms.txt`. This is for alphabetical consistency w/ the other aggregated reports: `exclusion_reasons.robot.template.tsv`, `excluded_terms_in_mondo_xrefs.tsv`. I updated the `reports/README.md`, etc, to match the name change.

@matentzn FYI just adding this small change to the slurp PR, but I can make this commit directly to `main` or do a separate PR if you prefer.

src/ontology/mondo-ingest.Makefile
joeflack4: Will do.

I wasn't actually sure if `l3` would work. Good to know! Gonna mark this one resolved now.

src/scripts/migrate.py
joeflack4: `ProntoImplementation` seems like it is more correct with the results it is returning and has more working methods available than `SqlImplementation`, so at this point I am using that.

src/scripts/migrate.py
```diff
@@ -0,0 +1,201 @@
...
+    #  i. If trouble, can try `SparqlImplementation`, but ~6 min to load and queries slow (cuz rdflib)
+    #  ii. use Sql > Pronto when .entities() fixed: https://github.com/INCATools/ontology-access-kit/issues/235
+    with open(onto_config_path, 'r') as stream:
+        onto_config = yaml.safe_load(stream)
+    # ontology = SqlImplementation(OntologyResource(slug=ontology_path, local=True))
+    ontology = ProntoImplementation(OntologyResource(slug=ontology_path, local=True))
+    sssom_df = pd.read_csv(sssom_map_path, comment='#', sep='\t')
+    mondo_termlist_df = pd.read_csv(mondo_terms_path, comment='#', sep='\t')
+
+    # Get relationships
```
joeflack4, line 109: FYI: This is how I determine the direct parents of a term, by first using OAK to get relationships and then building `    term_parent_map`.

src/scripts/migrate.py
```diff
@@ -0,0 +1,201 @@
...
+            new_triple += (curie,) if curie else (uri_or_curie,)
+        edge_triples_curies.append(new_triple)
+    # - Build lookup dict of all the direct parents of each term. Only includes direct parents owned by the ontology.
+    term_parent_map = {}
+    owned_prefixes = set(onto_config['base_prefix_map'].keys())
+    for triple in edge_triples_curies:
+        subject, predicate, object = triple
+        if subject not in term_parent_map:
+            term_parent_map[subject] = []
+        if predicate == 'rdfs:subClassOf' and object.split(':')[0] in owned_prefixes:
```
joeflack4, line 128: @matentzn In doing my first run of this, I observed cases where parents of a term are defined outside the ontology itself. For example, in OMIM, many terms are `rdfs:subClassOf` [`SO:0000704`](http://www.sequenceontology.org/browser/current_svn/term/SO:0000704).

I think that the intention of slurp is that it only cares about terms that are owned by the given ontology, so that is how I have designed this. E.g. `SO:0000704` may be a direct parent of some OMIM terms, but it is not a slurp candidate.

I think this is correct, but if I am wrong please let me know.

src/scripts/migrate.py
joeflack4: Actually, not doing it this way. Blah, all that work (not too much work) and I realized that actually, `[x for x in ProntoImplementation().relationships(subjects=[URI])]` will work in this case. I didn't notice before, because I was looking at the methods `ProntoImplementation().incoming_relationship_map` and `ProntoImplementation().outoing_relationship_map` which are broken right now. And I didn't notice `.relationships()`, which is defined in a parent class.

src/ontology/slurp/doid.tsv
```diff
@@ -0,0 +1 @@
+	mondo_id	xref	label	definition	qualified_parents
```
joeflack4, line 1: @matentzn All the output files being generated are empty. I imagine this is not expected. Perhaps we should look over the algorithm and inputs during our meeting, or live debug it.

src/scripts/migrate_utils.py
```diff
@@ -0,0 +1,257 @@
...
+            uri = converter.expand(self.curie)
+            self.uri = uri if uri != self.curie else None
+
+        # Set: label, definition
+        # todo: When OAK supports: I feel like this would be faster to do outside the Term class itself
+        #  ...e.g. build a dict of CURIE-> Term, then use use ontology.labels() on all curies and set label
+        #  ...Right now, OAK ProntoImplementation labels() is just proxy for label(), though SqlImplementation has, and
+        #  ...Both of them have definition() but not definitions().
+        if ontology:
+            self.label = ontology.label(self.curie)
```
joeflack4, line 121: ### Label issues
@matentzn Only having success getting labels for: DOID, NCIT. Getting 100% of them. FOR OMIM, ORDO, ICD10CM, and ICD10WHO, getting 0% of labels.

I created an issue in OAK: https://github.com/INCATools/ontology-access-kit/issues/276

### Definition issues
The line of code which this comment is on is where I'm getting the label. The very next line is where I'm getting the definition.

I'm only getting definitions for DOID, and only for about 11% of them. I see that they are defined by `IAO_0000115` on class annotations. From what I could see, none of the other ontologies had `IAO_0000115` present on any annotations. Perhaps this is why no definitions?

---

@matentzn How shall I proceed?
a. **Wait until OAK fixes**, assuming getting the labels and definitions here are not critical for `slurp`.
b. **I can try to contribute to OAK to fix the issue**. I know about where to do the work and have some idea of how to do it, but this would probably take time and hold us up.
c. **SPARQL**. As I mentioned, I can use SPARQL for now. However, if these are not a big deal for `slurp`, I would prefer (a) in the interest of time, or at least finish up the `slurp` and start using it, and then I can immediately start working on a ticket to add labels/definitions via SPARQL.

src/scripts/migrate.py
```diff
@@ -0,0 +1,143 @@
...
+        onto_config = yaml.safe_load(stream)
+        owned_prefix_map: Dict[PREFIX, URI] = onto_config['base_prefix_map']
+    sssom_df: pd.DataFrame = pd.read_csv(sssom_map_path, comment='#', sep='\t')
+    mondo_term_ids: Set[int] = get_mondo_term_ids(mondo_terms_path)
+
+    # Intermediates
+    owned_terms: List[Term] = _get_all_owned_terms(
+        ontology=ontology, owned_prefix_map=owned_prefix_map, cache_dir_path=CACHE_DIR,
+        onto_config_path=onto_config_path)
+    sssom_object_ids: Set[Union[URI, CURIE]] = set(sssom_df['object_id'])  # Usually CURIE, but spec allows URI
```
joeflack4, line 68: FYI: `sssom_object_ids`: I set the type to `Set[Union[URI, CURIE]]` and allowed the code to handle this.  it would be easier if just CURIE only, but the SSSOM spec says that the type is `RDF ref`, which i think allows for either a CURIE or URI.

src/scripts/migrate.py
```diff
@@ -0,0 +1,143 @@
...
+from argparse import ArgumentParser
+from typing import Dict, List, Set, Union
+
+import pandas as pd
+import yaml
+from oaklib.implementations import ProntoImplementation
+from oaklib.resource import OntologyResource
+from oaklib.types import CURIE, URI
+
+from migrate_utils import CACHE_DIR, PREFIX, Term, _get_all_owned_terms, _get_next_available_mondo_id, \
```
joeflack4, line 26: FYI: `migrate_utils.py`: I moved a lot of code into `migrate_utils`. This is basically all the stuff that allows us to write simpler code using a Python API that would be convenient. I think that when we have time, we might want to review this and see what we think we should actually suggest as changes to be incorporated into OAK, if possible.

src/scripts/migrate_utils.py
```diff
@@ -0,0 +1,257 @@
...
+TRIPLE = RELATIONSHIP
+SCRIPTS_DIR = os.path.dirname(os.path.realpath(__file__))
+PROJECT_DIR = os.path.join(SCRIPTS_DIR, '..', '..')
+ONTOLOGY_DIR = os.path.join(PROJECT_DIR, 'src', 'ontology')
+TEMP_DIR = os.path.join(ONTOLOGY_DIR, 'tmp')
+CACHE_DIR = TEMP_DIR
+
+
+# TODO: there are remaining todo's in this class
+class Term:
```
joeflack4, line 24: FYI: `Term`: I was going to change my code to use a list of `{'curie': ..., 'uri': ...}`, but I haven't been writing much OO recently, so I thought it would be nice to do this and see what you thought. The result is that not only is `Term` a class that has `curie` and `uri` properties, but it also has other properties and methods which allow for the kind of simplicity we were looking for from OAK.

src/scripts/migrate_utils.py
```diff
@@ -0,0 +1,257 @@
...
+    for x in mondo_term_uris:
+        x = x[1:] if x.startswith('<') else x
+        x = x[:-1] if x.endswith('>') else x
+        if not x.startswith(mondo_base_uri):
+            continue
+        mondo_ids.append(int(x.replace('http://purl.obolibrary.org/obo/MONDO_', '')))
+    return set(mondo_ids)
+
+
+def _get_direct_owned_parent_uris(
```
joeflack4, line 165: FYI: `_get_direct_owned_parents()`: For a given URI, returns the direct parents of a class that are 'owned'b y the given ontology.

This code could probably be copy/pasted into OAK with minimal changes if we want to add this feature to OAK (I do).

Some things I would add to this:
- accept both CURIE or URI
- return a CURIE or URI; defaults to whatever type was passed as the lookup term

src/scripts/migrate.py
```diff
@@ -0,0 +1,143 @@
...
+    # Read inputs
+    ontology: ProntoImplementation = _load_ontology(ontology_path, use_cache)
+    with open(onto_config_path, 'r') as stream:
+        onto_config = yaml.safe_load(stream)
+        owned_prefix_map: Dict[PREFIX, URI] = onto_config['base_prefix_map']
+    sssom_df: pd.DataFrame = pd.read_csv(sssom_map_path, comment='#', sep='\t')
+    mondo_term_ids: Set[int] = get_mondo_term_ids(mondo_terms_path)
+
+    # Intermediates
+    owned_terms: List[Term] = _get_all_owned_terms(
```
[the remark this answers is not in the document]
matentzn: Meeting: most of OAK uses curies/entityreferences as inputs and outputs. This issue can be resolved

src/ontology/run.sh
```diff
@@ -45,7 +45,7 @@ if [ -n "$USE_SINGULARITY" ]; then
         -W $WORK_DIR \
         docker://obolibrary/$ODK_IMAGE:$ODK_TAG $TIMECMD "$@"
 else
-    docker run -v $VOLUME_BIND -w $WORK_DIR -e ROBOT_JAVA_ARGS="$ODK_JAVA_OPTS" -e JAVA_OPTS="$ODK_JAVA_OPTS" --rm -ti obolibrary/$ODK_IMAGE:$ODK_TAG $TIMECMD "$@"
+    docker run -v $VOLUME_BIND -w $WORK_DIR -e ROBOT_JAVA_ARGS="$ODK_JAVA_OPTS" -e JAVA_OPTS="$ODK_JAVA_OPTS" -ti obolibrary/$ODK_IMAGE:$ODK_TAG $TIMECMD "$@"
```
joeflack4, line 48: @hrshdhgd We made a change here (removed `--rm`) that should allow our local python packages within the container to retain additions / updates during `pip install`s

src/ontology/slurp/ordo.tsv
```diff
@@ -0,0 +1,329 @@
+mondo_id	xref	label	definition	parents
```
joeflack4, line 1: @matentzn Since our recent change, I'm getting outputs now. Some of them have ~13k rows. Is that expected?

I'm still running on some of the ontologies. Will push those files as well soon.

I'm not getting any parents that I'm able to output into the `parents` column that meet the condition (at least that I thought you were asking for). I'll bring this up at the mondo meeting tomorrow.

src/scripts/migrate_utils.py
joeflack4: Label issues solved. Not sure why I wasn't getting any. I'm getting them now...

We made an issue to deal w/ the lack of `IAO_0000115 `

src/ontology/mondo-ingest.Makefile
joeflack4: Probably was fine to delete `$(COMPONENTSDIR)/mondo.owl:`. Gonna mark this resolved.

src/scripts/utils.py
```diff
@@ -0,0 +1,280 @@
...
+        if use_cache:
+            with open(cache_path_ontology, 'wb') as f:
+                pickle.dump(ontology, f, protocol=pickle.HIGHEST_PROTOCOL)
+
+    return ontology
+
+
+# TODO: implement this func
+# todo: IDs should be int or str? prolly str
+def _get_next_available_mondo_id(min_id: Union[str, int], mondo_ids: Set[int]) -> (int, Set[int]):
```
joeflack4, line 157: @matentzn Just want to check that I'm implementing this function as you intended:
```py
    next_id = min_id
    while True:
        next_id = next_id + 1
        if next_id not in mondo_ids:
            break
    mondo_ids.add(next_id)
    return next_id, mondo_ids
```

I didn't get too clear on `min_id`. Do I even need to check if it's in `mondo_ids`?

I noticed a bug in this, but I fixed it. I have to re-runt he outputs though. Right now in the new mondo IDs in the outputs have gaps. I was incrementing the ID by 2 instead of 1.

My main concern is this: If I run this on all the ontologies, then there will be overlap in `mondo_id` between all of the output TSV files. They're being assigned the same IDs. If this is an issue, then when I'm running the slurp script, I can check for any `slurp/*.tsv` files and read those IDs in as well and make sure I don't assign IDs that have already been assigned there.

src/ontology/slurp/ordo.tsv
matentzn: > I'm not getting any parents in any output.

Investigating this is the priority. The code looks ok - unfortunately you will have to debug :/ Sorry for not being helpful on this one!

src/scripts/utils.py
```diff
@@ -0,0 +1,299 @@
...
+    # These vars are here for stability reasons, just in case I get CURIES where I expect URIs or vice versa.
+    subclass_preds = [x + 'subClassOf' for x in ['rdfs:', 'http://www.w3.org/2000/01/rdf-schema#']]
+    owned_prefixes = set(owned_prefix_map.keys())
+    uri_converter = curies.Converter.from_prefix_map(owned_prefix_map)
+
+    direct_owned_parent_uris: List[URI] = []
+    # todo: how to combine theses?
+    # todo: diff results? how ot chck?
+    rels1: List[CURIE] = ontology.hierararchical_parents(curie)
+    rels: List[TRIPLE] = [x for x in ontology.relationships(subjects=[curie])]
```
joeflack4, line 220: @matentzn Tried to do `ontology.hierararchical_parents(curie)` as you suggested, but got an error. I made a PR with a fix: https://github.com/INCATools/ontology-access-kit/pull/287

src/scripts/utils.py
joeflack4: Completed new `min_id`, `max_id` (w/ raised error) and reading from `slurp/` changes we discussed.

src/ontology/slurp/ordo.tsv
joeflack4: There _may_ be a bug in our logic, I'm not sure. 99%+ of NCIT term objects I instantiated were able to locate their parents using OAK, but when running through our slurp algorithm, no were no parents that met the condition `parent_curie in sssom_object_ids`. 

## Attempt 1: `oaklib.ProntoImplementation`
I think there's a bug. I haven't created an issue yet, though. As mentioned above, NCIT terms were able to locate their parents through `oak`. But this was not the case for the other ontologies.

```py
print(f'{os.path.basename(ontology_path)}: % terms that "have parents": ',
    len([x for x in unmapped_terms if x.direct_parent_curies]) / len(unmapped_terms))
```

Results:
```
doid.owl: % terms that "have parents":  0.0
omim.owl: % terms that "have parents":  0.0
ordo.owl: % terms that "have parents":  0.0
icd10who.owl: % terms that "have parents":  0.0
icd10cm.owl: % terms that "have parents":  0.0
ncit.owl: % terms that "have parents":  0.9979720960415315
```

## Attempt 2: `oaklib.SqlImplementation`
### 2.1: `.outgoing_relationship_map()`
Got `RecursionError`. Made issue: https://github.com/INCATools/ontology-access-kit/issues/290

### 2.2. `.outgoing_relationship_map()`
Haven't looked further into this, so I didn't make an issue.

```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such table: edge
[SQL: SELECT edge.subject AS edge_subject, edge.predicate AS edge_predicate, edge.object AS edge_object 
FROM edge 
WHERE edge.object = ?]
[parameters: ('Orphanet:101334',)]
(Background on this error at: https://sqlalche.me/e/14/e3q8)
```

### 2.3. `.relationships()`
Same error as (2.2): `sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such table: edge`

## Attempt 3: SPARQL
Trying now.

joeflack4: SPARQL success!
```tsv
mondo_id	xref	label	definition	parents
ID	A oboInOwl:hasDbXref	LABEL	A IAO:0000115	SC %
MONDO:850001	DOID:0040003	benzylpenicillin allergy		MONDO:0000794|MONDO:0000775|MONDO:0005271|MONDO:0005046|MONDO:0021199|MONDO:0000001
MONDO:850002	DOID:0040004	amoxicillin allergy		MONDO:0000794|MONDO:0000775|MONDO:0005271|MONDO:0005046|MONDO:0021199|MONDO:0000001
...
```

joeflack4: Everything here is fixed. I just need to move to OAK once it's fixed. We can resolve this thread now.

src/scripts/migrate.py
```diff
@@ -0,0 +1,139 @@
...
+        qualified_parents = True
+        parent_mondo_ids = []
+        # TODO: are these conditions correct? We're getting no outputs, though many terms w/ parents exist
+        for parent_curie in t.direct_owned_parent_curies:
+            # Conversely, if any of T.parents is unmapped, T is not to be slurped
+            if parent_curie in sssom_object_ids:
+                parent_df: pd.DataFrame = sssom_df[sssom_df['object_id'] == parent_curie]
+                parent_mondo_id = parent_df.to_dict('records')[0]['subject_id']
+                parent_mondo_ids.append(parent_mondo_id)
+            if parent_curie not in sssom_object_ids and parent_curie in owned_term_curies:
```
joeflack4, line 77: ## Not getting any outputs
I was getting `qualified_parents = True` for all terms at first, so my outputs were very large.

With not much debugging, I figured out why that was and fixed it.

In this block, I needed to change `owned_terms` to `owned_term_curies`.
```py
    qualified_parents = False
    break
```

The `owned_terms` was a list of my `Term` objects, not curies. That's why this condition was never being met. However after fixing it, I'm no longer getting any outputs. It seems that, for right now, for all ontologies we're slurping, for all terms in all those ontologies, there is at least 1 parent which has not been mapped, and thus we cannot slurp.

This is strange, though. If there are unmapped parents of a term, then those unmapped parents should be slurp candidates as well. So I would expect to at least see some terms higher up in the hierarchy that qualify as slurpable...

I can try debugging this further, but if I can get a concrete example of a simple term that is slurpable, that would really help. The example you gave me `DOID:0060091` didn't work because it is already mapped. I can probably manually look through the inputs and try to find an example myself, but it would save me a lot of time if you happen to know one that should work.

matentzn: Here is how you find a good candidate:

1. Print the unmapped terms
2. Pick a random one (T)
3. Look at the _direct_ parents P of that term T (in DOID, nearly all terms will have only a single parent)
4. If the P is in `sssom_object_ids`,  T should be slurped

The vast majority of unmapped terms should fulfill that criterion!

matentzn: Make sure the parent method only looks at direct isa-parents, not _all_ parents (non-isa, or indirect isa, i.e. grandparents)

joeflack4: Hmm, I do believe that (1-4) is what the algorithm is doing explicitly. But I will do this a bit manually and see if I find an error in my logic, or otherwise find some good concrete examples to test. I guess I could make a unit test using some smaller subsets of the files (e.g. a `doid.owl` with just a few classes; a SSSOM file with just a few rows, etc).

In regards to which types of parents, there are some OAK methods that get only subclass/isa relationships. Still somewhat broken, so just using SPARQL now, and my query is only getting `rdfs:subClassOf*`.

src/ontology/slurp/ordo.tsv
joeflack4: Solved! via `rdfs:subClassOf*` --> `rdfs:subClassOf`

src/scripts/migrate.py
joeflack4: Solved via `rdfs:subClassOf*` -> `rdfs:subClassOf`

src/scripts/utils.py
joeflack4: You can close this whenever you see it, as I'm now getting parents using SPARQL. Just want you to be aware.

src/scripts/migrate.py
```diff
@@ -0,0 +1,137 @@
...
+    parser = ArgumentParser(description=package_description)
+    parser.add_argument(
+        '-o', '--ontology-path', required=True,
+        help='Path to ontology file, e.g. an `.owl` file.')
+    parser.add_argument(
+        '-c', '--onto-config-path', required=True,
+        help='Path to a config `.yml` for the ontology which contains a `base_prefix_map` which contains a '
+             'list of prefixes owned by the ontology. Used to filter out terms.')
+    parser.add_argument(
+        '-e', '--onto-exclusions-path', required=True,
```
joeflack4, line 99: @matentzn FYI: New arg `--onto-exclusions-path`

src/scripts/utils.py
```diff
@@ -0,0 +1,399 @@
...
+            uri = converter.expand(self.curie)
+            self.uri = uri if uri != self.curie else None
+
+        # Set: label, definition
+        # todo: Assign in batch (when OAK supports): I feel like would be faster to do outside the Term class itself
+        #  ...e.g. build a dict of CURIE-> Term, then use use ontology.labels() on all curies and set label
+        #  ...Right now, OAK ProntoImplementation labels() is just proxy for label(), though SqlImplementation has, and
+        #  ...Both of them have definition() but not definitions().
+        if ontology:
+            self.label = ontology.label(self.uri)
```
joeflack4, line 127: @matentzn **Fixed 'labels' issue**
OAK was not returning labels for some ontology. Seems to be a `prefix_map` issue. Also, the type definitions for these `.label()` functions only say they support CURIE, but they also support URI, which is how I was able to fix this bug. I updated the [corresponding OAK issue](https://github.com/INCATools/ontology-access-kit/issues/212) with these new updates.

src/scripts/utils.py
```diff
@@ -0,0 +1,399 @@
...
+        self.definition = definition
+        self.ontology = ontology
+        self.owned_prefix_map = owned_prefix_map
+        self.general_prefix_map = general_prefix_map
+        if owned_prefix_map or general_prefix_map:
+            self._converter = curies.Converter.from_prefix_map(
+                general_prefix_map if general_prefix_map else owned_prefix_map)
+        self.uri = uri
+        self.curie = curie
+        self.direct_owned_parent_uris = direct_owned_parent_uris
```
joeflack4, line 60: @matentzn **Why OMIM parents are (correctly) not showing up**
I took a look, and it looks like the way we have `omim.owl` does not have any parents that are 'owned' by OMIM. I'm assuming this is not an issue with OMIM itself or our OMIM issue, and thus not an issue for us. But I want you to review and check to be sure.

Example OMIM class:
```owl
    <owl:Class rdf:about="https://omim.org/entry/100650">
        <rdfs:subClassOf rdf:resource="http://purl.obolibrary.org/obo/SO_0000704"/>
        <rdfs:subClassOf>
            <owl:Restriction>
                <owl:onProperty rdf:resource="http://purl.obolibrary.org/obo/RO_0002525"/>
                <owl:someValuesFrom rdf:resource="http://purl.obolibrary.org/obo/CHR_9606chr12q24.12"/>
            </owl:Restriction>
        </rdfs:subClassOf>
```

As you can see, it has parents, but they're from `SO`, `RO`, and `CHR`. As they are not 'owned' by the ontology, they will not appear int he slurp output.

src/ontology/slurp/icd10cm.tsv
```diff
@@ -0,0 +1,22 @@
+mondo_id	xref	label	definition	parents
+ID	A oboInOwl:hasDbXref	LABEL	A IAO:0000115	SC %
+MONDO:850001	ICD10CM:A00-B99	Certain infectious and parasitic diseases (A00-B99)
```
joeflack4, line 3: @matentzn **ICD10CM outputs are only grouping classes**
Example: `MONDO:850001	ICD10CM:A00-B99	Certain infectious and parasitic diseases (A00-B99)`

All of the other rows are like that one; grouping classes. There's one exception, though I guess you could call it a "top level grouping class":
`MONDO:850010	ICD10CM:ICD-10-CM	ICD-10-CM TABULAR LIST of DISEASES and INJURIES`

Should these indeed be included int he output? I'm not sure if these are indeed appropriate slurp candidates, but if they are not, then I suppose we want to add these grouping classes into the exclusion tables (and re-run the slurp)?

src/ontology/run.sh
matentzn, line 48:
```suggestion
    docker run -v $VOLUME_BIND -w $WORK_DIR -e ROBOT_JAVA_ARGS="$ODK_JAVA_OPTS" -e JAVA_OPTS="$ODK_JAVA_OPTS" --rm -ti obolibrary/$ODK_IMAGE:$ODK_TAG $TIMECMD "$@"
```

matentzn: No changes to run.sh on main branch please.

src/scripts/utils.py
joeflack4: We looked into this and decided that we will [add genes to the exclusion table](https://github.com/monarch-initiative/mondo-ingest/issues/82).

src/ontology/slurp/icd10cm.tsv
joeflack4: Nico said this is correct.

src/ontology/run.sh
joeflack4: Oh, ok. We made this change together. Thought this was something we were changing permanently in order to address the issue of Python packages not being updated within the container between runs. It didn't work anyway :(, was gonna bring it up at next check-in meeting. It's still on my list of things to discuss whenever we have time.

src/scripts/utils.py
joeflack4: Marking this resolved

src/scripts/exclusion_term_expansion.py
```diff
@@ -142,7 +143,7 @@ def expand_ontological_exclusions(
     # Variable names: kids0=Children excluded; kids1=Children included
     prefix_sparql_strings = [f'prefix {k}: <{v}>' for k, v in prefix_map.items()]
     df_kids1 = exclusions_df[exclusions_df['exclude_children'] == True]
-    df_kids0 = exclusions_df[exclusions_df['exclude_children'] == False]
+    df_kids0 = exclusions_df[exclusions_df['exclude_children'] != True]
```
joeflack4, line 146: @matentzn **Ambiguity of null `exclude_children`**
I fixed a bug that was causing me to get no results when I re-ran exclusion table for OMIM. However, I don't know if my fix is correct/preferable, so I opened an issue: https://github.com/monarch-initiative/mondo-ingest/issues/83

joeflack4: I was going to make a separate PR for the bugfix, but I just did it here. Probably woulda been best for me to just do it on `main` and rebase, but whatevs.

src/ontology/mondo-ingest.Makefile
```diff
@@ -218,12 +212,18 @@ $(REPORTDIR)/%_excluded_terms_in_mondo_xrefs.tsv $(REPORTDIR)/%_excluded_terms_i
 	--component-signature-path $(REPORTDIR)/component_signature-$*.tsv \
 	--outpath $@
 
-# Exclusions: combined
-$(REPORTDIR)/term_exclusions.txt $(REPORTDIR)/exclusion_reasons.robot.template.tsv: $(foreach n,$(ALL_COMPONENT_IDS), $(REPORTDIR)/$(n)_term_exclusions.txt)
-	cat $(REPORTDIR)/*_term_exclusions.txt > $(REPORTDIR)/term_exclusions.txt; \
+# Exclusions: all artefacts for single ontology
+exclusions-%:
```
joeflack4, line 216: @matentzn We had a goal for running (i) all exclusion artefacts of a certain type for all ontologies, but not a goal for running (ii) all exclusion artefacts for a single ontology, which is something I found useful today.

Actually, we are still lacking (iii) all exclusion artefacts for all ontologies. Maybe something to add.

joeflack4:
@matentzn This afternoon, did: (1) adding of genes to OMIM exclusions, re-ran exclusions pipeline and slurp, (2) re-ran slurp for all ontologies, following our addition/changes to columns.

As you suggested, gonna merge this now. I imagine I'll be making continuous tweaks to the slurp pipeline very soon, but I'll make new (hopefully a lot smaller) PRs for that.

Issue #38, opened 2022-07-18
Basic matching pipeline


matentzn:
I would like to start working on a basic lexical matching pipeline for Mondo ingest. This is somewhat disconnected from all the rest.

The goal is to maintain a continuously updated set of lexical matches in this repo, release these mappings, and feed them to boomer by some kind of criteria. 

It is important the matches have a solid `confidence` value.

matentzn:
I have provided a basic setup for you here: https://github.com/monarch-initiative/mondo-ingest/pull/39, please work on that branch.

You can run the pipeline by running:

```
cd src/ontology
sh run.sh make lexical_matches
```

hrshdhgd:
So right now `oaklib`'s `lexmatch` gives me the following:

**command**: 
```
runoak -i src/ontology/tmp/merged-mondo-ingest-edit.owl lexmatch -L src/ontology/metadata/mondo.sssom.config.yml -o test.sssom.tsv
```
**test.sssom.tsv** :
```
# curie_map: {}
# license: CC-0
# mapping_set_id: default
```
Above is the `curie_map` and `metadata` that needs work but below is the mapping file (partial copy-paste)

| subject_id | subject_label | predicate_id | object_id | object_label | mapping_justification | mapping_tool | subject_match_field | object_match_field | match_string |
|---|---|---|---|---|---|---|---|---|---|
| NCIT:C100101 | Naratuximab Emtansine | skos:closeMatch | NCIT:C170199 | Naratuximab Emtansine | semapv:LexicalMatching | oaklib | rdfs:label | rdfs:label | Naratuximab Emtansine |
| NCIT:C1008 | Azaribine | skos:closeMatch | NCIT:C76914 | Azaribine | semapv:LexicalMatching | oaklib | rdfs:label | rdfs:label | Azaribine |
| NCIT:C101267 | Hormone Receptor Positive | skos:closeMatch | NCIT:C157055 | Hormone Receptor Positive | semapv:LexicalMatching | oaklib | rdfs:label | rdfs:label | Hormone Receptor Positive |
| NCIT:C102533 | Myopia | skos:closeMatch | http://www.omim.org/phenotypicSeries/PS160700 | Myopia | semapv:LexicalMatching | oaklib | rdfs:label | rdfs:label | Myopia |
[...]

PR #39, opened 2022-07-18
Include basic lexical matching pipeline to mondo ingest

matentzn:
- This is a skeleton setup to fix https://github.com/monarch-initiative/mondo-ingest/issues/38
 - All checkmarks in https://github.com/INCATools/ontology-access-kit/issues/191#event-7018102236 should be met.

### DIFF
--- a/src/ontology/config/mondo-match-rules.yaml
+++ b/src/ontology/config/mondo-match-rules.yaml
@@ -1,0 +1,63 @@
+rules:
+  - description: default
+    postconditions:
+      predicate_id: skos:closeMatch
+      weight: 0.0
+
+  - description: exact to exact
+    preconditions:
+      subject_match_field_one_of:
+        - oio:hasExactSynonym
+        - rdfs:label
+        - skos:prefLabel
+      object_match_field_one_of:
+        - oio:hasExactSynonym
+        - rdfs:label
+        - skos:prefLabel
+    postconditions:
+      predicate_id: skos:exactMatch
+      weight: 2.0
+
+  - description: >-
+     label to label; note this is additive with the exact to exact rule,
+      so the score just represents an additional small boost
+    preconditions:
+      subject_match_field_one_of:
+        - rdfs:label
+      object_match_field_one_of:
+        - rdfs:label
+    postconditions:
+      predicate_id: skos:exactMatch
+      weight: 0.5
+
+  - description: xref match
+    preconditions:
+      subject_match_field_one_of:
+        - oio:hasDbXref
+        - skos:exactMatch
+      object_match_field_one_of:
+        - oio:hasDbXref
+        - skos:exactMatch
+    postconditions:
+      predicate_id: skos:exactMatch
+      weight: 4.0
+
+  - preconditions:
+      subject_match_field_one_of:
+        - oio:hasExactSynonym
+        - rdfs:label
+      object_match_field_one_of:
+        - oio:hasBroadSynonym
+    postconditions:
+      predicate_id: skos:broadMatch
+      weight: 2.0
+
+  - preconditions:
+      subject_match_field_one_of:
+        - oio:hasExactSynonym
+        - rdfs:label
+      object_match_field_one_of:
+        - oio:hasNarrowSynonym
+    postconditions:
+      predicate_id: skos:narrowMatch
+      weight: 2.0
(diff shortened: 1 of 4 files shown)


hrshdhgd:
Thoughts @matentzn ? As of now, the `metadata` and `curie_map` are injected outside of OAK (it's done within this project). The `lexmatch` feature does not have any provision to take a yaml input as of now (`metadata/mondo.sssom.config.yml`). Does OAK be held responsible to inject the `metadata` and `curie_map`?

cc:  @cmungall

src/ontology/mondo-ingest.Makefile
```diff
@@ -9,7 +9,7 @@ report-mapping-annotations
 ### Standard constants #############
 ####################################
 MAPPINGSDIR=../mappings
-
+SCRIPTDIR=../scripts
```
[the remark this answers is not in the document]
hrshdhgd: Changed

src/scripts/match-mondo-sources-all-lexical.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+
+if __name__ == '__main__':
+    input = None
+    config_file = None
+    output = None
+    rules = None
+    # TODO: there must be a better way to parse arguments.
+    for arg in sys.argv:
```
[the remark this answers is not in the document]
hrshdhgd: Fixed!

src/scripts/match-mondo-sources-all-lexical.py
```diff
@@ -0,0 +1,93 @@
...
+    with open(config, "r") as f:
+        sssom_yaml = yaml.safe_load(f)
+        
+    resource = OntologyResource(slug=f"sqlite:///{Path(input).absolute()}")
+    oi = SqlImplementation(resource=resource)
+    lexical_index = create_lexical_index(oi)
+    save_lexical_index(lexical_index, OUT_INDEX_DB)
+    msdf = lexical_index_to_sssom(oi, lexical_index)
+    if rules:
+        msdf = lexical_index_to_sssom(oi, lexical_index, ruleset=load_mapping_rules(rules))
```
matentzn, line 80: Matches in the sssom file you have checked in are still all `close`?

src/scripts/match-mondo-sources-all-lexical.py
```diff
@@ -0,0 +1,93 @@
...
+)
+from sssom.writers import write_table
+import sys
+from pathlib import Path
+import yaml
+import click
+
+SRC = Path(__file__).resolve().parents[1]
+ONTOLOGY_DIR = SRC / "ontology"
+OUT_INDEX_DB = ONTOLOGY_DIR / "tmp/merged.db.lexical.yaml"
```
matentzn, line 31: Is this way to construct paths platform independent? I.w. will it work on windows?

matentzn (commented):
love it so far

src/scripts/match-mondo-sources-all-lexical.py
hrshdhgd: Yes, here's a [blog](https://realpython.com/python-pathlib/) that states it.

src/scripts/match-mondo-sources-all-lexical.py
hrshdhgd: I think it has something to do with the rules.yaml. If we don't provide any, the default for all would be `close` as per my [comment here](https://github.com/INCATools/ontology-access-kit/issues/98#issuecomment-1190725945).

hrshdhgd: I confirmed it! It is the rules file that determines the matches in the file.

matentzn: Yes, lets add a rules file permanentlt to `src/ontology/config/mondo-match-rules.yaml` and work on tweaking that moving forwards. Thank you for checking!

src/scripts/match-mondo-sources-all-lexical.py
matentzn: Thank you :)

src/ontology/config/mondo-match-rules.yaml
```diff
@@ -0,0 +1,63 @@
+rules:
```
matentzn, line 1: looks great! Did you update teh table with that yet?

hrshdhgd: Yes.

PR #42, opened 2022-07-25
ICD10CM update in `mondo.sssom.config`

joeflack4:


src/ontology/metadata/mondo.sssom.config.yml
```diff
@@ -20,7 +20,7 @@ curie_map:
   oboInOwl: http://www.geneontology.org/formats/oboInOwl#
   NCBITaxon: http://purl.obolibrary.org/obo/NCBITaxon_
   skos: http://www.w3.org/2004/02/skos/core#
-  ICD10CM: https://icd.codes/icd10cm/
+  ICD10CM: http://purl.bioontology.org/ontology/ICD10CM/
```
joeflack4, line 23: Just this change

matentzn: Can you apply this change here: https://github.com/monarch-initiative/mondo-ingest/blob/main/src/ontology/mondo-ingest.Makefile#L143 as well?

matentzn: (What I mean is, mondo.sssom.config is actually copied from the modno repo as you can see by the Makefile goal)

joeflack4: [Done](https://github.com/monarch-initiative/mondo/pull/5209#discussion_r930364073)!

PR #46, opened 2022-07-28
Patch: Exclusion tables

joeflack4:
Updates:
```
    Feature: Exclusion tables
    - uri_to_curie: Refactored to a combo of my code plus OAK's after all 3 failed : my code by itself, oaklib, bioregistry.
```

### DIFF
--- a/src/scripts/exclusion_term_expansion.py
+++ b/src/scripts/exclusion_term_expansion.py
@@ -22,8 +22,8 @@ from argparse import ArgumentParser
 from typing import Dict, List, Union
 
 import yaml
-from jinja2 import Template
 import pandas as pd
+from jinja2 import Template
 
 
 # Vars
@@ -35,7 +35,6 @@ CONFIG = {
 }
 
 # # Static
-# noinspection DuplicatedCode
 THIS_SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
 SCRIPTS_DIR = THIS_SCRIPT_DIR
 PROJECT_DIR = os.path.join(SCRIPTS_DIR, '..', '..')
@@ -51,19 +50,22 @@ SPARQL_TERM_REGEX_PATH = os.path.join(SPARQL_DIR, 'get-terms-by-regex.sparql.jin
 
 
 # Functions
-# TODO: use: https://bioregistry.readthedocs.io/
-#  en/latest/api/bioregistry.curie_from_iri.html?highlight=get_curie#bioregistry.curie_from_iri
-def uri_to_curie(uri: str) -> str:
-    """Takes an ontological URI and returns a CURIE. Works on the following patterns:
-    - http://.../PREFIX/CODE
-    - http://.../PREFIX_CODE"""
-    uri_list: List[str] = uri.split('/')
-    end_bit = uri_list[-1]
-    if '_' in end_bit:
-        curie = end_bit.replace('_', ':')
-    else:
-        curie = uri_list[-2] + ':' + end_bit
-    curie = curie.replace('<', '').replace('>', '')
+# TODO: evaluate https://github.com/cthoyt/curies , which likely has fully superseded new bioregistry option
+#  https://github.com/biopragmatics/bioregistry/issues/480#issuecomment-1199235747
+def uri_to_curie(uri: str, prefix_map: Dict[str, str]) -> str:
+    """Takes an ontological URI and returns a CURIE. Works on the following patterns
+    todo: eventually OAK might also have an optimal solution for this."""
+    if uri.startswith('<'):
+        uri = uri[1:]
+    if uri.endswith('>'):
+        uri = uri[:-1]
+    curie = None
+    for k, v in prefix_map.items():
+        if uri.startswith(v):
+            curie = uri.replace(v, f"{k}:")
+    if not curie:
+        err = f'Could not locate a suitable URI base in `prefix_map`:\n- uri: {uri}\n- prefix_map: {prefix_map}'
+        raise ValueError(err)
     return curie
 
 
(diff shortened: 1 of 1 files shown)


joeflack4:
This is a necessary patch. In my latest update before we merged, I didn't check all the results. ICD10WHO results, at least, should be incorrect until applying this.

src/scripts/exclusion_term_expansion.py
```diff
@@ -51,19 +52,11 @@ class having `True` for `exclude_children`, but the child class has `False`.
...
-    if '_' in end_bit:
-        curie = end_bit.replace('_', ':')
-    else:
-        curie = uri_list[-2] + ':' + end_bit
-    curie = curie.replace('<', '').replace('>', '')
+def uri_to_curie(uri: str, prefix_map: Dict[str, str]) -> str:
+    """Takes an ontological URI and returns a CURIE. Works on the following patterns
+    # todo: Use this solution until I get OAK to work for this use case
+    # todo: Might be able to speed this up by doing this op on whole column first"""
+    curie = curie_from_iri(uri.replace('<', '').replace('>', ''), prefix_map=prefix_map)
```
joeflack4, line 59: I noticed this problem with angle brackets when I tried to use it the `bioregistry` feature you mentioned, so I opened up [an issue](https://github.com/biopragmatics/bioregistry/issues/479).

I know that eventually we want to use `oaklib`.

joeflack4: *sigh*. Things take so much longer than expected. This `bioregistry` feature is too slow. I created an issue: https://github.com/biopragmatics/bioregistry/issues/480

I looked and the `oaklib` implementation only comes as class methods which require me to do a more elaborate setup and load up an ontology, and I don't have time to do that tonight if I want to produce some results. For now I'll borrow the `oak` code and repurpose it for both my PRs, but I'll need to change this soon.

src/scripts/exclusion_term_expansion.py
```diff
@@ -51,19 +51,29 @@ class having `True` for `exclude_children`, but the child class has `False`.
...
+    curie = uri
+    uri = curie.replace('<', '').replace('>', '')
+    if prefix_map:
+        for k, v in prefix_map.items():
+            if uri.startswith(v):
+                curie = uri.replace(v, f"{k}:")
+        if uri.startswith(obo_purl):
+            uri = uri.replace(obo_purl, "")
+            curie = uri.replace("_", ":")
     else:
```
matentzn, line 69: I would prefer not to have a default for when no prefix_map is provided. just let it fail hard in this case, rows 70-76 are very error prone guesswork and you can imagine dozens of cases where it is simply wrong

src/scripts/exclusion_term_expansion.py
```diff
@@ -51,19 +51,29 @@ class having `True` for `exclude_children`, but the child class has `False`.
...
+    # todo: defaulting prefix_map to None is temp for now but need to change to oak anyway soon
+    # todo: Use this solution until I get OAK to work for this use case"""
+    obo_purl = "http://purl.obolibrary.org/obo/"
+    curie = uri
+    uri = curie.replace('<', '').replace('>', '')
+    if prefix_map:
+        for k, v in prefix_map.items():
+            if uri.startswith(v):
+                curie = uri.replace(v, f"{k}:")
+        if uri.startswith(obo_purl):
```
matentzn, line 66: why is this in the `if prefix_map` section of the code? It should be in the `else` part! If there is a prefix_map, that should always be preferred. Indeed, see comment below, I would prefer if we always use prefix_map, and raise an exception of the URI cannot be checked with the prefix_map. (prefix_map should be required).

matentzn (changes requested):

src/scripts/exclusion_term_expansion.py
matentzn: @joeflack4 no worries, it takes always longer to try and re-use code then to build it yourself for such methods. In the long term it will be worth, it, thank you for your patience! 

The ankle brackets issue: This was actually a mistake in _my_ code: I should have not served you IRIs with `<>`. It is good to know that these ankle brackets denotes "entity reference" is RDF. So, `http://purl.obolibrary.org/obo/MONDO_123` is an IRI, `<http://purl.obolibrary.org/obo/MONDO_123>` is a an entity reference in RDF (for example, a turtle or SPARQL file). I should removed these from the files i provided to you.

Seems the caching issue is also solved - for future iterations of the function please fold it in. :) Thanks!

src/scripts/exclusion_term_expansion.py
joeflack4: That suggestion makes sense. I think there might have been some cases in the past where I didn't have a `prefix_map`, but that's not a situation we want to be in anyway, and I think I've gotten past that point too, so I can remove it.

src/scripts/exclusion_term_expansion.py
joeflack4: I saw your comment [on Slack](https://monarchinitiative.slack.com/archives/C0222P0736F/p1659356271536479) about this as well. This was towards the end of a really looong day of work. I was just trying to get one of the libraries to work and clock out for the night. This code came from OAK, actually: https://github.com/INCATools/ontology-access-kit/blob/main/src/oaklib/interfaces/basic_ontology_interface.py#L136

Although there is a `TODO` in there to change that.

It's NP; I understand your logic completely. I'll remove it.

src/scripts/exclusion_term_expansion.py
joeflack4: done!

src/scripts/exclusion_term_expansion.py
joeflack4: done!

src/scripts/exclusion_term_expansion.py
joeflack4: @matentzn Thanks, I understand. I did change the code to be even more robust though, if you want to leave this in:

Rather than a simple find/replace:
```py
    if uri.startswith('<'):
        uri = uri[1:]
    if uri.endswith('<'):
        uri = uri[:-1]
```

I can't think of any situation in which this would negatively effect things, unless we want to fail early and often, which I guess is maybe Pythonic.

I haven't utilized the `bioregistry` since it was updated a few days ago, but there's a `todo` in there for me to check out Charlie's new library. I don't think we should let that hold up a patch that will fix a bugfix though.

matentzn: Ok to all. Thank you

src/scripts/exclusion_term_expansion.py
```diff
@@ -51,19 +50,22 @@ class having `True` for `exclude_children`, but the child class has `False`.
...
-        curie = uri_list[-2] + ':' + end_bit
-    curie = curie.replace('<', '').replace('>', '')
+# TODO: evaluate https://github.com/cthoyt/curies , which likely has fully superseded new bioregistry option
+#  https://github.com/biopragmatics/bioregistry/issues/480#issuecomment-1199235747
+def uri_to_curie(uri: str, prefix_map: Dict[str, str]) -> str:
+    """Takes an ontological URI and returns a CURIE. Works on the following patterns
+    todo: eventually OAK might also have an optimal solution for this."""
+    if uri.startswith('<'):
+        uri = uri[1:]
+    if uri.endswith('<'):
```
matentzn, line 60:
```suggestion
    if uri.endswith('>'):
```

joeflack4: Haha! Ooops. @_@

joeflack4: Fixed

matentzn (approved):

Issue #51, opened 2022-08-01
Filter matches to only relevant prefixes


matentzn:
The lexical matching pipeline should only contain matches between:

https://github.com/monarch-initiative/mondo-ingest/blob/main/src/ontology/metadata/mondo.sssom.config.yml#L26

- [x] Filter all matches according to the prefixes above
- [x] Ensure that all prefixes in that list occur at least once in the resulting mapping set

hrshdhgd:
Fixed by #52

Issue #50, opened 2022-08-01
Adding synonymizer config to matching pipeline


matentzn:
I think the best way to handle the structural synonym matching is to:

- Create a synonymiser config the specifies all the rules for generating derived, preprocessed versions of labels (or other synonyms) as another synonym time, with an source annotation "MONDO:synonymizer" which points to the synonym generation config. This config should do stuff like:
   - Stripping out redundant parts of the terms (disease, disorder, of, to)
   - Replacing certain other ones ("aberrant" --> "abnormal")
   - General label normalisation - lowercasing, stemming (this could avoid much of the AE/BE issues). (extend synonymiser)
 - Add a synonymiser config to `src/ontology/config`
 - Add a step preceeding `mondo-ingest.db` which
    - computes a synonymiser table
    - outputs it as turtle (extend synonymiser)
    - merges it into mondo-ingest before transforming it to mondo-ingest.db
 
As soon as the general architecture stands (you can start with a rule that normalises "disorder" to "disease").

Note that I will want to use this later for other things, like enriching Mondo itself with synonyms.

nicolevasilevsky:
- when we are doing matching, we want disorder to be replaced by disease
- type 1 -> Type I
- 1 -> type 1
- include 'the' in the label, eg abnormality of the face -> abnormality of face
- of
- plural
- 's, eg Usher's syndrome -> Usher syndrome

matentzn:
Create a second output table like this:

| ID | Synonym | Source | Type |
| --- | --- | --- | --- |
| ID | A oio:hasExactSynonym | >A oio:source |  TYPE |
| MONDO:123 | The Synonym | MONARCH:Synonimizer | owl:Class |
| MONDO:222 | The Syn | MONARCH:Synonimizer | owl:Class |

cmungall:
I would push for a slight variation:

have synonymizer output KGCL. That we decouple it from any implementation

You add then just apply these changes to the .db (or to the .obo if a curator wants them...)

matentzn:
Yes you are right. That would make it possible to bypass robot in this pipeline. 

I still think synonymiser should be a method in oak though, see https://github.com/INCATools/ontology-access-kit/issues/281

Issue #49, opened 2022-08-01
Lexical mapping analysis and feedback/review


hrshdhgd:
- [ ] Create a table 
   - [ ] All matches: For this one, we want general feedback, false assumptions and suggestions (do not spend more than 30 mins)
   - [ ]  Table with just ORDO matches that are not already I MONDO. Goal is to remove everything in `tmp/mondo.sssom.tsv`. 
   - [ ] Compare `tmp/mondo.sssom.tsv` and `src/mappings/mondo-sources-all-lexical.sssom.tsv` 
     - [ ] Implement in sssom-py a method on MSDF `msdf1.remove_mappings(msdf2) -> msdf` output: all rows for `KEY_FEATURES`  that is [msdf1.df - msdf2.df] Implement CLI as well (ideally)
- [ ] Review mappings and get feedback for bad mappings

hrshdhgd:
Do we still need just ORDO mappings?

matentzn:
No all. I think we need to count how many mappings by prefix their are to make sure that all of the ontologies are taking into account by the mapping process

matentzn:
https://github.com/monarch-initiative/mondo/issues/5232

Our first OAK run revealed mostly wrong matches, need to make sure this does not happen again

matentzn:
Issues with the current matching to deal with

- [ ] missed synonyms (bag of words matching could be better than relying on synonyms?): `neuroleptic malignant syndrome` ->	`Malignant neuroleptic syndrome` not caught
- [ ] irrelevant categorical information or synonyms in brackets: `Leprosy [Hansen’s disease]`, `Pregnancy with abortive outcome (O00-O08)`
- [ ] Broken encodings `WaldenstrÃ¶m macroglobulinemia` - any smart idea how to handle with now access to the source? Here we need a smart way. What comes to mind is replacing broken chars with regex wildcards like `Waldenstr.m macroglobulinemia`.

hrshdhgd:
https://github.com/monarch-initiative/mondo-ingest/issues/78#issuecomment-1946865977

PR #52, opened 2022-08-02
Filtered prefixes by `subject_prefixes`

hrshdhgd:
- [x] All lexical matches generated are filtered using  `subject_prefixes` in `mondo.sssom.config.yml` [issue #49 ]
- [ ] issue #50 
- [x] issue #51

### DIFF
--- a/src/scripts/match-mondo-sources-all-lexical.py
+++ b/src/scripts/match-mondo-sources-all-lexical.py
@@ -12,7 +12,7 @@
 
 import logging
 from pathlib import Path
-from typing import List
+from typing import List, TextIO
 from oaklib.resource import OntologyResource
 from oaklib.implementations.sqldb.sql_implementation import SqlImplementation
 from oaklib.utilities.lexical.lexical_indexer import (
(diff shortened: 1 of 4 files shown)


src/scripts/match-mondo-sources-all-lexical.py
```diff
@@ -85,11 +102,105 @@ def run(input:str, config: str, rules:str, output: str):
     
     # msdf.prefix_map = sssom_yaml['curie_map']
     # msdf.metadata = sssom_yaml['global_metadata']
+    msdf.df = filter_prefixes(df=msdf.df, filter_prefixes=prefix_of_interest, features=[SUBJECT_ID, OBJECT_ID])
+    # TODO: replace line below by msdf.remove_mappings(mapping_msdf) once imported from SSSOM.
```
matentzn, line 106: Shall we make an sssom release and do this, or do you think its not worth waiting?

src/scripts/match-mondo-sources-all-lexical.py
```diff
@@ -85,11 +102,105 @@ def run(input:str, config: str, rules:str, output: str):
...
         write_table(msdf, f)
 
+    objects = msdf.df[OBJECT_ID].drop_duplicates()
+    prefixes = objects.str.split(":").str.get(0).drop_duplicates()
+    prefix_args = tuple([x+":%" for _, x in prefixes.iteritems() if x != "MONDO"])
+    kwargs = {"subject_id": ("MONDO:%",), "object_id": prefix_args}
+    with open(str(SRC / Path(output.replace("lexical", "lexical-2"))), "w") as f:
+        filter_file(
+            input=str(SRC / Path(output)),
+            output=f,
```
matentzn, line 133: I have never seen this way of doing it but looks cool. Learned something new!

src/scripts/match-mondo-sources-all-lexical.py
```diff
@@ -85,11 +102,105 @@ def run(input:str, config: str, rules:str, output: str):
...
+            **kwargs
+            )
+
+# TODO: Replace the method below by sssom.utils.filter_prefixes.
```
matentzn, line 137: Maybe if there are these two case (and the remove_mappings one above), lets keep the PR open and make an sssom release to fix this. Just to be clean, we are not in a super rush now.

matentzn (changes requested):
Just changing to using sssom-py instead of the duplicated methods

src/scripts/match-mondo-sources-all-lexical.py
hrshdhgd: I think we should make a release.

src/scripts/match-mondo-sources-all-lexical.py
hrshdhgd: I agree!

PR #69, opened 2022-08-16
Rename preflabel and fix ORDO signature

### LINKED ISSUE #65: `components/icd10cm.owl` missing labels

### LINKED ISSUE #67: Fix problem with ORDO ingest
The ordo "relevant signature" pipeline is broken. It operates on RAW ORDO, while it should operate on the fixed version where part of are replaced by isa.

matentzn:
This pull request fixes #67 by expanding the list of relevant diseases to all three clinical branches (see issue). Now, the signature is around what we expect (ca 9K entities). 

This PR also fixes #65 by making sure that all property renamings happen to all ontologies in the set, so that ontologies, for example, that use prefLabel, end up being renamed to `rdfs:label`.

### DIFF
--- a/src/ontology/config/property-map-1.sssom.tsv
+++ b/src/ontology/config/property-map-1.sssom.tsv
@@ -1,3 +1,4 @@
 subject_id	object_id
 http://purl.org/dc/elements/1.1/source	oboInOwl:source
 http://www.w3.org/2004/02/skos/core#prefLabel	http://www.w3.org/2000/01/rdf-schema#label
+http://www.w3.org/2004/02/skos/core#altLabel	oboInOwl:hasExactSynonym
--- a/src/ontology/config/property-map-1.sssom.tsv
+++ b/src/ontology/config/property-map-1.sssom.tsv
@@ -1,2 +1,3 @@
 subject_id	object_id
 http://purl.org/dc/elements/1.1/source	oboInOwl:source
+http://www.w3.org/2004/02/skos/core#prefLabel	http://www.w3.org/2000/01/rdf-schema#label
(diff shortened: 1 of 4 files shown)


src/ontology/config/property-map-1.sssom.tsv
```diff
@@ -1,2 +1,4 @@
 subject_id	object_id
 http://purl.org/dc/elements/1.1/source	oboInOwl:source
+http://www.w3.org/2004/02/skos/core#prefLabel	http://www.w3.org/2000/01/rdf-schema#label
```
joeflack4, line 3: Used to update some properties of same/similar type that we don't want to use, with properties that we do want to use.

src/ontology/mondo-ingest.Makefile
```diff
@@ -83,15 +89,22 @@ $(COMPONENTSDIR)/doid.owl: $(TMPDIR)/doid_relevant_signature.txt
 		remove -T config/properties.txt --select complement --select properties --trim true \
 		annotate --ontology-iri $(URIBASE)/mondo/sources/doid.owl --version-iri $(URIBASE)/mondo/sources/$(TODAY)/doid.owl -o $@; fi
 
-ICD10CM_URL="https://data.bioontology.org/ontologies/ICD10CM/submissions/21/download?apikey=8b5b7825-538d-40e0-9e9e-5ab9274a9aeb"
+ICD10CM_URL="https://data.bioontology.org/ontologies/ICD10CM/submissions/22/download?apikey=8b5b7825-538d-40e0-9e9e-5ab9274a9aeb"
 
+# This preprocessing is necessary, because the bioportal version of ncit accidentally
```
joeflack4, line 94: This addresses the cause of a recent issue we had where the ICD10CM component `.owl` was empty, and later it was not empty but had no labels.

Issue #45, opened 2022-07-27
Fill out `exclusion_reason`


joeflack4:
## Summary
We finally merged the exclusion tables feature I was working on. [`src/ontology/config/`](https://github.com/monarch-initiative/mondo-ingest/tree/xrefed-terms-outside-relevant-scope/src/ontology/config) now contains 3 files for each ontology:
1. `%_exclusions.tsv`
2. `%_term_exclusions.txt`
3. `%_exclusion_reasons.robot.template.tsv`

The files with the pattern of (1) are the ones you guys will need to update.

Example (`icd10cm_exclusions.tsv`):
|term_id        |term_label                           |exclusion_reason      |exclude_children|
|---------------|-------------------------------------|----------------------|----------------|
|ICD10CM:B95-B97|Bacterial and viral infectious agents|disease causing agents|True            |

## Tasks
For each ontology @matentzn has asked:
- [ ] 1. **Fill blanks**: If there are any entries where `exclusion_reason` is blank, fill it out.
- [ ] 2. **CURIEfy**: The values of the `exclusion_reason` column should all be one of the standard CURIEs for [Mondo exclusion reasons](https://mondo.readthedocs.io/en/latest/editors-guide/exclusion-reasons/), not the labels for those CURIEs.

## Additional info
Edits should be made to the the `%_exclusions.tsv` files themselves. I know that typically the work is done in Google Sheets though. I have the links for a couple of those here: [ICD10CM](  https://docs.google.com/spreadsheets/d/1cZPUReTl34Vu2a03tm2921ehARiIpp4fjeQfS27XzKA/edit#gid=1119821904), [Orphanet/ORDO](https://docs.google.com/spreadsheets/d/16ftBJ8mYqEvSEVNi1tGxIlDfL88vYrs7tSUjSnGqrBw/edit#gid=0)

matentzn:
If I could give an award for best issue and pull request discipline it would be you @joeflack4! Thank you for always taking the time to write up clear, awesomely formatted issues.

nicolevasilevsky:
@joeflack4 we discussed this on the Mondo call. @sabrinatoro and I will work on this initially in this [spreadsheet](https://docs.google.com/spreadsheets/d/16ftBJ8mYqEvSEVNi1tGxIlDfL88vYrs7tSUjSnGqrBw/edit#gid=1096559933), then @matentzn will merge it in when we're done.

We are on it - we will let you know when we're done.

[...]

PR #76, opened 2022-09-08
update exclusion list

sabrinatoro:
addresses #45
I updated the exclusion lists for ICD10CM and ORDO. These are now the source of truth for exclusion terms
@matentzn @joeflack4


joeflack4:
Thanks for the notice! @matentzn Whenever I get news of these updates tables, do you want me to run the exclusion pipeline? Or should I leave that to you? I think you might have mentioned it is not yet so simple for us to add some PR merge CI github workflows for this kinda stuff.

matentzn (commented):
@joeflack4 can you review this? Looks good to me. are you gonna be fine with the capital TRUE?

joeflack4:
I glanced at them and they look fine to me.

I think all-caps is the standard for GoogleSheets and Excel. Pretty sure that `pandas` correctly interprets these as booleans.

matentzn (approved):

PR #97, opened 2022-10-07
Slurp: leading 0 padding on Mondo IDs

joeflack4:
Updates
```
Slurp
- Added leading 0 padding to Mondo IDs
```

### DIFF
--- a/src/scripts/migrate.py
+++ b/src/scripts/migrate.py
@@ -75,8 +75,7 @@ def slurp(
                 break
         if qualified_parents:
             next_mondo_id, mondo_term_ids = _get_next_available_mondo_id(next_mondo_id, max_id, mondo_term_ids)
-            # mondo_id = 'MONDO:' + ('0' * (len(str(next_mondo_id)) - 7)) + str(next_mondo_id)  # min->max needs no pad
-            mondo_id = 'MONDO:' + str(next_mondo_id)
+            mondo_id = 'MONDO:' + str(next_mondo_id).zfill(7)  # leading 0-padding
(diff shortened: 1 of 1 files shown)


src/scripts/migrate.py
```diff
@@ -75,8 +75,7 @@ def slurp(
                 break
         if qualified_parents:
             next_mondo_id, mondo_term_ids = _get_next_available_mondo_id(next_mondo_id, max_id, mondo_term_ids)
-            # mondo_id = 'MONDO:' + ('0' * (len(str(next_mondo_id)) - 7)) + str(next_mondo_id)  # min->max needs no pad
-            mondo_id = 'MONDO:' + str(next_mondo_id)
+            mondo_id = 'MONDO:' + ('0' * (len(str(next_mondo_id)) - 7)) + str(next_mondo_id)  # min->max needs no pad
```
joeflack4, line 78: The padding

matentzn: I  think it is more robust to use zfill:

https://stackoverflow.com/questions/39402795/how-to-pad-a-string-with-leading-zeros-in-python-3

matentzn (changes requested):

src/scripts/migrate.py
joeflack4: Ok; didn't even know that existed. This is such a simple programming problem that I didn't even think twice. I'll switch to use that, though.

joeflack4: done

matentzn (approved):

PR #100, opened 2022-10-08
Bugfixes: (i) stale slurp prereqs, (ii) sorting of output tables

joeflack4:
Updates
```
Slurp
- Bugfix: Added -B to call within `slurp-%` make goal, which forces prerequisites to be updated. This solves the issue where prereqs were not being updated. One of the issues that manifested which this fixes is that `mondo.sssom.tsv` was not updated, leading to slurp candidates that had already been mapped.
- Add: New slurp make goals to run the pipeline w/out updating prereqs. Useful if we know that the prereqs are up to date and want to run it faster.
- Update: Slurp outputs (and some intermediary files, such as signature files and exclusion tables) for all ontologies.
- Update: Outputs now sorted (all columns) for consistency

Exclusion tables
- Update: Outputs now sorted (all columns) for consistency
```


src/ontology/mondo-ingest.Makefile
```diff
@@ -366,6 +366,6 @@ slurp/%.tsv: $(COMPONENTSDIR)/%.owl $(TMPDIR)/mondo.sssom.tsv $(REPORTDIR)/%_ter
 	--outpath $@
 
 slurp-%:
-	$(MAKE) slurp/$*.tsv
+	$(MAKE) slurp/$*.tsv -B
```
joeflack4, line 369: @matentzn Adding `-B` should fix it. You went over this w/ me very recently.

Honestly, this should have been pretty obvious to us. Additionally, this is likely an issue for a good many of our `make` goals; that they are not pre-designed to / nor have we (or at least, not I) made it a practice to force update prereqs every time we run them.

I think we will want to put some thought into how we do this. Maybe it's as simple as just remembering to pass `-B` whenever we want to update prereqs. Or maybe there's some other means of foolproofing, or designing variations of goals which either force update prereqs or run normally.

Note that this fix, so to speak, if we want to keep it this way, only applies when running `make slurp-%`. Running `make slurp/%.tsv` will not force update prereqs.

joeflack4:
@matentzn I also ran `slurp-ordo` and got the updated outputs. I checked and verified that indeed, after updating prereqs, the problem appears to be fixed; the problematic example terms which should not have been in the slurp output (Orphanet:101334, Orphanet:454831) are no longer there.

**Important: other ontologies have not yet been re-run**
Just want to share with my preference that I would prefer not to have to run the whole pipeline manually and push whenever we make an update. I fixed this and re-ran for ORDO, but this still needs to be re-run for the other ontologies. Sometimes, it's not too bad; I can run this in the background while I do other stuff. But often, it can be a big inconvenience, and doesn't feel like a great use of my time. Obviously I know we want to automate this, probably w/ GitHub actions. Until we do that, I'm just wondering who the best person to do this is. If you feel like I'm still the best person to be re-running pipelines when changes like these happen until we do more automation, I trust your judgement.

src/ontology/mondo-ingest.Makefile
```diff
@@ -365,7 +365,12 @@ slurp/%.tsv: $(COMPONENTSDIR)/%.owl $(TMPDIR)/mondo.sssom.tsv $(REPORTDIR)/%_ter
 	--slurp-dir-path slurp/ \
 	--outpath $@
 
-slurp-%:
+slurp-no-updates-%:
```
joeflack4, line 368: New slurp make goals to run the pipeline w/out updating prereqs. Useful if we know that the prereqs are up to date and want to run it faster.
```make
slurp-no-updates-%:
	$(MAKE) slurp/$*.tsv

slurp-all-no-updates: slurp-no-updates-omim slurp-no-updates-doid slurp-no-updates-ordo slurp-no-updates-icd10cm slurp-no-updates-icd10who slurp-no-updates-ncit
```

matentzn: Good idea.

matentzn (approved):
That was it? Ok lucky us :) THANKS for getting to the bottom of this so quickly!

joeflack4:
@matentzn Just re-ran the pipeline after making changes to (i) updating slurp prereqs, (ii) 0-padding, (iii) sorting. Files have been committed w/ this PR.

src/scripts/exclusion_term_expansion.py
```diff
@@ -292,6 +292,10 @@ def run(
         df_results = pd.DataFrame()
         df_results['term_id'] = ''
 
+    # Sort
+    if len(df_results) > 0:
+        df_results = df_results.sort_values(['term_id', 'exclusion_reason'])
```
joeflack4, line 297: Sorting: exclusion tables

src/scripts/migrate.py
```diff
@@ -82,8 +82,12 @@ def slurp(
                 'original_label': t.label if t.label else '', 'definition': t.definition if t.definition else '',
                 'parents': '|'.join(parent_mondo_ids)})
 
-    result = pd.DataFrame([ROBOT_TEMPLATE_HEADER] + terms_to_slurp)
-    result.to_csv(outpath, sep="\t", index=False)
+    # Sort, add robot row, save and return
+    result = pd.DataFrame(terms_to_slurp)
+    result = result.sort_values(
```
joeflack4, line 87: Sorting: slurp

matentzn (approved):

joeflack4:
Merging and closing per approval.

There are some things we should hash out in regards to (i) automating run of pipeline, (ii) how to formalize our practice of running goals with regards to updating `make` prerequisites; when / when not to do that, and how to standardize that in the `makefile`.

PR #103, opened 2022-10-14
Updated ICD10CM mappings etc

joeflack4:
## Updates
```
- Updated all mappings via `make mappings`
```

## Related
https://github.com/monarch-initiative/mondo-ingest/issues/78


hrshdhgd (approved):

joeflack4:
I got a lot of messages like this when running `make mappings`, so I'm including the full log just in case we need it:
> `WARNING:root:ICD-10:<CODE> does not follow any known prefixes`
[make mappings.txt](https://github.com/monarch-initiative/mondo-ingest/files/9789868/make.mappings.txt)

joeflack4:
@hrshdhgd I think you said you needed an updated `mondo.sssom.tsv`? 

I'm actually not sure how to get that. In `mondo-ingest`, here's the command:
```make
tmp/mondo.sssom.tsv:
	wget http://purl.obolibrary.org/obo/mondo/mappings/mondo.sssom.tsv -O $@
```

I looked at the `mondo` repo and I don't see how to create it either.

nicolevasilevsky:
@joeflack4 thanks for doing this, but I don't think we need this at the moment. At least not to address the questions that came up about the ICD10 mappings on the tech call today. :)

Issue #85, opened 2022-10-01
Exclusion tables: all rows must have an exclusion reason


matentzn:
Please make sure that all exclusion table rows have exclusion reasons documented.

Once this is done, we should add a check in your code @joeflack4 that makes the extensional list generation fail if an exclusion reason is not provided.

joeflack4:
OK, that is an easy thing for me to add. I'll await when this is ready for me to add that.

joeflack4:
Yep, those are the ones that are missing exclusion reasons from what I can see in `mondo-ingest` `src/ontology/config/`.

sabrinatoro:
OK, I'll get on it!

PR #105, opened 2022-11-02
update omim-exclusion.tsv

sabrinatoro:
Addresses #85

@joeflack4 @matentzn This PR only updates the OMIM-exclusion.tsv. I am still waiting on discussion wrt "allergy" terms which is required in order to update the do-exclusion.tsv. Therefore I will add this file in another PR.


matentzn:
The diff is a bit enormous, 17K lines removed?

sabrinatoro:
I have tried to re-submit, but the same thing happens: for whatever reason, ALL the lines with genes are shown in the difference as deleted and re-added. 
I don't know why it does this.

sabrinatoro:
was replaced by https://github.com/monarch-initiative/mondo-ingest/pull/107

PR #106, opened 2022-11-02
updated ncit sparql and exclusion list

sabrinatoro:
Addresses https://github.com/monarch-initiative/mondo/issues/5252

@joeflack4 and @matentzn
I created a new exclusion code : **MONDO:excludeSpeciesoutofscope** 
I added this new exclusion code to our source of truth [here](https://docs.google.com/spreadsheets/d/18V-_j4b22LzCytxFEM3pBEl6-_ODJY06gpUwH0F6JVY/edit#gid=1644570180) and [here](https://mondo.readthedocs.io/en/latest/editors-guide/exclusion-reasons/). It probably needs to be added somewhere in the code (???). I am leaving this part to you.

Please note that I am aware of a term that was "forgotten", and therefore I would expect this term to get imported in Mondo (this is a good way to test that the import works as expected). Please let me know when it is time to review that the import work. Thank you!


matentzn (approved):
Looks great!

PR #107, opened 2022-11-03
update omim-exclusion.tsv

sabrinatoro:
Addresses #85


sabrinatoro:
I created a new PR just to make sure I did the right thing. However the same very big difference is shown.
Even though only the first 212 lines were changed, the diff shows that ALL the lines with genes (which is the rest of the file) were deleted and re-added.
I don't know why it does this.

sabrinatoro:
@matentzn and @joeflack4 please see THIS PR instead of the other.

sabrinatoro:
I am adding "false" to all the lines. Please wait before reviewing
(and sorry for all the extra steps!)

sabrinatoro:
@matentzn and @joeflack4 : this is ready for your review. Thank you!

matentzn (approved):
Looks great!

PR #108, opened 2022-11-12
Mapped deprecated terms

joeflack4:
Previous PR name: Slurp: (i) re-run of outputs, (ii) deprecated analysis (needs generalization from OMIM into slurp)

## Updates
    Slurp
    - Update: slurp/omim.tsv and all artefact dependencies, following latest release addressing OMIM issues #98 (invalid refs OMIM:1 -> OMIM:14) and #84 (OMIM terms with 'None' as label).

    Deprecated terms in Mondo
    - Add: Script and make goals for this. Decoupled this from slurp/
    - Add: reports/omim_mapped_deprecated_terms.robot.template.tsv: A list of obsolete/deprecated OMIM terms that remain within Mondo.
    - Add: Docs

    Mapping progress report
    - Updates: Docs: Removed some unnecessary things / simplified some things

    Misc
    - Updates: Scripting: --sssom-map-path --> --mondo-mappings-path

## Reviewers
- Harshad & Vinícius: general comments
- Nicole and Sabrina to inspect outputs.
- Nico to inspect [`reports/omim_mapped_deprecated_terms.robot.template.tsv`](https://github.com/monarch-initiative/mondo-ingest/pull/108/commits/66e7654801be2767d0a9be2dd54ff527ee9f20eb#diff-b0d295ecd7d05fde88526ba07223a9bfbc08cc572d5ead04804bd3d49d2741b5) and related scripting.

PR #118, opened 2022-12-01
updated omim exclusion list

sabrinatoro:
related to https://github.com/monarch-initiative/mondo-ingest/pull/108

### DIFF
--- a/src/ontology/config/omim_exclusions.tsv
+++ b/src/ontology/config/omim_exclusions.tsv
@@ -210,6 +210,243 @@ OMIM:618079	low density lipoprotein cholesterol level quantitative trait locus 8
 OMIM:618377	pain sensitivity quantitative trait locus 1	MONDO:excludeGene	FALSE
 OMIM:618406	body mass index quantitative trait locus 20	MONDO:excludeGene	FALSE
 OMIM:618807	lipoprotein(a) quantitative trait locus	MONDO:excludeGene	FALSE
+OMIM:146810	immune response to synthetic polypeptide--irphegal	MONDO:excludeBiologicalProcess	FALSE
+OMIM:146820	immune response to synthetic polypeptide--irgat	MONDO:excludeBiologicalProcess	FALSE
+OMIM:146950	immune response to synthetic polypeptide--irhgal	MONDO:excludeBiologicalProcess	FALSE
+OMIM:146960	immune response to synthetic polypeptide--irtgal	MONDO:excludeBiologicalProcess	FALSE
+OMIM:147080	immune response to synthetic polypeptide--irglphe 1	MONDO:excludeBiologicalProcess	FALSE
+OMIM:147090	immune response to synthetic polypeptide--irglphe 2	MONDO:excludeBiologicalProcess	FALSE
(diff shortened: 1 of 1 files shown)


sabrinatoro:
@joeflack4 when this is merged, can you please create a new omim.ttl for me to review? Thank you!

joeflack4:
@sabrinatoro sure thing. I might not see exactly when this merged. Actually, I just checked, and it looks like we did create a [GH action](https://github.com/monarch-initiative/omim/actions/workflows/main.yml) for the OMIM ingest that will build and [create a new release](https://github.com/monarch-initiative/omim/releases/). It's one of those where you can click "run workflow". Probably takes 5 minutes or so. It looks like the last few runs were successful. Can you give that a go and let me know if it works?

src/ontology/config/omim_exclusions.tsv
```diff
@@ -210,6 +210,243 @@ OMIM:618079	low density lipoprotein cholesterol level quantitative trait locus 8
...
+OMIM:603991	zinc finger protein 109	MONDO:excludeGene	FALSE
+OMIM:603992	zinc finger protein 110	MONDO:excludeGene	FALSE
+OMIM:603993	zinc finger protein 111	MONDO:excludeGene	FALSE
+OMIM:603995	zinc finger protein 113	MONDO:excludeGene	FALSE
+OMIM:603997	zinc finger protein 118	MONDO:excludeGene	FALSE
+OMIM:603998	zinc finger protein 119	MONDO:excludeGene	FALSE
+OMIM:604000	zinc finger protein 122	MONDO:excludeGene	FALSE
+OMIM:604301	oculomedin	MONDO:excludeGene	FALSE
+OMIM:604754	zinc finger protein 271, pseudogene	MONDO:excludeGene	FALSE
+OMIM:604966	protocadherin-alpha gene cluster	MONDO:excludeGene	FALSE
```
matentzn, line 291: Remember that you do not _have_ to exclude these 1 by 1. @joeflack4 can you give @sabrinatoro and example on how to exclude all terms that end with `gene cluster`, `protein`, `pseudogene` and start with `blood group`?

sabrinatoro: @matentzn I know... :-) but since I was going through the list, it was easy to just add it "in bulk". 
I will work on "exclusion rules" in the upcoming few weeks (hopefully before the end of the year)

sabrinatoro:
@matentzn @joeflack4 
Can we merge this? It looks like the comments are more "to do later"

src/ontology/config/omim_exclusions.tsv
joeflack4: Sure. @sabrinatoro I just realized I need to add this to the list of documentation I need to do.

The exclusion tables support regular expressions. I often like to use sites like https://www.regexpal.com/ to check my syntax. But in this case it's a simple regex, e.g.: `.*gene cluster` should match any line of text that ends with 'gene cluster'.

In order to use a regex with the exclusion tables, it needs `REGEX:` before the regex. So in the `term_label` column, you will want it to say `REGEX:.*gene cluster`, and then it will find all terms with a label that matches that regex.

joeflack4:
I defer to Nico but I'm fine w/ merging~

PR #124, opened 2022-12-03
Slurp: (i) bugfix: obsolete mondo terms, (ii) ORDO updates, (iii) bugfix: exclusions

joeflack4:
Updates
```
Slurp
- Bugfix: scripting: Now checks if parents' mapped mondo terms are obsolete and, if so, excludes from slurp.
- Update: outputs: slurp/ordo.tsv

Exclusions
- Bugfix: Now raises `robot` error if one received.
- Bugfix: Fixed issue where Mondo terms were being queried as well rather than just the terms in the given ontology.
- Bugfix: config/ordo_exclusions.tsv: Had some MONDO terms that are now removed.
```


src/scripts/exclusion_term_expansion.py
```diff
@@ -120,7 +120,12 @@ def sparql_jinja2_robot_query(
             f.write(instantiated_str)
         with open(command_save_path, 'w') as f:
             f.write(command_str)
-        subprocess.run(command_str.split())
+        result = subprocess.run(command_str.split(), capture_output=True, text=True)
```
joeflack4, line 123: @matentzn Now capturing output of `robot` and raising error where it occurs. I should've done this before.

src/ontology/config/ordo_exclusions.tsv
```diff
@@ -58,58 +58,3 @@ Orphanet:238621	Ileal pouch anal anastomosis related faecal incontinence	TRUE	MO
 Orphanet:505395	Ventilator-induced diaphragmatic dysfunction	TRUE	MONDO:excludePhenotype
 Orphanet:90076	Partial deep dermal and full thickness burns	TRUE	MONDO:excludePhenotype
 Orphanet:90080	Scarring in glaucoma filtration surgical procedures	TRUE	MONDO:excludePhenotype
-MONDO:0850029	antibiotic therapy dose selection	TRUE	MONDO:excludeNonDisease
```
joeflack4, line 61: @nicolevasilevsky These were added a couple of days ago, but this causes an error in the exclusion expansion script (and slurp, which also uses it as a step), because it only expects terms from the given ontology.

Did you mean to excluded any Orphanet terms that are mapped to these Mondo terms, or some other functionality?  If there is a use case here, we can talk with Nico about how to do, but my guess is that we will want to keep only the given ontology's terms here.

For now, I've removed these rows within this PR.

src/ontology/slurp/ordo.tsv
```diff
@@ -1,118 +1,57 @@
 mondo_id	mondo_label	xref	xref_source	original_label	definition	parents
```
joeflack4, line 1: @nicolevasilevsky This should be alright now after the bugfix for obsolete parents, but can you check.

src/scripts/migrate.py
```diff
@@ -60,27 +60,26 @@ def slurp(
...
+        qualified_parent_mondo_ids = []
         for parent_curie in t.direct_owned_parent_curies:
             # Conversely, if any of T.parents is unmapped, T is not to be slurped
             if parent_curie in sssom_object_ids:
-                parent_df: pd.DataFrame = sssom_df[sssom_df['object_id'] == parent_curie]
-                parent_mondo_id = parent_df.to_dict('records')[0]['subject_id']
-                parent_mondo_ids.append(parent_mondo_id)
+                parent: Dict = sssom_df[sssom_df['object_id'] == parent_curie].to_dict('records')[0]
+                parent_mondo_id, parent_mondo_label = parent['subject_id'], parent['subject_label']
+                if not parent_mondo_label.startswith('obsolete'):
```
joeflack4, line 71: @nicolevasilevsky @matentzn I don't know if this is the best way to check for obsoletes in this case, but checking `arent_mondo_label.startswith('obsolete')` was a fast/easy way to do it with the current slurp setup.

src/scripts/migrate.py
```diff
@@ -60,27 +60,26 @@ def slurp(
...
+                    qualified_parent_mondo_ids.append(parent_mondo_id)
             if parent_curie not in sssom_object_ids and parent_curie in owned_term_curies:
-                qualified_parents = False
                 break
-        if qualified_parents:
```
joeflack4, line 76: @matentzn This variable not needed. just can check for the list of qualified parents.

src/ontology/config/ordo_exclusions.tsv
matentzn: How did THAT happen?

matentzn: No this is just wrong.

src/ontology/slurp/ordo.tsv
matentzn: This looks really good to my eyes, well done!

src/scripts/exclusion_term_expansion.py
matentzn: excellent

src/scripts/migrate.py
matentzn: not reliable, as, I think, not all Mondo classes have a label. You can perhaps proxy:

- if starts with "obsolete" OR there is no label

joeflack4: Ok, sounds good. Will do. I hope this will be sufficient into the future!

src/ontology/config/ordo_exclusions.tsv
joeflack4: this is resolved on my end

joeflack4:
@matentzn I forgot to bring this one up at the meeting too. You reviewed this previously, and it didn't look like you had anything that I still needed to resolve. I re-looked at all the threads and marked them resolved. This should be ready to merge.

matentzn (approved):

PR #127, opened 2022-12-22
OMIM exclusions & exclusion expansion scripting

joeflack4:
## Updates
OMIM exclusions & exclusion expansion scripting
- Added the SO 'gene' term to list of OMIM exclusions, including its children.
- Updates scripting to include expansion for ontologies not native/owned by the ontology in question (e.g. SO is not 'native' to OMIM)
- - Update: Remove duplicates from reports/ONTO_exclusions.txt

## Additional info
This was requested here: https://github.com/monarch-initiative/mondo-ingest/pull/108#issuecomment-1319554021


src/ontology/config/omim_exclusions.tsv
```diff
@@ -1,4 +1,5 @@
 term_id	term_label	exclusion_reason	exclude_children
+SO:0000704	gene	MONDO:excludeGene	TRUE
```
joeflack4, line 2: Added `SO:0000704` (`gene`)

src/ontology/metadata/omim.yml
```diff
@@ -3,6 +3,24 @@ label: Online Mendelian Inheritance in Man
...
+   CL: http://purl.obolibrary.org/obo/CL_
+   HGNC: "https://identifiers.org/hgnc:"
+   HGNC_symbol: https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/
+   IAO: http://purl.obolibrary.org/obo/IAO_
+   NCBIGene: https://www.ncbi.nlm.nih.gov/gene/
+   NCBITaxon: http://purl.obolibrary.org/obo/NCBITaxon_
+   ORPHA: http://www.orpha.net/ORDO/Orphanet_
+   PMID: http://www.ncbi.nlm.nih.gov/pubmed/
+   RO: http://purl.obolibrary.org/obo/RO_
+   SO: http://purl.obolibrary.org/obo/SO_
```
joeflack4, line 16: @matentzn This is the relevant part for your review.

In order to expand SO, we need more than just `base_prefix_map` (the list of "owned" "native" ontology prefixes).

I considered two ways of doing this:
a. Just provide all of the prefixes that are inside the ontology itself, which is what I did here.
b. Try and just expand it without providing a prefix_map for it. https://github.com/biopragmatics/bioregistry#generating-iris

Maybe there's a different way you prefer. I went with (a) because it's safer. But I should note that now I've changed the exclusion term expansion script to use `prefix_map` instead of `base_prefix_map`. I just had to remove all references to `base_prefix_map` and change them to `prefix_map`. 

Questions
1. If you agree with (a), then we are no longer using `base_prefix_map`, so should we remove it?
2. Should I update `prefix_map` in all of the `metadata/ONTO.yml` to the full prefix_map in each of the ontologies?
3. If (2) is yes, then should we automate that? I wonder if OAK or some other tool has a way to extract the `prefix_map` from a given ontology file already.

src/ontology/reports/omim_exclusion_reasons.robot.template.tsv
```diff
@@ -1,9 +1,11 @@
 term_id	exclusion_reason
```
joeflack4, line 1: Ugh, so this is more work for me but I feel like I should prioritize other stuff before this. However, that's up to @matentzn 

Here's the problem I noticed. The following hypothesis proved to be false.

Hypothesis: If I add `SO:<gene>` to `config/omim_exclusions.tsv`, and and run the exclusion term expansion, I get _outputs A_. Then if I remove all of the other OMIM terms in that file which had the exclusion reason `MONDO:excludeGene` and rerun, I get _outputs B_. Outputs _A_ and _B_ should be exactly the same.

But they were not the same. I could do an analysis of the differences between the lists and look further, but I will wait on that.
[before and after.zip](https://github.com/monarch-initiative/mondo-ingest/files/10290792/before.and.after.zip)

joeflack4: Never mind. I don't know what happened in this first run, but I re-ran, and I see that there is a small difference, but it is an expected one. there was a gene term that was previously missing from config/omim_exclusions.tsv, and it makes sense that it is showing in the output since it is a descendant of SO:<gene>

src/ontology/metadata/omim.yml
joeflack4: I had some previous thoughts in the above comment that I found incorrect and removed.

src/scripts/exclusion_term_expansion.py
```diff
@@ -297,13 +297,19 @@ def run(
         df_results = pd.DataFrame()
         df_results['term_id'] = ''
 
+    # Filter out non-owned
```
joeflack4, line 300: @matentzn I just now filtered results so any non-owned terms don't show up in the output. `NCBITaxon:9606`(not sure how this one got there because it's not in `config/omim_exclusions.tsv`) and `SO:0000704` (gene) were in the output. I figure that these outputs shouldn't include terms foreign to the ontology, but if that's incorrect, let me know.

src/scripts/exclusion_term_expansion.py
```diff
@@ -297,13 +297,19 @@ def run(
...
+    df_results.drop('prefix', inplace=True, axis=1)
+
     # Sort
     if len(df_results) > 0:
         df_results = df_results.sort_values(['term_id', 'exclusion_reason'])
 
     # Add simplified version
     # - df_results_simple: Simpler dataframe which is easier to use downstream for other purposes
-    df_results_simple = df_results[['term_id']]
+    df_results_simple = pd.DataFrame(set(df_results['term_id']))
```
joeflack4, line 312: FYI: there were some duplicate terms in the output, because some of the terms got excluded for multiple reasons. This will still show up in the robot TSV output, because the exclusion reason is one of the columns. But for the TXT output, it's just a flat list of terms, so I removed any duplicates.

joeflack4:
Ready to review and merge.

src/ontology/metadata/omim.yml
matentzn: a) was the correct way. The `prefix_map` should be used for expansions, the `base_prefix_map` is more a lookup which IRI prefixes belong to which ontology.. So you did it correctly.

> Should I update prefix_map in all of the metadata/ONTO.yml to the full prefix_map in each of the ontologies?

Not enough context for this - I would prefer a single prefix map for everything and base prefixes for each individual ontologies. Feel free to put on list and ask in meeting with some more context.

src/scripts/exclusion_term_expansion.py
matentzn: I am not sure if this is _always_ correct, but for all ontologies we are dealing with so far, you are right.

src/scripts/exclusion_term_expansion.py
matentzn: Interesting, and weird. Please make an issue about it with some examples and ask Sabrina to take a look.

matentzn (approved):
Looks good to me, thanks!

src/ontology/metadata/omim.yml
joeflack4: That sounds like a good idea. Although, what if two different ontologies (let's just say ORDO and DO) are using 2 different versions / URIs for their prefixes for a given external ontology (let's just say RO for sake of example).

A way to fix this would be, for `prefix_map` to not be of type `Dict[str, str]`, but of type `Dict[str, List[str]` to accommodate for various different versions / URIs for each of these prefixes.

I'll bring up at the meeting.

joeflack4:
Looks like this is approved; merging.

src/scripts/exclusion_term_expansion.py
joeflack4: Done! https://github.com/monarch-initiative/mondo-ingest/issues/128

PR #135, opened 2023-01-21
New SPARQL queries to remove dbXrefs

### LINKED ISSUE #134: Only ingest hasDbXref of controlled set of prefixes
All sources receive a series of preprocessing steps, see for example here:

https://github.com/monarch-initiative/mondo-ingest/blob/main/src/ontology/mondo-ingest.Makefile#L63

We need another such operation that allows us to get rid of Prefixes like IMDRF we don't care about.

Note however my question in OBO slack, ontotools channel:

image

For now:

Try to 

1. Write a sparql query that deletes xref annotations if they do not start with a certain prefix, see https://github.com/monarch-initiative/mondo/blob/master/src/sparql/update/delete-axiom-annotations-by-prefix.ru, or much better https://github.com/monarch-initiative/mondo/blob/master/src/sparql/update/rm-related-where-label.ru
2. add it to the make pipeline of all 5 currently processed ingests

hrshdhgd:
This removes the following dbXrefs:

- From NCIT: 
  - `IMDRF`
- From DOID
  - `UMLS_ICD9CM_2005_AUI`
  - `ICD11`
  - `SNOMEDCT_US_*`

Fixes #134


src/ontology/metadata/doid.metadata.sssom.yml
```diff
@@ -28,6 +29,8 @@ curie_map:
   SNOMEDCT_US_2020_09_01: http://identifiers.org/snomedct/
   SNOMEDCT_US_2021_07_31: http://identifiers.org/snomedct/
   SNOMEDCT_US_2021_09_01: http://identifiers.org/snomedct/
+  SNOMEDCT_US_2022_07_31: http://identifiers.org/snomedct/
+  SNOMEDCT_US_2022_10_31: http://identifiers.org/snomedct/
```
hrshdhgd, line 33: I added these but also added a SPARQL query to remove these. @matentzn , do we need to make a decision on this now? It seems like every release will have a new prefix (since there is a date).

hrshdhgd:
For now I have the SPARQL queries for NCIT and DOID separate. Both are essentially the same with just the difference of prefixes. We could potentially just add all prefixes in one query and reuse it wherever needed. Thoughts?

hrshdhgd:
`sh run.sh make mappings -B`  runs out pf heap space.

src/ontology/metadata/doid.metadata.sssom.yml
matentzn: The snomed ones I just want to require so they don't fall under the remove situation; but it is a bit annoying this :)

src/sparql/rm_dbxref_doid.ru
```diff
@@ -0,0 +1,20 @@
+PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
```
matentzn, line 2: There is afaics no need to split into two queries - just make 1 and run it for both ontologies. (rm_xref_by_prefix.ru) if a prefix should be deleted in one ontology it should be in all.

src/sparql/rm_dbxref_doid.ru
```diff
@@ -0,0 +1,20 @@
...
+  ?x oboInOwl:hasDbXref ?xref .
+  OPTIONAL { 
+    ?x a owl:Axiom ;
+    owl:annotatedSource ?cls ;
+    owl:annotatedProperty ?pred ;
+    owl:annotatedTarget ?obj ;
+    oboInOwl:hasDbXref ?xref ;
+    ?p1 ?o2 .
+  }
+  FILTER( STRSTARTS(str(?xref), "UMLS_ICD9CM_2005_AUI:") || STRSTARTS(str(?xref), "ICD11:") || STRSTARTS(str(?xref), "SNOMEDCT_US_"))
```
matentzn, line 19: Keep ICD11 and SNOMED. There is a fix_xref sparql query where these should be rewired to the form we want. ICD11 is new and should simply be added to the prefix map for now!

matentzn (changes requested):
Excellent work! So glad it works now :)

matentzn (approved):

PR #185, opened 2023-02-11
Bugfix: NCIT mirror signature

joeflack4:
Updates
```
Bugfix: NCIT mirror signature
- Add: Script to get mirror signature (owned terms) via OAK.
- Update: Make goal for mirror signature to use script if NCIT.
```

### DIFF
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -328,8 +328,13 @@ mondo-ordo-subclass: $(REPORTDIR)/mondo_ordo_unsupported_subclass.tsv
 reports/mirror_signature-mondo.tsv: tmp/mondo.owl
 	$(ROBOT) query -i $< --query ../sparql/classes.sparql $@
 
-reports/mirror_signature-%.tsv: component-download-%.owl
-	$(ROBOT) query -i $(TMPDIR)/$<.owl --query ../sparql/classes.sparql $@
+# todo: it'd be ideal if only component-download-%.owl was only needed for the 'else' clause
+reports/mirror_signature-%.tsv: component-download-%.owl $(COMPONENTSDIR)/%.db metadata/%.yml
+	@if [ $* = "ncit" ]; then\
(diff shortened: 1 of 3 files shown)


src/ontology/mondo-ingest.Makefile
```diff
@@ -328,8 +328,13 @@ mondo-ordo-subclass: $(REPORTDIR)/mondo_ordo_unsupported_subclass.tsv
 reports/mirror_signature-mondo.tsv: tmp/mondo.owl
 	$(ROBOT) query -i $< --query ../sparql/classes.sparql $@
 
-reports/mirror_signature-%.tsv: component-download-%.owl
-	$(ROBOT) query -i $(TMPDIR)/$<.owl --query ../sparql/classes.sparql $@
+# todo: it'd be ideal if only component-download-%.owl was only needed for the 'else' clause
+reports/mirror_signature-%.tsv: component-download-%.owl $(COMPONENTSDIR)/%.db metadata/%.yml
```
joeflack4, line 332: FYI: I changed `reports/mirror_signature-%.tsv` into an if/else. If NCIT, it runs the python script which solves the `robot` memory issue. Otherwise, it generates it the normal way using SPARQL/`robot`.

src/scripts/mirror_signature_via_oak.py
```diff
@@ -0,0 +1,67 @@
+"""An alternative way of creating a `reports/mirror_signature-ONTOLOGY.tsv` which uses OAK instead of `robot`.
```
joeflack4, line 1: FYI: The script which creates a `mirror_signature-%.tsv` via OAK instead of SPARQL/`robot`.

joeflack4, line 1: FYI: The script which creates a `mirror_signature-%.tsv` via OAK instead of SPARQL/`robot`.

joeflack4 (commented):
@hrshdhgd @souzadevinicius This PR fixes an issue where our ODK container was running out of memory (even with 20GB) when trying to simply generate a list of NCIT classes using SPARQL/`robot`.

This is a small PR. Feel free to review and let me know if you have any comments~

Issue #183, opened 2023-02-10
Remove all "inactive" classes from ORDO from ORDO ingest?


matentzn:
Should all of these https://www.ebi.ac.uk/ols/ontologies/ordo/terms?iri=http%3A%2F%2Fwww.orpha.net%2FORDO%2FOrphanet_C041&lang=en&viewMode=All&siblings=false

be excluded from ORDO, i.e. added to the exclusion table?

sabrinatoro:
Thinking about what purpose would ingest these inactive classes would be. 
![Screenshot 2023-02-13 at 9 27 23 AM](https://user-images.githubusercontent.com/12737987/218529304-a976e460-21bf-4f11-9f7a-6550edb8bf3e.png)

- the deprecate and obsolete classes would not be available in orphanet anymore. Therefore, I think it is safe to add these to the list of excluded terms. 
Note: the ID that are already in Mondo would have to be updated to "Mondo:EquivalentObsolete"

- the "non rare" branch is trickier: we want all the diseases in Mondo (rare and non-rare)
BUT, the orphanet ids (and any other information about the terms) are not available on the orphanet website, therefore I don't think it is useful to have an x-ref to an orphanet ID that is not available in orphanet anymore. 
Note: same comment about having to update the one existing in Mondo to "Mondo: EquivalentObsolete".

Therefore I think we can safely exclude all the classes under "inactive" in ordo.

@nicolevasilevsky, do you want to weight in before I add these to the exclusion list?

matentzn:
Remeber you only need to add the parent to the exclusion list, and say that children should be excluded too.

matentzn:
Also, you need to advice @joeflack4 on whether all of these terms should be tagged, by the ingest pipeline as owl:deprecated. The question you need to answer is:

Are from their intention, these "inactive" classes to be considered deprecated?

nicolevasilevsky:
I spot-checked a couple of the non-rare disorders in Orphanet and they are already xref'd in Mondo. It seems like we should maintain these xrefs.

Orphanet does have obsoleted classes, so I don't think these are considered obsolete in Orphanet. I don't really know what they mean by inactive, maybe they aren't maintained or kept up to date? Should we have a tag in Mondo that notes these are inactive in Orphanet?

matentzn:
> Orphanet does have obsoleted classes

Can you tell me how I can tell from https://www.orphadata.com/ordo/ which classes are obsolete?

[...]

PR #186, opened 2023-02-13
Added "inactive clinical entity" to the orphanet exclusion list

sabrinatoro:
Closes #https://github.com/monarch-initiative/mondo-ingest/issues/183

@joeflack4 I added the parent term "inactive clinical entity" with the "TRUE" to exclude all the children of that term. Please note that this term is however not in the "clinical entity" branch. We can discuss this.

update: based on https://github.com/monarch-initiative/mondo-ingest/issues/187, this PR might not be relevant.


joeflack4:
@sabrinatoro Thanks! I'll merge this now and continue discussion in #183 and #187.

@sabrinatoro @matentzn Note that when I merge something like this, I am not immediately running it and using this new exclusion. Running the pipeline takes a long time (and I don't even think we have a proper `make all` to do that right now). I tend to run everything when I'm asked or if I'm working on a PR and have a big block of time to do that. Would be great if we could get a GItHub action to do this now that (i) Harshad has had some initial success with this, (ii) OAK is addressing some of the memory issues we've been having with NCIT.

Issue #169, opened 2023-02-02
OMIM terms without any symbols should be on the exclude list #411


sabrinatoro:
from https://www.omim.org/help/faq#1_3
No symbol before an entry number generally indicates a description of a phenotype for which the mendelian basis, although suspected, has not been clearly established or that the separateness of this phenotype from that in another entry is unclear.

OMIM terms without any symbols should be on the exclude list, using the exclusion code : MONDO:excludeTrait (@nicolevasilevsky , do you agree with this exclusion code?)
query Mondo for terms with Mondo:EquivalentTo = OMIM terms without symbols (--> these terms should be obsoleted) .

FYI @matentzn @joeflack4

Issue #110, opened 2022-11-18
exclude non-diseases from OMIM slurp


nicolevasilevsky:
https://github.com/monarch-initiative/mondo-ingest/pull/108

⬆️ The OMIM slurp includes a lot of non-diseases and it didn't used to. @sabrinatoro mentioned that certain types of terms (like those in brackets) are not diseases. The non-diseases need to be either manually reviewed and excluded or maybe there is a way to bulk exclude them.

We discussed this on the tech call on 11/18/22

matentzn:
@sabrinatoro can you roughly group the existing non-disease terms into categories (like anatomy, etc) and figure out if there is a way to distinguish them from phenotypes in OMIM somehow? Then the goal here is to use this information for @joeflack4 to add superclass axioms like OMIM:blood subclass Of UBERON:AnatomicalEntity. Then we can exclude them in the exclusion system.

sabrinatoro:
There are a few for which I would like Nicole's input, but I have reviewed the omim list and will update the exclusion list. A few things to keep in mind: 
1) we do have the code MONDO:excludeNonDisease that we can use as a blanket exclusion code if we don't have the time to review
2) @matentzn The terms are grouped for non-disease (based on exclusion code). However, I don't think it is that simple, and I don't know that this will help us much. For example, the blood types are excluded because they are traits. I don't think you will find them in UBERON, but also, I don't know how you would know which omim terms are anatomy (they don't have any information for this). Also, MONDO:excludedgene includes more than "just" genes: it includes pseudogenes, proteins, locus. I don't know that all of these can be caught bases on what is in the omim files (which I am reviewing right now, so we can make more sense of them)

matentzn:
This is now highest priority for you @joeflack4 because we changed curation strategy

PR #60, opened 2022-08-05
Fix empty ICD10 CM

matentzn:
fixes #58

PR #203, opened 2023-02-19
OMIM exclusion updates

joeflack4:
Updates
```
OMIM Exclusions
- Reverted implementation of last commit. Changing from manipulation of the select ntensional exclusions table to instead parse triples and update only the extensional, full exclusion table
- Rename: exclusion_term_expansion.py --> exclusion_table_creation.py: So as not to mislead that all this script is doing is expanding an intensional list of exclusions; it actually does more than that now, such as adding exclusions directly from the ontology when looking for terms with predicate MONDO:exclusionReason.
- Update: exclusion_table_creation.py: To add terms with predicate MONDO:exclusionReason. Also updated the CLI docs
- Regenerated outputs

Misc
- Update: Makefile: help: Added exclusions workflow descriptions.
- Update: docs/developer/workflows.md: Added exclusions workflow descriptions.
```

Related
- https://github.com/monarch-initiative/omim/pull/91


docs/developer/workflows.md
```diff
@@ -28,3 +28,16 @@ These workflows will create a [report of deprecated terms with Mondo xrefs](../r
 2. `mapped-deprecated-terms-artefacts`: Creates a reports/%_mapped_deprecated_terms.robot.template.tsv for all ontologies.
 3. `mapped-deprecated-terms-docs`: Creates a report of statistics for mapped deprecated terms (docs/reports/mapped_deprecated.md) and pages for each ontology which list their deprecated terms with existing xrefs in Mondo.
 4. `mapped-deprecated-terms`: Creates a report of statistics for mapped deprecated terms (docs/reports/mapped_deprecated.md) and pages for each ontology which list their deprecated terms with existing xrefs in Mondo. Also creates a reports/%_mapped_deprecated_terms.robot.template.tsv for all ontologies.
+
+
+## Exclusions
```
joeflack4, line 33: @matentzn @sabrinatoro FYI: Docs
I added some docs for how to update the OMIM exclusions with non-diseases, but I the docs for running the exclusion tables generally were missing, so I added those.

src/ontology/mondo-ingest.Makefile
```diff
@@ -465,3 +475,23 @@ help:
 	@echo "Creates a report of statistics for mapped deprecated terms (docs/reports/mapped_deprecated.md) and pages for each ontology which list their deprecated terms with existing xrefs in Mondo.\n"
 	@echo "mapped-deprecated-terms"
 	@echo "Creates a report of statistics for mapped deprecated terms (docs/reports/mapped_deprecated.md) and pages for each ontology which list their deprecated terms with existing xrefs in Mondo. Also creates a reports/%_mapped_deprecated_terms.robot.template.tsv for all ontologies.\n"
+	@echo "reports/%_term_exclusions.txt"
```
joeflack4, line 478: FYI: `make` `help`: I added some docs for how to update the OMIM exclusions with non-diseases, but I the docs for running the exclusion tables generally were missing, so I added those.

src/scripts/omim_intensional_exclusions_add_non_diseases.py
```diff
@@ -0,0 +1,41 @@
+"""Add non-diseases to OMIM intensional exclusions.
```
joeflack4, line 1: FYI: Script for updating OMIM intensional exclusions file with non-diseases.

src/ontology/mondo-ingest.Makefile
```diff
@@ -224,13 +224,23 @@ $(REPORTDIR)/%_excluded_terms_in_mondo_xrefs.tsv $(REPORTDIR)/%_excluded_terms_i
 	--component-signature-path $(REPORTDIR)/component_signature-$*.tsv \
 	--outpath $@
 
+# Exclusions: specific ontologies
+.PHONY: omim-intensional-exclusions-add-non-diseases
+omim-intensional-exclusions-add-non-diseases:
```
joeflack4, line 229: @matentzn `make` goal for updating OMIM intensional exclusions file with non-diseases.
The file `config/omim_exclusions.tsv` already exists; we just need to update it, hence why this is a phony goal. 

I was going to make a 2nd goal for `config/omim_non_disease_exclusions.tsv`, but I was worried about it not getting updated if it already existed, so I added the download  as an additional step in the 1st / only goal.

docs/developer/workflows.md
joeflack4: @sabrinatoro I don't know if you should / want to review this. I updated the `config/omim_exclusions.tsv` to add non-diseases. Then I ran the exclusion pipeline to generate new artefacts that were updated:

- `reports/omim_exclusion_reasons.robot.template.tsv`
- `reports/omim_term_exclusions.txt`

Feel free to check any of these files. More importantly, you will probably want to look at this [OMIM PR](https://github.com/monarch-initiative/omim/pull/91) where I tagged you on a comment.

I also see this related issue, which I might as well do this week as well. It would fit in quite easily with what I've done here:
- https://github.com/monarch-initiative/mondo-ingest/issues/169

src/ontology/mondo-ingest.Makefile
```diff
@@ -224,13 +224,23 @@ $(REPORTDIR)/%_excluded_terms_in_mondo_xrefs.tsv $(REPORTDIR)/%_excluded_terms_i
...
+
+
 # Exclusions: all artefacts for single ontology
 exclusions-%:
 	$(MAKE) $(REPORTDIR)/$*_term_exclusions.txt
 	$(MAKE) $(REPORTDIR)/$*_exclusion_reasons.ttl
 	$(MAKE) $(REPORTDIR)/$*_excluded_terms_in_mondo_xrefs.tsv
 
-exclusions-all:
+exclusions-all: omim-intensional-exclusions-add-non-diseases
```
joeflack4, line 243: @matentzn I made `omim-intensional-exclusions-add-non-diseases` a dependency of `exclusions-all`. However, I did not add it as a dependency to `$(REPORTDIR)/%_exclusion_reasons.robot.template.tsv`, since that is not OMIM-specific. This means that running OMIM-related exclusions will sometimes update with the latest non-diseases. I'm not really sure how best do do this given our usual design.

docs/developer/workflows.md
joeflack4: @sabrinatoro @matentzn I also see this related issue, which I might as well do this week as well. It would fit in quite easily with what I've done here.

- https://github.com/monarch-initiative/mondo-ingest/issues/169

matentzn: Correct me if I am wrong @joeflack4. You produce a list of non-disease terms at ingest time, which you then _parse_ here, resulting in additional exclusions. 

If this is correct, I think this is slightly hacky. I would prefer:

1. You tag the respective non-diseases with a standard annotation triple in the ingest (e.g. `?x MONDO:exclusionReason MONDO:nonDisease`). 
2. Your generic exclusion script will look for this tag, and use it to produce additional intensional exclusions.
3. No custom scripts for ingest necessary. We can use the same system for all future ingests, tagging classes at ingest time with this triple, and it will just work.

joeflack4: This makes sense. I thought about this to see if there were any disadvantages, and the only one I could think of is that, given the way we set this up now, this makes it less clear to the curator that these things are being excluded, since we have intensional exclusion files, e.g. `config/omim_exclusions.tsv` that are being maintained by the curator / Sabrina at the moment, and these are used to create extensional exclusion files, e.g. `reports/omim_term_exclusions.txt` and `reports/omim_exclusion_reasons.robot.template.tsv`.

This is still my top priority issue which I will likely pick up on again tomorrow, and I can move forward with this change.

joeflack4: donezo

src/scripts/exclusion_table_creation.py
```diff
@@ -276,23 +318,34 @@ def expand_intensional_exclusions(
...
     expanded_intensional_exclusions_df = expand_intensional_exclusions(
-        onto_path=onto_path, exclusions_path=exclusions_path, prefix_map=prefix_uri_map)
+        onto_path=onto_path, exclusions_path=select_intensional_exclusions_path, prefix_map=prefix_uri_map,
+        use_cache=use_cache)
+    # - Get non-inclusions
     non_inclusions_df = get_non_inclusions(
         mirror_signature_path=mirror_signature_path, component_signature_path=component_signature_path,
         prefix_map=prefix_uri_map)
-    df_results = pd.concat([expanded_intensional_exclusions_df, non_inclusions_df])
+    # - Get by triples with MONDO:exclusionReason predicate
```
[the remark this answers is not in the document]
joeflack4: @nicolevasilevsky @matentzn @sabrinatoro Per exclude [non-diseases from OMIM slurp #110](https://github.com/monarch-initiative/mondo-ingest/issues/110), non-diseases are now excluded from OMIM.

#### `MONDO:excludeNonDisease` and `MONDO:excludeGene`
_However_, I just want you guys to know that there are effectively 0 new exclusions from _this_ change. This is because, as far as I can determine, the only terms I can be sure are non-diseases are genes. And, we're already excluding all of these with this single line in `config/omim_exclusions.tsv`:
```tsv
SO:0000704	gene	MONDO:excludeGene	TRUE
```

See this [related discussion in my OMIM PR](https://github.com/monarch-initiative/omim/pull/91#discussion_r1111318316) where I go over which symbols indicate non-disease. I'd like to confirm that I have this correct.

#### `MONDO:excludeTrait`
There _are_ new exclusions though from new `MONDO:excludeTrait` entries. This was the result of [OMIM terms without any symbols should be on the exclude list #169](https://github.com/monarch-initiative/mondo-ingest/issues/169).

src/ontology/mondo-ingest.Makefile
```diff
@@ -29,7 +29,8 @@ $(TMPDIR)/ordo_relevant_signature.txt: component-download-ordo.owl | $(TMPDIR)
 
 # This section is concerned with transforming the incoming sources into the
 # Monarch Ingest schema.
-# https://github.com/monarch-initiative/omim/issues/60
+# Illegal punning on some properties #60: https://github.com/monarch-initiative/omim/issues/60
```
joeflack4, line 32: @matentzn There was a comment w/ a link to [Illegal punning on some properties #60](https://github.com/monarch-initiative/omim/issues/60) here on the `$(COMPONENTSDIR)/omim.owl` goal. I don't remember if I added this, or you. Is this because the illegal punning is being fixed here in `mondo-ingest` but it should eventually be fixed in OMIM?

joeflack4:
@matentzn I think this is good to merge right now, though if you think I did this too early, let me know and I'll keep similar effort PRs open longer.

matentzn:
Refresh content PRs I trust you, but generally a 48 hour PR period is good. I would however like to see https://github.com/monarch-initiative/mondo-ingest/issues/211 prioritised as it makes reviewing these much easier (and therefore faster)

joeflack4:
Thanks for the guidance; will respond in that issue.

Issue #192, opened 2023-02-13
NCIT: `OutOfMemoryError` (can't update `ncit.owl` or exclusions)


joeflack4:
## Overview
Was recently able to fix [an out of memory error for mirror_sig](https://github.com/monarch-initiative/mondo-ingest/issues/171) by switching to OAK. For now this is the only good idea we have for solving this more generally. The result may entail a broad OAK-ification refactor of various scripts since we can (and should) re-use that logic not just for NCIT, but for the other ontologies.

## Sub-task list
- [ ] 1. Fix issue for exclusions
- [ ] 2. Fix issue for `components/ncit.owl`
- [ ] 3. Fix issue anywhere else in pipeline (not sure where if any)

### 1. Fix issue for exclusions
This just entails switching from `robot` calls to OAK.

Would fix `OutOfMemoryError`: [log.txt](https://github.com/monarch-initiative/mondo-ingest/files/10727239/log.txt)

### 2. Fix issue for `components/ncit.owl`
[ncit-error.txt](https://github.com/monarch-initiative/mondo-ingest/files/10819524/ncit-error.txt)

See [discussion below](https://github.com/monarch-initiative/mondo-ingest/issues/192#issuecomment-1442603918).

## Additional details
We tried solving with a very reasonable 20GB of memory in our ODK image, but to no avail:
<img width="863" alt="216852883-eb51646e-b6fd-4a95-913d-bab97cb35171" src="https://user-images.githubusercontent.com/13045020/218599599-e4cab022-a37a-40be-bd3e-2f6a9d1ecf2b.png">

joeflack4:
@matentzn I was rerunning the [slurp today](https://github.com/monarch-initiative/mondo-ingest/pull/210) and, as I suspected might happen, NCIT failed with an out of memory error. 

Here's the part that's hard for me. It failed on `components/ncit.owl`. Up until now, the `components/*.owl` file is what I need to even begin to start using OAK with `mondo-ingest`.

Do you have any ideas as to how I might get around this problem? Should I/we do something like this:
#### Option A
i. Start with `component-download-ncit.owl.owl`
ii. Rewrite the goal for `components/ncit.owl` to one that uses OAK, if that's possible (possibly a heavy lift?)

#### Option B
Use `component-download-ncit.owl.owl` instead of `components/ncit.owl` if that is not a terrible idea.

[...]

PR #214, opened 2023-02-24
Exclusions: NCIT finding branch

joeflack4:
Updates
```
    Exclusions: NCIT
    - Add: NCIT 'Finding' branch root term to select intensional exclusion table.
    - Update: Re-ran exclusion outputs
    - Bugfix: Sorted simple extensional list of terms again
```


src/ontology/config/ncit_exclusions.tsv
```diff
@@ -1,4 +1,5 @@
 term_id	term_label	exclusion_reason	exclude_children
+NCIT:C3367	Finding	MONDO:excludeNonDisease	TRUE
```
joeflack4, line 2: @sabrinatoro This is all that I think needed to be added.

joeflack4: @sabrinatoro I'm surprised though that after doing this, 5,229 less items are showing up in the exclusion tables, rather than more rows. Perhaps this is just because of other recent changes in NCIT.

Note though that I'm not sure how out-of-date what I have for NCIT is. Could be 2 months old? We've been running into [memory issues](https://github.com/monarch-initiative/mondo-ingest/issues/192), so we're stuck with this version of NCIT for now, and I don't know when we'll solve it.

joeflack4 (commented):
@sabrinatoro I'm going to merge this now, but let me know if you spot anything that looks amiss.

PR #215, opened 2023-02-24
Remove object_label from lexical pipeline

matentzn:
@hrshdhgd unfortunately this is not supported by OBO format, so I am removing it. I am assuming that this will prevent this from showing up in the template table?


hrshdhgd (approved):

hrshdhgd:
That's correct.

Issue #227, opened 2023-03-07
Mapping: NCIT term not from the "disease' branch


sabrinatoro:
From here:
https://github.com/monarch-initiative/mondo-ingest/blob/main/src/ontology/lexmatch/unmapped_ncit_lex.tsv

subject_id | subject_label | predicate_id | object_id | object_label | mapping_justification | mapping_tool | confidence | subject_match_field | object_match_field | match_string | comment
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
MONDO:0018079 | thymic epithelial neoplasm | MONDO:equivalentTo | NCIT:C69122 | Ten | semapv:LexicalMatching | oaklib | 0.8 | oio:hasExactSynonym | rdfs:label | ten | LEXMATCH

This match is done on the abbreviation TEN in Mondo. 
The NCIT code is not in the disease branch (it represents "A natural number greater than 9 and less than 11 and the quantity that it denotes: the sum of nine and one.")

This is the only example, therefore it might not be a real issue

sabrinatoro:
Other example from NCIT from "other hierarchy":

MONDO:0018473 | hyperlipoproteinemia type 3 | MONDO:equivalentTo | NCIT:C34710 | Remnant Hyperlipidemia | semapv:LexicalMatching | oaklib | 0.8 | oio:hasExactSynonym | rdfs:label | remnant hyperlipidemia | LEXMATCH |   |   |   |   |   |   |   |   |   |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --

sabrinatoro:
MONDO:0019810 | toxic epidermal necrolysis | MONDO:equivalentTo | NCIT:C69122 | Ten | semapv:LexicalMatching | oaklib | 0.8 | oio:hasExactSynonym | rdfs:label | ten | LEXMATCH
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --

joeflack4:
@matentzn @sabrinatoro It looks like the only reason that these appear as "unmapped mappables" is that the mapping progress monitor has not been run since these exclusions were updated. I re-ran and now they no longer appear there.

**Explanation:**
I checked `reports/ncit_term_exclusions.txt`, and these terms `NCIT:C69122` and `NCIT:C34710` appear in the list, which is good.

[...]

PR #235, opened 2023-03-11
Update: NCIT mapping status artefacts & docs

joeflack4:


joeflack4:
See my comments in issue #227

Issue #226, opened 2023-03-07
Mappings: mapping to a term obsoleted at the source


sabrinatoro:
## Overview
From here: 
https://github.com/monarch-initiative/mondo-ingest/blob/main/src/ontology/lexmatch/unmapped_ncit_lex.tsv

subject_id | subject_label | predicate_id | object_id | object_label | mapping_justification | mapping_tool | confidence | subject_match_field | object_match_field | match_string | comment
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
MONDO:0009691 | mycosis fungoides | MONDO:equivalentTo | NCIT:C27800 | Mycosis Fungoides (Working Formulation) | semapv:LexicalMatching | oaklib | 0.8497788952 | rdfs:label | rdfs:label | mycosis fungoides | LEXMATCH

The NCIT concept is obsoleted according to https://ncithesaurus.nci.nih.gov/ncitbrowser/pages/concept_details.jsf

note: so far, this is the only example, maybe this will be resolved after we re-run the mapping

## Depends on
- #192

sabrinatoro:
other example: 


MONDO:0019468 | T-cell prolymphocytic leukemia | MONDO:equivalentTo | NCIT:C70649 | T-Cell Chronic Lymphocytic Leukemia | semapv:LexicalMatching | oaklib | 0.8 | oio:hasExactSynonym | rdfs:label | t-cell chronic lymphocytic leukemia | LEXMATCH
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --

matentzn:
Important interaction of Joes work with Harshads. I am running a new release now. @joeflack4 this ticket can be closed if Sabrina's examples show up in the exclusion list of the next run.

joeflack4:
Did an initial rerun, and these two terms continue to not show up in the exclusions. I think this is related to #227, so will address that one and come back to this.

matentzn:
I am not sure how we decided to deal with obsolete classes in the the progress monitor, but we do _not want to slurp nor map_ obsolete terms; we do, however, want to _sync_ them up, which is the part of the process that you developed for syncing obsoleted. This is complicated, but makes sense:

If a term is obsolete at the source, there is no point to 
   - map it with Mondo if it is not already mapped.
   - Import it into Mondo if it is not already imported.

[...]

PR #239, opened 2023-03-19
Mapping progress monitor: 'Tot deprecated unmapped' column

joeflack4:
Updates
```
Mapping progress monitor
- Update: docs/reports/unmapped.md: Added column 'Tot deprecated unmapped'
```


docs/reports/unmapped.md
```diff
@@ -1,12 +1,12 @@
...
-| [NCIT](./unmapped_ncit.md)         | 174,300     | 150,309        | 5,055            | 18,936                                    | 6,728                     | 12,208                      | 64.5%                     |
-| [OMIM](./unmapped_omim.md)         | 28,659      | 17,286         | 1,345            | 10,028                                    | 9,779                     | 249                         | 2.5%                      |
-| [ORDO](./unmapped_ordo.md)         | 9,448       | 292            | 0                | 9,156                                     | 8,934                     | 222                         | 2.4%                      |
-| [DOID](./unmapped_doid.md)         | 13,648      | 2,599          | 2,467            | 11,048                                    | 9,748                     | 1,300                       | 11.8%                     |
+| Ontology                           | Tot terms   | Tot excluded   | Tot deprecated   | Tot deprecated unmapped   | Tot mappable _(!excluded, !deprecated)_   | Tot mapped _(mappable)_   | Tot unmapped _(mappable)_   | % unmapped _(mappable)_   |
```
joeflack4, line 2: I re-ran mapping progress docs and got these results:
| Ontology                           | Tot deprecated   | Tot deprecated unmapped |
|:-----------------------------------|:-----------------|:--------------------------|
| [ICD10WHO](./unmapped_icd10who.md) | 0                | 0                         |
| [ICD10CM](./unmapped_icd10cm.md)   | 0                | 0                         |
| [NCIT](./unmapped_ncit.md)         | 5,055            | 5,022                     |
| [OMIM](./unmapped_omim.md)         | 1,345            | 1,255                     |
| [ORDO](./unmapped_ordo.md)         | 0                | 0                         |
| [DOID](./unmapped_doid.md)         | 2,467            | 2,448                     |

src/scripts/unmapped_docs.py
```diff
@@ -58,6 +58,7 @@ def update_mapping_progress_in_docs():
             'Tot terms': f"{len(df):,}",
             'Tot excluded': f"{len(df[df['is_excluded'] == True]):,}",
             'Tot deprecated': n_deprecated,
+            'Tot deprecated unmapped': f"{len(df[(df['is_deprecated'] == True) & (df['is_mapped'] == False)]):,}",
```
joeflack4, line 61: Per [#226](https://github.com/monarch-initiative/mondo-ingest/issues/226#issuecomment-1468044153), I added a new column `Tot deprecated unmapped`.

I also double checked the logic to make sure that obsoletes/deprecated terms are not included in the "unmapped mappables". Indeed, it appears they are not:
```python
        mappable_df = df[(df['is_excluded'] == False) & (df['is_deprecated'] == False)]
        unmapped_mappable_df = mappable_df[mappable_df['is_mapped'] == False]
```

joeflack4 (commented):
@matentzn @sabrinatoro This should, I believe, bring conclusion to #226. Take a look at the markdown table I included here as well.

docs/reports/unmapped.md
joeflack4: We have a high amount of terms that are deprecated/obsolete and mapped. I'm actually not sure though, is this number supposed to be low, high, or doesn't really matter? I'm guessing doesn't really matter.

matentzn (commented):
I expected this column to drastically reduce the number of terms in the unmapped mappable column? It didn't?

src/scripts/unmapped_docs.py
matentzn: Very unfortunate. I hoped these large numbers in the unmapped mappable column would somehow decrease..

joeflack4: Yeah, as I probably mentioned on our last call, I was not under the impression that this was bugged. All I did here was add a new column. It doesn't seem that there is a need to edit the calculations for 'mappable unmapped', so the number didn't change.

If you want, we can do additional checking. Like we could look through these TSVs / dataframes and just double check a bunch of terms and make sure they are being correctly marked as mapped, etc.

matentzn (approved):

PR #250, opened 2023-03-24
Bugfix: Mapping progress angle brackets

joeflack4:
- Bugfix: Removed angle brackets from URIs. This was causing results to not be generated for mapping status artefacts if angle brackets were present.

### DIFF
--- a/src/scripts/unmapped_tables.py
+++ b/src/scripts/unmapped_tables.py
@@ -19,6 +19,8 @@ from oaklib import get_implementation_from_shorthand
 from oaklib.types import CURIE, URI
 from sssom.util import is_curie
 
+from utils import remove_angle_brackets
+
 
 def create_mapping_status_tables(
     db_path: str, exclusions_path: str, mondo_mappings_path: str, onto_config_path: str, outpath_full: str,
@@ -32,7 +34,7 @@ def create_mapping_status_tables(
         prefix_map: Dict[str, str] = onto_config['base_prefix_map']
         owned_prefixes: List[str] = list(prefix_map.keys())
         converter = curies.Converter.from_prefix_map(prefix_map)
-        prefix_preplacement_map = {
+        prefix_replacement_map = {
             f'{alias}:': f'{preferred}:' for preferred, alias in onto_config['prefix_aliases'].items()} \
             if 'prefix_aliases' in onto_config else {}
     excluded_terms: Set[CURIE] = set(pd.read_csv(exclusions_path, header=None, names=['id']).fillna('')['id'])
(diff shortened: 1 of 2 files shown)


joeflack4:
@souzadevinicius @hrshdhgd FYI in case you have time to review.

src/scripts/utils.py
```diff
@@ -183,6 +183,18 @@ def _get_next_available_mondo_id(min_id: int, max_id: int, mondo_ids: Set[int])
     return next_id, mondo_ids
 
 
+def remove_angle_brackets(uris: Union[URI, List[URI]]):
```
joeflack4, line 186: The new `remove_angle_brackets()` func. I've previously written this logic before and been asked by some to leverage some stuff in `curies` or other libraries, so I might refactor to do that. But it is such a simple function to implement.

src/scripts/unmapped_tables.py
```diff
@@ -68,7 +72,7 @@ def create_mapping_status_tables(
         'subject_id': curies_owned, 'subject_label': labels_owned, 'is_mapped': is_mapped, 'is_excluded': is_excluded,
         'is_deprecated': is_deprecated})
     if len(df) == 0:
-        return {'full': df, 'simple': df}
+        raise RuntimeError('Mapping status: No "owned terms" found in the ontology. Not expected.')
```
joeflack4, line 75: @matentzn @hrshdhgd Previously, at this point if there were 0 terms found during this generation of mapping status / list of unmapped terms TSVs for the mapping progress monitor, I was not throwing an error. I realize now that this was a mistake. Now throws an error if no results, which is what I was getting before fixing the angle brackets issue.

src/scripts/unmapped_tables.py
```diff
@@ -41,7 +43,9 @@ def create_mapping_status_tables(
 
     # Get all terms and labels
     ids_all: List[Union[CURIE, URI]] = [x for x in oi.entities(filter_obsoletes=False)]
+    ids_all = remove_angle_brackets(ids_all)
```
joeflack4, line 46: FYI: Usage
`    ids_all = remove_angle_brackets(ids_all)`
`    ids_sans_deprecated = remove_angle_brackets(ids_sans_deprecated)`

src/scripts/utils.py
hrshdhgd: If `semsql make` is provided with all prefixes needed, do you think this will be needed? At that point, the `<>` are ignored by virtue of it's absence in the `curie_map`. On the flip side, some IRIs which may be desired but lost in the process may not be highlighted. @matentzn , thoughts?

src/scripts/unmapped_tables.py
hrshdhgd: I'm assuming (and maybe I'm wrong) this is related to the Slack conversation we had regarding `omit.db`. Did recreating the file eliminate the need of handling `<>` ?

joeflack4: Not prior to running this, but I will be re-running all of these.

I'm reluctant to remove this extra step using `remove_angle_brackets()` again though. I had been doing it in the past, and then I was no longer seeing URIs with angle brackets, so I removed this step, but then they appeared again. It affects performance very little, and I want to account for unknown unknowns / possible issues in the future.

One thing I can possibly see happening is a mismatch between `config/prefixes.csv` and the prefixes in the `metadata/%.yml` files. There's some redundancy there, and I think Nico is aware. We've had conversations before where we bounced between a single file vs multiple files, using an external registry, etc. So until we choose to use only one of those options I am wary of such a mismatch occurring.

matentzn: I think its ok to have this as a failsafe! But we do neet to remember when adding an ingest to add the appropriate prefixes to the various files we work with

@joeflack4 while you are working on adding medgen, can you write a 7 step guide on how to "add an ingest to Mondo INgest" and add it to the docs? This should include making sure the prefixes are up to date as well.

matentzn (approved):

joeflack4:
Merging per approval. Thanks!

PR #252, opened 2023-03-24
Bugfix: Slurp: terms with no parents

joeflack4:
Updates
```
    - Bugfix: Slurp: When a term had no parents, it was not being considered a slurpable / migratable term.
    - Update: Slurp TSVs: Reran artefacts following bugfix
```


joeflack4:
@souzadevinicius @hrshdhgd FYI in case you have time to review.

src/scripts/migrate.py
```diff
@@ -9,8 +9,7 @@
 - https://incatools.github.io/ontology-access-kit/
 - https://incatools.github.io/ontology-access-kit/intro/tutorial02.html
 
-TODO's (major):
-  -
+todo: refactor to take in 'unmapped mapppable' terms i.e. reports/%_unmapped_terms.tsv instead of other params?
```
joeflack4, line 12: FYI: Idea: It may be possible to take in `reports/%_unmapped_terms.tsv ` (generated in step 1 of the mapping progress monitor) as a parameter to `slurp` instead of several other params, simplifying it.

joeflack4 (commented):
@sabrinatoro (and maybe @nicolevasilevsky too?) Following this bugfix I reran `slurp-all` to create new slurp TSVs, and I imagine it needs a review~

src/ontology/slurp/omim.tsv
```diff
@@ -1,56 +1,387 @@
...
+MONDO:0859081	chromosome xq13 duplication syndrome	OMIM:301069	MONDO:equivalentTo	chromosome xq13 duplication syndrome		
+MONDO:0859082	thrombophilia, x-linked, due to factor 8 defect	OMIM:301071	MONDO:equivalentTo	thrombophilia, x-linked, due to factor 8 defect		
+MONDO:0859083	systemic lupus erythematosus 17	OMIM:301080	MONDO:equivalentTo	systemic lupus erythematosus 17		
+MONDO:0859084	epilepsy, x-linked 2, with or without impaired intellectual development and dysmorphic features	OMIM:301091	MONDO:equivalentTo	epilepsy, X-linked 2, with or without impaired intellectual development and dysmorphic features		
+MONDO:0859085	neurodevelopmental disorder with gait disturbance, dysmorphic facies, and behavioral abnormalities, x-linked	OMIM:301094	MONDO:equivalentTo	neurodevelopmental disorder with gait disturbance, dysmorphic facies, and behavioral abnormalities, X-linked		
+MONDO:0859086	intellectual developmental disorder, x-linked 110	OMIM:301095	MONDO:equivalentTo	intellectual developmental disorder, X-linked 110		
+MONDO:0859087	removed from database	OMIM:301880	MONDO:equivalentTo	removed from database
```
matentzn, line 81: What are all these "removed from database" rows?

src/scripts/migrate.py
matentzn: Good idea! Make sure this dependency is explicit in the makefile if you decide to implement this in a future PR

src/scripts/migrate.py
```diff
@@ -90,7 +90,7 @@ def slurp(
                     qualified_parent_mondo_ids.append(parent_mondo_id)
             if parent_curie not in sssom_object_ids and parent_curie in owned_term_curies:
                 break
-        if qualified_parent_mondo_ids:
+        if qualified_parent_mondo_ids or no_parents:
```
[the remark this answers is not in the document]
matentzn: Looks good!

src/ontology/slurp/omim.tsv
joeflack4: I was surprised to see that too. This is a message I see for OMIM deprecated terms, so I'm not sure why they are here. I'll look and double check if these are indeed deprecated and, if so, why they are being included here (they shouldn't be).

matentzn (approved):

matentzn (changes requested):
- [x] The "removed from database" terms in OMIM are still there.

joeflack4:
> The "removed from database" terms in OMIM are still there.
The fix is in. This was a flaw in the original OMIM ingest logic that we did not notice until now. Basically, deprecated terms often get migrated to a new MIM ID. However, if the term was deprecated without being assigned a new MIM ID, it was erroneously not being marked deprecated.

Fix: https://github.com/monarch-initiative/omim/commit/5a057ae532a5253c2a6c35bff10dc59acce32d71#r106303254
New release: https://github.com/monarch-initiative/omim/releases/tag/2023-03-27

matentzn:
Excellent thank you!

PR #253, opened 2023-03-24
Mapping progress update (OMIM): following angle bracket bugfix

joeflack4:
Updates
```
    - Update: Mapping progress monitor: Re-ran all mapping progress artefacts and documentation, following a bugfix where terms were not correctly being parsed due to an <angle brackets> bug
```

Also includes:
- #250


src/ontology/reports/omim_mapping_status.tsv
```diff
@@ -373,6 +373,17 @@ OMIM:620244	lymphatic malformation 13	False	False	False
 OMIM:620245	episodic kinesigenic dyskinesia 3	False	False	False
 OMIM:620247	cardiomyopathy, dilated, 1oo	False	False	False
 OMIM:620253	cataract 50 with or without glaucoma	False	False	False
+OMIMPS:146590		False	False	False
```
joeflack4, line 376: @matentzn @sabrinatoro @nicolevasilevsky After the bugfix in #250 and some help from @hrshdhgd , I recreated all my SemanticSQL DB files and then re-ran the progress monitor. It looks like only OMIM had updates.

I'm kind of surprised the OMIMPS terms weren't there before. I think I had added the prefix for OMIMPS, but I didn't notice this bug until now, so they weren't getting picked up until I fixed it. Kinda surprised the bug only affected OMIM though.

matentzn: Ok, thanks!

matentzn (approved):

joeflack4:
Holding off on merging this pending approval of #250, which this PR also includes.

Issue #256, opened 2023-03-27
Faster builds: `md5`


joeflack4:
## Overview
Mirror of: https://github.com/INCATools/ontology-development-kit/issues/840

However, @matentzn there may be some goals in the `mondo-ingest.Makefile` where we can implement this, independent of changes to ODK.

joeflack4:
Had a first draft PR for this. Closing it, but we can always re-open if needed:
- #267

PR #267, opened 2023-04-01
Check MD5 hash before downloading

joeflack4:
## Overview
Proof of concept: downloading an MD5 file and checking against a local copy of the hash the last time the file was downloaded, then doing something if they are different.

## Addresses
Addresses no GitHub issue yet created. However, addresses the problem of our pipeline needlessly spending time downloading files even when they have not changed.

### DIFF
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -29,9 +29,54 @@ $(TMPDIR)/ordo_relevant_signature.txt: component-download-ordo.owl | $(TMPDIR)
 #######################################
 ### Ingest Components #################
 #######################################
-
 # This section is concerned with transforming the incoming sources into the
 # Monarch Ingest schema.
+
+# TODO: temporarily moved over from ODK Makefile as proof of concept for https://github.com/monarch-initiative/mondo-ingest/issues/256
+# todo: temporarily making static url because md5 not automated yet
+#OMIM_URI=https://github.com/monarch-initiative/omim/releases/latest/download/omim.ttl
+OMIM_URI=https://github.com/monarch-initiative/omim/releases/download/2023-03-27/omim.ttl
+.PHONY: component-download-omim.owl
+# This one works!
+component-download-omim.owl: | $(TMPDIR)
+	if [ $(MIR) = true ] && [ $(COMP) = true ]; then wget $(OMIM_URI).md5 -O $(TMPDIR)/tmp.md5; \
+	if diff $(TMPDIR)/tmp.md5 $(TMPDIR)/omim.ttl.md5 >/dev/null 2>&1; \
+		then echo same; else echo different; fi; fi
+
+# failed attempts
+# https://stackoverflow.com/questions/2019989/how-to-assign-the-output-of-a-command-to-a-makefile-variable
+# - i see DIFF=hi in terminal, but echo is blank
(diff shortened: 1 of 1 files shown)


src/ontology/mondo-ingest.Makefile
```diff
@@ -29,9 +29,44 @@ $(TMPDIR)/ordo_relevant_signature.txt: component-download-ordo.owl | $(TMPDIR)
...
 # This section is concerned with transforming the incoming sources into the
 # Monarch Ingest schema.
+
+# TODO: temporarily moved over from ODK Makefile as proof of concept for https://github.com/monarch-initiative/mondo-ingest/issues/256
+OMIM_URI=https://github.com/monarch-initiative/omim/releases/latest/download/omim.ttl
+.PHONY: component-download-omim.owl
+component-download-omim.owl: | $(TMPDIR)
+	echo couldn't find a way that worked
+
+# failed attempts
```
joeflack4, line 41: @matentzn I had a feeling this first attempt using ShellScript wouldn't work out. I'm not great at ShellScript, much less nesting commands, much less within `make`. I spent a little over an hour trying many variations.

I think I could do this if I was able to refactor the way we're downloading all of our components, perhaps (a) by splitting into multiple goals / introducing additional variables and/or temp files. Even easier if (b) leveraging Python. But, especially if that involves changing ODK, I can't see those methods being approved.

Probably best to close this and not worry about it for now, unless you want to take a stab or have me try out one of the above proposals.

joeflack4: @matentzn Some success!

I should have realized earlier that this was the perfect use case for ChatGPT. Although what it did was not different, I think, than something I saw on StackOverflow, for whatever reason, this time I realized did something different than before. I took the literal code of the `if` statement, rather than putting it in `[`brackets`]` like we are doing with most of our goals that use `if` statements. When I removed the brackets, it worked.

That is, this failed:
```make
	if [ diff $(TMPDIR)/tmp.md5 $(TMPDIR)/omim.ttl.md5 >/dev/null 2>&1 ]; \
		then echo same; else echo different; fi; fi
```

But this succeeded:
```make
	if diff $(TMPDIR)/tmp.md5 $(TMPDIR)/omim.ttl.md5 >/dev/null 2>&1; \
		then echo same; else echo different; fi; fi
```

This is why I don't like ShellScript for anything more than a few lines; it's so arcane. Even something as short as this trips me up.

---

Now that we have a working proof of concept, need to decide what to do next.

One thing we need to do is keep a local copy of these raw downloads, e.g. `omim.ttl`, or at least keep its MD5 the last time it ran, e.g. `$(TMPDIR)/omim.ttl.md5`. Thus, we would need to change it so that `robot` doesn't take as input the URLs of these artefacts (e.g. `$(ROBOT) merge -I https://github.com/monarch-initiative/omim/releases/latest/download/omim.ttl`; we'd have to download it first, if that's alright with you.

matentzn: Lets discuss in next call! the principle sounds great, but I would like it to be explained once face2face

joeflack4:
@twhetzel FYI this is something I'd like to do at some point that would help reduce `build-mondo-ingest` time but especially make things run faster / easier locally. Refer to my comment in OP for what this is about.

joeflack4:
Closing this, but leaving the issue open:
- #256

PR #271, opened 2023-04-07
Components w/ only unmapped mappables

joeflack4:
Updates
- Add: makefile goal: components/%-unmapped.owl: Creates an OWL component of the ontology which consists only of the subset of terms which are current mapping candidates.
- Add: makefile goal: recreate-unmapped-components: Runs components/%-unmapped.owl for all ontologies.
- Update: .gitignore: To allow exception so that these new component files can be committed.
- Add: components/*-unmapped.owl for each ontology


src/ontology/mondo-ingest.Makefile
```diff
@@ -199,6 +199,14 @@ $(REPORTDIR)/%_mapping_status.tsv $(REPORTDIR)/%_unmapped_terms.tsv: $(REPORTDIR
 	--outpath-simple $(REPORTDIR)/$*_unmapped_terms.tsv \
 	--outpath-full $(REPORTDIR)/$*_mapping_status.tsv
 
+components/%-unmapped.owl: components/%.owl reports/%_unmapped_terms.tsv
```
joeflack4, line 202: @matentzn All prereqs keep re-running each time.
I'm not sure why, but even with `IMP=false COMP=false MIR=false`, this goal still re-runs the entire prerequisite tree even if all of the artefacts already exist. I looked a the terminal output, and I see that `if [ false = true ] && [ false = true ]` is occurring where I would expect in the component goal. I don't see `-B` anywhere in any of these prerequisites. I could look a bit harder, but at the moment I'm not sure why it seems like everything is running.

src/ontology/mondo-ingest.Makefile
```diff
@@ -199,6 +199,14 @@ $(REPORTDIR)/%_mapping_status.tsv $(REPORTDIR)/%_unmapped_terms.tsv: $(REPORTDIR
...
+	cut -f 1 reports/$*_unmapped_terms.tsv | tail -n +2 > reports/$*_unmapped_terms.txt
+	$(ROBOT) filter -i components/$*.owl -T reports/$*_unmapped_terms.tsv -o $@
+	rm reports/$*_unmapped_terms.txt
+
+.PHONY: recreate-unmapped-components
+recreate-unmapped-components: $(patsubst %, components/%-unmapped.owl, $(ALL_COMPONENT_IDS))
```
joeflack4, line 208: @matentzn What you had in the GitHub issue for this goal was slightly off. I tried a variation that didn't work, then asked ChatGPT and the robot overlord succeeded!

joeflack4 (commented):
@hrshdhgd @souzadevinicius Please review if you guys have time~

src/ontology/mondo-ingest.Makefile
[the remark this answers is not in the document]
matentzn: We will just check them in!

matentzn: Which is the first goal that is being re-run? So when you run a target, which is the very first goal that is rerun although it shouldn't?

src/ontology/mondo-ingest.Makefile
```diff
@@ -199,6 +199,14 @@ $(REPORTDIR)/%_mapping_status.tsv $(REPORTDIR)/%_unmapped_terms.tsv: $(REPORTDIR
 	--outpath-simple $(REPORTDIR)/$*_unmapped_terms.tsv \
 	--outpath-full $(REPORTDIR)/$*_mapping_status.tsv
 
+components/%-unmapped.owl: components/%.owl reports/%_unmapped_terms.tsv
+	cut -f 1 reports/$*_unmapped_terms.tsv | tail -n +2 > reports/$*_unmapped_terms.txt
```
matentzn, line 203: Never overwrite another file in a make goal - either add this to the reports/%_unmapped_terms.tsv goal, or create an entirely new file here that you feed into the filter.

src/ontology/mondo-ingest.Makefile
matentzn: I am sure :P Great job here!

src/ontology/mondo-ingest.Makefile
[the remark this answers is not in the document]
joeflack4: Got it. I just added `!src/ontology/components/*-unmapped.owl` to the `.gitignore`, reran, and they are now committed.

src/ontology/mondo-ingest.Makefile
joeflack4: Hmm, makes sense, but technically I'm actually not overwriting anything (you might have misread `.tsv` as `.txt` in the first part of the goal). As in your latter suggestion, I'm creating a new file and then feeding it to the filter.

The whole goal is:
```make
	$(ROBOT) filter -i components/$*.owl -T reports/$*_unmapped_terms.tsv -o $@
	rm reports/$*_unmapped_terms.txt
```

We didn't previously have `reports/*_unmapped_terms.txt`. We had `reports/*_mapping_status.tsv` (`subject_id`, `subject_label`, `is_mapped`, `is_excluded`, `is_deprecated`) and `reports/*_unmapped_terms.tsv` (`subject_id`, `subject_label`). So, the only goal which uses this new `reports/*_unmapped_terms.txt` is `components/%-unmapped.owl`; So I just create it on the fly and then delete it afterwards.

Wondering if you would like me to:
(a) keep doing this
(b) Put it in `tmp/` instead since it's just deleted immediately
(c) If we want to commit `reports/*_unmapped_terms.txt` as well, I can add it to the same goal that makes those other 2 mapping artefacts.

src/ontology/mondo-ingest.Makefile
joeflack4: Good suggestion. I took a look, and it's: `reports/mirror_signature-%.tsv`

It depends on `component-download-%.owl`. Hmm... Maybe the issue here is that these component download goals are phony. Is this what triggers the whole pipeline to run? 

If this is the case, then even my suggestion with the MD5 stuff won't fix it. Is there any way that we can make these not phony?

matentzn: There is a bug somewhere I think, I noticed that as well. This is how it should work:

.PHONY goals for download always get triggered. They then create the downloaded file somewhere, but _only if it has changed_. Example:

https://github.com/obophenotype/bio-attribute-ontology/blob/master/src/ontology/Makefile#L423

Note that goal that creates the file works like this: https://github.com/obophenotype/bio-attribute-ontology/blob/master/src/ontology/Makefile#L567

So, `myfile.owl` depends on `mirror/chebi.owl` (as its only dependency). mirror/chebi.owl triggers the phony `mirror-chebi`. This may or may not change `mirror/chebi.owl` (using file equivalence as a proxy). If `mirror/chebi.owl`, `myfile.owl` will not be rebuild either. Somewhere this logic is being disrupted..

joeflack4: Ah, I looked at the code and your explanation and I see how this works. I'll compare to what's going on here briefly with `mondo-ingest` and see if I can easily find a difference.

PR #282, opened 2023-04-14
Bugfix: `components/%-unmapped.owl`

joeflack4:
Bugfix: components/%-unmapped.owl: It wasn't actually filtering because of a typo in the goal.

### DIFF
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -201,7 +201,7 @@ $(REPORTDIR)/%_mapping_status.tsv $(REPORTDIR)/%_unmapped_terms.tsv: $(REPORTDIR
 
 components/%-unmapped.owl: components/%.owl reports/%_unmapped_terms.tsv
 	cut -f 1 reports/$*_unmapped_terms.tsv | tail -n +2 > reports/$*_unmapped_terms.txt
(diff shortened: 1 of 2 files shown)


src/ontology/mondo-ingest.Makefile
```diff
@@ -201,7 +201,7 @@ $(REPORTDIR)/%_mapping_status.tsv $(REPORTDIR)/%_unmapped_terms.tsv: $(REPORTDIR
 
 components/%-unmapped.owl: components/%.owl reports/%_unmapped_terms.tsv
 	cut -f 1 reports/$*_unmapped_terms.tsv | tail -n +2 > reports/$*_unmapped_terms.txt
-	$(ROBOT) filter -i components/$*.owl -T reports/$*_unmapped_terms.tsv -o $@
+	$(ROBOT) filter -i components/$*.owl -T reports/$*_unmapped_terms.txt -o $@
```
joeflack4, line 204: @matentzn This was the bugfix. -_-. I was filtering on the `.tsv` and not the `.txt`. I checked the DOID output and the obsolete terms I checked for are gone.

I reran for all ontologies, but only DOID, NCIT, and OMIM are updated. I'm sort of confused by that and wondering if I should be concerned. I can do a little analysis but first--do you have any thoughts?

matentzn: Nope, sounds right to me! Only these have deprecated terms!

PR #287, opened 2023-04-18
Sort SSSOM after parsing.

### LINKED ISSUE #286: Sort SSSOM outputs for smaller diffs
Make sure that sssom parse results in mondo ingest are sorted (columns and rows) after being created, for example:

https://github.com/monarch-initiative/mondo-ingest/blob/main/src/mappings/omim.sssom.tsv

hrshdhgd:
Fixes #286


matentzn:
Even an empty sssom TSV should be valid!

hrshdhgd:
Fixed it in sssom .

Issue #279, opened 2023-04-14
Make sure all OMIMPs terms have to have children


matentzn:
See https://github.com/monarch-initiative/mondo-ingest/blob/main/src/ontology/slurp/omim.tsv

E.g. [congenital myopathy 123](https://omim.org/entry/619967) (`OMIM:619967`)should be child of its phenotypic series, but is not

joeflack4:
Made progress on this on Saturday. Should be able to pick up on this / complete it this Wednesday or Thursday.

PR #290, opened 2023-04-26
Bugfix: OMIMPS children

joeflack4:
Bugfix: OMIMPS children
- Bugfix: Where OMIMPS children weren't present
- Bugfix: Slurp cache was being used. This is what caused the above bug.
- Bugfix: When robot is not found in path, it will try a specific path. This won't be a bugfix for most people. This is more of an issue with Joe's specific setup.
- Update: Now removes cached files if --use-cache is False. There are times where it could be good to retain an old cache, but opting to do this for the following bugfix reason.
- Bugfix: Delete files from cache if --use-cache is False. This is a quick way to fix potential edge case where running the a jinja_sparql() encounters an issue when running a subprocess that does not result in an error, but also does not result in successful output being written. In that case, previously cached results were erroneously being re-read.

### DIFF
--- a/src/scripts/utils.py
+++ b/src/scripts/utils.py
@@ -4,7 +4,6 @@ import pickle
 import subprocess
 import sys
 from datetime import datetime
-from glob import glob
 from typing import Dict, List, Set, Union
 
 import curies
@@ -280,7 +279,7 @@ def _get_all_owned_terms(
         direct_parents_df: pd.DataFrame = jinja_sparql(
             onto_path=ontology_path,
             jinja_path=os.path.join(PROJECT_DIR, 'src', 'sparql', 'get-terms-direct-parents.sparql.jinja2'),
-            prefix_map=owned_prefix_map, use_cache=True, values=' '.join(query_terms))
+            prefix_map=owned_prefix_map, use_cache=use_cache, values=' '.join(query_terms))
         direct_parents_df = direct_parents_df.applymap(uri_converter.compress)
         direct_owned_parents_map = {}
         for _index, row in direct_parents_df.iterrows():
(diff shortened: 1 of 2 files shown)


src/scripts/utils.py
```diff
@@ -280,7 +279,7 @@ def _get_all_owned_terms(
         direct_parents_df: pd.DataFrame = jinja_sparql(
             onto_path=ontology_path,
             jinja_path=os.path.join(PROJECT_DIR, 'src', 'sparql', 'get-terms-direct-parents.sparql.jinja2'),
-            prefix_map=owned_prefix_map, use_cache=True, values=' '.join(query_terms))
+            prefix_map=owned_prefix_map, use_cache=use_cache, values=' '.join(query_terms))
```
joeflack4, line 282: @matentzn This was the main issue. I feel bad about this somehow slipping in. It was previously always using the cache. The cache has still saved me a lot of time in the past, but this one did take a few hours for me to fix as well as you all's attention.

src/scripts/utils.py
```diff
@@ -394,13 +393,25 @@ def jinja_sparql(
...
             f.write(instantiated_str)
         with open(command_save_path, 'w') as f:
             f.write(command_str)
-        result = subprocess.run(command_str.split())
+        try:
+            result = subprocess.run(command_str.split())
+        except FileNotFoundError as e:
+            if 'robot' in str(e):
+                # joeflack4 2023/04/25: Suddenly my PATH is wrong. Could be virtualenvwrapper+PyCharm issue.
+                command_str = command_str.replace('robot query', '/usr/local/bin/robot query')
```
joeflack4, line 411: There's a strange issue on my end that I could not quickly solve. Without this change, I was not able to run things, so leaving it this way for now but we can discuss if you want. This only happens when running in PyCharm (for debugging). When I look at my `PATH` in my terminal with my virtualenv active, it is correct.
>                 `# joeflack4 2023/04/25: Suddenly my PATH is wrong. Could be virtualenvwrapper+PyCharm issue.`

joeflack4 (commented):
I don't know how you guys came across https://omim.org/entry/607628 specifically as an issue ([GH issue](https://github.com/monarch-initiative/mondo-ingest/issues/279)). I still don't see it appearing as a new entry in `slurp/omim.tsv`. Is it expected to be there now?

I did, however, see that it was not appearing as a result of an intermediate SPARQL query, and it did appear after I fixed the caching issue.

matentzn (approved):
I will test this now!

PR #292, opened 2023-04-27
Bugfix: Migrate: `FileNotFoundError`

joeflack4:
Bugfix: Migrate: Tried to remove files that might not have been there, resulting in FileNotFoundError. Also 2 out of 3 of the file removals seemed redundant, so only removing 1 of them now.

### DIFF
--- a/src/scripts/utils.py
+++ b/src/scripts/utils.py
@@ -396,9 +396,8 @@ def jinja_sparql(
     # - If return from cache, empties existing cache before running
     os.makedirs(results_dirpath, exist_ok=True)
     if not (os.path.exists(results_path) and use_cache):
-        os.remove(instantiated_query_path)
-        os.remove(command_str)
-        os.remove(results_path)
+        if os.path.exists(results_path):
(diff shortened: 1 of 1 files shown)


src/scripts/utils.py
```diff
@@ -396,9 +396,8 @@ def jinja_sparql(
     # - If return from cache, empties existing cache before running
     os.makedirs(results_dirpath, exist_ok=True)
     if not (os.path.exists(results_path) and use_cache):
-        os.remove(instantiated_query_path)
-        os.remove(command_str)
-        os.remove(results_path)
+        if os.path.exists(results_path):
```
joeflack4, line 399: @matentzn My bad, totally forgot to check if the file exists before removing it. This might even be excessive edge case handling; the file removal may not even be necessary. But just want to prevent the case where there is some cached results of the SPARQL query present and it uses those even if we're not passing a `--use-cache` param.

joeflack4 (commented):
I'm going to go ahead and merge this now because it feels very striaghtforward.

PR #293, opened 2023-04-27
Deletion of redundant SemSQL intermediaries

joeflack4:
Update: Makefile goals for SemanticSQL .db's: Deleting redundant `%-relation-graph.tsv.gz` files.

### DIFF
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -391,9 +391,11 @@ tmp/merged.owl: tmp/mondo.owl mondo-ingest.owl tmp/mondo.sssom.ttl
 tmp/merged.db: tmp/merged.owl
 	@rm -f .template.db
 	@rm -f .template.db.tmp
+	@rm -f tmp/merged-relation-graph.tsv.gz
 	RUST_BACKTRACE=full semsql make $@ -P config/prefixes.csv
 	@rm -f .template.db
 	@rm -f .template.db.tmp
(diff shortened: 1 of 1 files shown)


src/ontology/mondo-ingest.Makefile
```diff
@@ -391,9 +391,11 @@ tmp/merged.owl: tmp/mondo.owl mondo-ingest.owl tmp/mondo.sssom.ttl
 tmp/merged.db: tmp/merged.owl
 	@rm -f .template.db
 	@rm -f .template.db.tmp
+	@rm -f tmp/merged-relation-graph.tsv.gz
```
joeflack4, line 394: I noticed some lingering `%-relation-graph.tsv.gz` files, so also deleting them along with these other files we were deleting before. I don't know why, but the `.template.db` and `.template.db.tmp` files were appearing in the calling directory `src/ontology/`, whereas `%-relation-graph.tsv.gz` were appearing in the directory where the final `.db` file was being created.

Not really a big issue, but it's is getting a little verbose. Regarding my justification to remove them before and after: Just want to delete them beforehand just to prevent any possible interaction of these lingering artefacts with current or future version of `semsql`. And removing them afterwards just for good cleanup.

FYI I made an issue about this in the SemanticSQL repo some time ago: 
- https://github.com/INCATools/semantic-sql/issues/47

joeflack4 (commented):
@hrshdhgd Just want your eyes on this as well since you too are a SemSQL user.

matentzn (approved):

PR #311, opened 2023-05-17
Ran `update_repo`

joeflack4:
All of these changes were made programmatically by running 'sh run.sh make update_repo'
- Update: Ontologies now alphabetically sorted. Changes propagated to the Makefile
- Update: Some misc spelling / grammar / docstyle changes
- Update: Github workflow: actions/checkout v2 -> v3
- Update: ODK version: v1.4 -> v1.5-dev
- Add: goal: test_fast
- Update: run.sh: A block of code regarding OWL API conversion


src/ontology/Makefile
```diff
@@ -75,6 +75,10 @@ all_odk: odkversion test all_assets
 test: odkversion reason_test sparql_test robot_reports $(REPORTDIR)/validate_profile_owl2dl_$(ONT).owl.txt
 	echo "Finished running all tests successfully."
 
+.PHONY: test
+test_fast:
```
joeflack4, line 79: @matentzn I thought that `update_repo` only made updates based on `mondo-ingest-odk.yaml` based on what you said the other day, but I think what I should have taken away is that I will almost never need to use it unless I'm making changes to that file.

There are a few other changes not related to my `mondo-ingest-odk.yaml` updates that you can see in the diff, if you don't mind taking a look.

docs/odk-workflows/RepositoryFileStructure.md
```diff
@@ -30,9 +30,9 @@ These are the components in MONDO-INGEST
 
 | Filename | URL |
 | -------- | --- |
+| doid.owl | http://purl.obolibrary.org/obo/doid.owl |
```
joeflack4, line 33: @matentzn These are the changes related to my updates to `mondo-ingest-odk.yaml`. I just sorted the list of ontologies alphabetically. This made similar ordering changes as well to the `Makefile`.

joeflack4 (commented):
@hrshdhgd @souzadevinicius FYI. Not sure if you are familiar. I haven't set up other projects that use ODK, but the primary thing going on here is that there is an ODK YAML `mondo-ingest-odk.yaml`. When I make edits to that, I run `sh run.sh make update_repo` from `src/ontology`, and it propagates those config changes appropriately. It looks like it also does some other things.

src/ontology/Makefile
matentzn: Yes, this is because you get updates that are only available in the ODK `:dev` tag. Its fine!

matentzn (approved):

Issue #283, opened 2023-04-14
GARD : Mondo mapping


sabrinatoro:
I reviewed https://docs.google.com/spreadsheets/d/1w5Xnzr5uNFcPrQqCT8mGBFHnGhwXDfQVzHKofw6kB7c/edit#gid=244617092

1) the created mapping (when there is a MONDO:id) are correct for all the one I checked
examples: 
subject_id | in_gard | in_mondo | status | GARD->Orphanet | GARD->Orphanet->OMIM | GARD->Orphanet->Mondo
-- | -- | -- | -- | -- | -- | --
GARD:10001 | TRUE | FALSE | gard | Orphanet:53689 | OMIM:214700 | MONDO:0008964
GARD:10045 | TRUE | FALSE | gard | Orphanet:79303 | OMIM:235555 | MONDO:0009339
GARD:10061 | TRUE | FALSE | gard | Orphanet:50809 | OMIM:609655 | MONDO:0012330

2) in the cases where multiple OMIM are associated for 1 GARD and/or 1 Orphanet: 
These OMIMs id represents children terms of the Mondo term represented by the corresponding GARD/ORDO. (Note that the Mondo term is mapped to GARD/Orphanet correctly for the terms I checked). 
Therefore, if one needs to make a mapping between omim and gard, these omim ids would represent more specific concept than GARD. (omim = skos:narrowMatch of GARD ??I can never remember if it is the correct way to look at it??)
Examples: 
subject_id | in_gard | in_mondo | status | GARD->Orphanet | GARD->Orphanet->OMIM | GARD->Orphanet->Mondo
-- | -- | -- | -- | -- | -- | --
GARD:10413 | TRUE | FALSE | gard | Orphanet:730 | OMIM:600666\|OMIM:613095\|OMIM:173900 | MONDO:0004691
GARD:156 | TRUE | FALSE | gard | Orphanet:588 | OMIM:613153\|OMIM:613150\|OMIM:613154\|OMIM:615350\|OMIM:253800\|OMIM:615181\|OMIM:236670\|OMIM:253280 | MONDO:0018939

3) I noticed that some terms do not have a Mondo mapping. Most (maybe all?) of these have the same orphanet and omim ID (the number is the same for the orphanet and omim id). I think the orphanet IDs are incorrect because I couldn't find them in orphanet. 
The OMIM ID do match correctly to the Mondo term for which this omim id is an exact match
Examples: 
[...]

PR #312, opened 2023-05-18
Analysis: Slurp working correctly?

joeflack4:
Temp: An analysis of slurp using ORDO as an example just to make sure that slurp is working correctly.

@matentzn Don't merge this. Unless you or Sabrina et al want me to actually create these new `*_mapping_status-with_slurp_info.tsv` files every time. Could be useful.


src/ontology/reports/ordo_mapping_status-with_slurp_info.tsv
```diff
@@ -0,0 +1,10867 @@
+subject_id	subject_label	is_mapped	is_excluded	is_deprecated	possible_slurp_candidate	parents	qualified_parent_mondo_ids	actual_slurp_candidate
```
joeflack4, line 1: @sabrinatoro @matentzn Take a look at [src/ontology/reports/ordo_mapping_status-with_slurp_info.tsv](https://raw.githubusercontent.com/monarch-initiative/mondo-ingest/bugfix-slurp-missing-non-excluded/src/ontology/reports/ordo_mapping_status-with_slurp_info.tsv)

joeflack4: I copy/pasted this information int he corresponding issue: https://github.com/monarch-initiative/mondo-ingest/issues/283#issuecomment-1552356970

PR #313, opened 2023-05-18
Bugfix: Bug `KeyError: 'mondo_id'` in `migrate.py`

joeflack4:
- Bugfix: Slurp/Migrate: Fixed issue where when there were no results it would try to sort an empty data frame and result in KeyError on the first column (i.e. mondo_id).

### DIFF
--- a/src/scripts/migrate.py
+++ b/src/scripts/migrate.py
@@ -128,8 +128,9 @@ def slurp(
 
     # Sort, add robot row, save and return
     result = pd.DataFrame(terms_to_slurp)
-    result = result.sort_values(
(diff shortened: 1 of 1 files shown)


src/scripts/migrate.py
```diff
@@ -128,8 +128,9 @@ def slurp(
 
     # Sort, add robot row, save and return
     result = pd.DataFrame(terms_to_slurp)
-    result = result.sort_values(
-        ['mondo_id', 'mondo_label', 'xref', 'xref_source', 'original_label', 'definition', 'parents'])
+    if len(result) > 0:
```
joeflack4, line 131: Simple bugfix. Sorry, I try to remember to do this check where I see if the dataframe is empty before trying to sort it, but I forgot to do so here.

Issue #314, opened 2023-05-18
ERROR: `max() arg is an empty sequence`


matentzn:
```
python3 ../scripts/migrate.py \
--ontology-path components/ncit.owl \
--mondo-mappings-path tmp/mondo.sssom.tsv \
--onto-config-path metadata/ncit.yml \
--onto-exclusions-path reports/ncit_term_exclusions.txt \
--min-id 850056 \
--max-id 999999 \
--mondo-terms-path reports/mirror_signature-mondo.tsv \
--slurp-dir-path slurp/ \
--outpath slurp/ncit.tsv
/usr/local/lib/python3.10/dist-packages/rdflib_jsonld/__init__.py:9: DeprecationWarning: The rdflib-jsonld package has been integrated into rdflib as of rdflib==6.0.0.  Please remove rdflib-jsonld from your project's dependencies.
  warnings.warn(
/usr/local/lib/python3.10/dist-packages/pronto/ontology.py:283: NotImplementedWarning: cannot process plain `owl:AnnotationProperty`
  cls(self).parse_from(_handle)  # type: ignore
Traceback (most recent call last):
  File "/work/src/ontology/../scripts/migrate.py", line 227, in <module>
    cli()
  File "/work/src/ontology/../scripts/migrate.py", line 223, in cli
    return slurp_docs() if docs else slurp(**d)
  File "/work/src/ontology/../scripts/migrate.py", line 77, in slurp
    next_mondo_id = max(next_mondo_id, max(slurp_ids) + 1)
ValueError: max() arg is an empty sequence
make[2]: *** [mondo-ingest.Makefile:439: slurp/ncit.tsv] Error 1
```

matentzn:
I think we need some kind of integration testing for our pipeline.

joeflack4:
True, I think we do. I think in this case an integration test would pass, because we'd be setting the inputs in stone, and it wouldn't catch this edge case. I think what's causing it is another similar issue where there are no slurp candidates. So for that, an unit test would be just as good / easier I think.

I'll get to fixing this edge case, sorry.

joeflack4:
FYI, this could possibly be problematic / impossible for me to troubleshoot directly because I run out of memory when creating NCIT artefacts locally.

PR #315, opened 2023-05-19
Bugfix: Slurp: `max() arg is an empty sequence`

joeflack4:
- Bugfix: Fixed issue where during slurp we were collecting any assigned mondo IDs already present in the slurp files so as not to assign duplicate IDs. However, during this process, if a slurp file was empty, this was causing an error.

### DIFF
--- a/src/scripts/migrate.py
+++ b/src/scripts/migrate.py
@@ -74,7 +74,7 @@ def slurp(
     for f in slurp_files:
         slurp_df = pd.read_csv(os.path.join(slurp_dir_path, f), sep='\t')
         slurp_ids = [int(x.split(':')[1]) for x in list(slurp_df['mondo_id'])[1:]]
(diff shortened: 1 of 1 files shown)


src/scripts/migrate.py
```diff
@@ -74,7 +74,7 @@ def slurp(
     for f in slurp_files:
         slurp_df = pd.read_csv(os.path.join(slurp_dir_path, f), sep='\t')
         slurp_ids = [int(x.split(':')[1]) for x in list(slurp_df['mondo_id'])[1:]]
-        next_mondo_id = max(next_mondo_id, max(slurp_ids) + 1)
+        next_mondo_id = max(next_mondo_id, max(slurp_ids) + 1) if slurp_ids else next_mondo_id
```
joeflack4, line 77: @matentzn addresses:
- #314

joeflack4 (commented):
Done

PR #316, opened 2023-05-19
Various slurp / migrate updates

joeflack4:
## Overview
Corrects some previous bugs and adds in some new logic that should have been there before. Also simplifies code. Adds some documentation.

## Commits
`d1e6a3b79b4b6e8413a80aa4a15a502a4e0891d7`
    - Update: Parent logic has been re-added by default, but changed.
      - CLI arg --parent-conditions-on now changed to --parent-conditions-off. This defaults to False unless the flag is present.
      - Logic now includes checking parent conditions unless --parent-conditions-off
      - Logic for _valid_parent_conditions() (renamed from _check_parent_conditions()), changed from '(all_parents_mapped and qualified_parent_mondo_ids) or no_parents or no_valid_mondo_parents' to 'all parents should either be (a) mapped, (b) obsolete, or (c, new) excluded'.
      - `parents` column in output TSV: Now only shows parents that have a skos:exactMatch or skos:broadMatch in Mondo
    - Update: Simplification: Several inputs removed, and now uses `reports/%_unmapped_terms.tsv` as an input. This includes a list of terms that has already been prefiltered for being unmapped, non-deprecated, and non-excluded, so this logic doesn't need to be done again in the igration script.
    - Update: `slurp/*.tsv` files after re-running

`d8743412650f62ebf487842cdcbed3d2370c3377`
    - Update: Docs: docs/developer/workflows.md: (i) Explained how the process works, including how it is determined that a term be slurpable / migratable, (ii) Updated codestyle of section: line length.
    - Update: Code comments: Concerining slurpable/migratable terms, corrected some incorrect and contradictory statements, and consolidated them, and made the verbiage the sam as what is in the docs.
    - Bugfix: If all of a term's parents are marked obsolete in Mondo, the term is slurpable.
    - Bugfix: If any of the term's parents are not mapped, it should not have been slurpable.
    - Bugfix: Qualified parents must only be those where their existing mapping is one of skos:exactMatch or skos:narrowMatch.
    - Add: Param: parent_conditions_on: defaulting to False. Basically, when this param is False, the end result is that the terms in `slurp/%.tsv` will be exactly the same as `reports/%_unmapped_terms.tsv`, which is the same as the list of terms in `reports/%_mapping_status.tsv` where `is_mapped`, `is_deprecated`, and `is_obsolete` are False.
- Update: Removed variable owned_term_curies and its usage because it was duplicative; where it was used, it had already been verified that a term's parents were 'owned' by the ontology.


docs/developer/workflows.md
```diff
@@ -14,12 +14,22 @@ Workflows should all be run from the `src/ontology/` directory.
 ---
 
 ## Slurp / migration
-These workflows will determine slurpable / migratable terms. That is, terms that are candidates for integration into Mondo.
+These workflows will determine slurpable / migratable terms. That is, terms that are candidates for integration into 
+Mondo.
+
+#### What determines what is a slurpable / migratable term?
+To be migratable, the term (i) must not already be mapped, (ii) must not be excluded (e.g. not in
```
joeflack4, line 21: @sabrinatoro @matentzn I don't know if this is the best place for it, but I added details about what makes a term slurpable / migratable to the documentation.

src/scripts/migrate.py
```diff
@@ -96,25 +96,31 @@ def slurp(
...
+                if parent_mondo_label.startswith('obsolete'):
+                    obsolete_mondo_parent_ids.append(parent_mondo_id)
+                elif parent['predicate_id'] in ['skos:exactMatch', 'skos:narrowMatch']:
                     qualified_parent_mondo_ids.append(parent_mondo_id)
-            if parent_curie not in sssom_object_ids and parent_curie in owned_term_curies:
-                break
-        if qualified_parent_mondo_ids or no_parents:
+
+        no_valid_mondo_parents: bool = len(t.direct_owned_parent_curies) == len(obsolete_mondo_parent_ids)
+        if (all_parents_mapped and qualified_parent_mondo_ids) or no_parents or no_valid_mondo_parents:
```
joeflack4, line 123: @matentzn Here's the if statement where it checks to make sure the conditions for determining if the term should be added to slurp. Most importantly, can you carefully look over what I wrote in the docs? And then if you want to double check my logic here to make sure it matches, feel free. But I think it is correct. 

Some things in particular that I want you to know:
1. Bugfix: Obsolete parents: I believe I fixed correctly as instructed. (i) if just some of the parents are obsolete in Mondo, a term can still be slurped, (ii) if _all_ of the parents are obsolete in Mondo, it will it will go in the slurp.
2. Bugfix: All parents must be mapped: There was a bug here, sorry. If a parent was not mapped, the loop would previously break, but this would not always have the intended effect. So some things may have slipped into slurp previously that shouldn't have been there.
3. Bugfix: `skos:exactMatch` or `skos:narrowMatch`: I wrote in the comments that at least 1 of the parents must be mapped with one of these mapping predicates, but I never added the check. It considered any mapping to be OK. In practice, I don't know if this would have happened very often, but the logic is there now to catch it.

joeflack4:
@nlharris @souzadevinicius FYI, just tagging you guys to review another of my PRs. I think it's not quite as important to look at the particulars of the bugfixes / code updates here. But if we are to be building redundancy, then you should probably look at [these docs updates](https://github.com/monarch-initiative/mondo-ingest/pull/316/files/a5aafe936efadd3097fd81a08d7a2eb3aa142ebc#diff-d5a57dc86cd20c4b192e7aaf03833951867f8a966548749b7dcce598d3c17efc) regarding the `mondo-ingest` migration (formerly known as 'slurp') pipeline. It details the specific conditions under which a term can be validly integrated into Mondo.

docs/developer/workflows.md
matentzn: This is great!

docs/developer/workflows.md
```diff
@@ -14,12 +14,22 @@ Workflows should all be run from the `src/ontology/` directory.
...
-These workflows will determine slurpable / migratable terms. That is, terms that are candidates for integration into Mondo.
+These workflows will determine slurpable / migratable terms. That is, terms that are candidates for integration into 
+Mondo.
+
+#### What determines what is a slurpable / migratable term?
+To be migratable, the term (i) must not already be mapped, (ii) must not be excluded (e.g. not in 
+`reports/%_term_exclusions.txt`), and (iii) must not be deprecated / obsolete. Then, (iv) there are conditions 
+concerning its parents. It must either (a) have no parents, or (b) have no valid parents in Mondo (i.e. all of its 
+parent terms are marked obsolete in Mondo), or (c) all its parents must be mapped, and at least 1 of those parent's 
+mappings must be one of `skos:exactMatch` or `skos:NarrowMatch`.
```
matentzn, line 25: The better way to phrase this I think is that:

```
Any existing parents must be either mapped, or deprecated.
```

This implies the special cases you list there:
- no parents
- parents obsolete
- parents mapped

The way you phrase it, you are missing the case where 1 parent is obsolete, another mapped (neither (a), (b), (c) applies in that case).

joeflack4: I think you may be onto something, I thought about both the logic and phrasing a few times and it still feels clunky.

However, can you double check that I've implemented correctly? What you are suggesting now is different from waht I was under the impression you were asking for in the past:
(i) the term itself can't be deprecated/obsolete in order to be slurped
(ii) there is no check to see if any of the term's parents are marked obsolete within their ontology
(iii) the term's parents, when checking if they are "qualified parents", it should not be considered qualified if that parent is mapped to a Mondo term which is obsoleted.

This is implemented by looking for the parent term in under `object_id` in `mondo.ssom.tsv` and then checking `row['subject_id'].startswith('obsolete')`.

The `parents_obsolete` column in the [slurp integrity spreadsheet](https://docs.google.com/spreadsheets/d/1kc_RGzdRye7nnCLKYpG0DBKTBQxgezd_0MjJell1Q2c/edit#gid=1042657882) would be more appropriately called `parent_mondo_mappings_obsolete`. Maybe it might be good to look at some of those entries in the spreadsheet to help determine if this logic is currently codified correctly, or if it's even needed at all, or otherwise needs to be somehow changed.

matentzn: Column "L" should be more like `all_parents_mapped_or_obsolete_in_mondo`, else LGTM.

I would say though that the logic is now expressed in a very complex manner. Everything seems right to me apart from the "valid parents" condition. This should IMO look more like:

```
def _valid_parent(parents: List)->Bool:
    for p in parents:
        if !is_mapped(p) and !is_deprecated_in_mondo(p):
             return False
     return True
```

joeflack4: Hmm, actually if "all parents mapped", that also implies "all parents mapped or obsolete in Mondo", for in order for the parent to be obsolete in Mondo, it must be mapped in Mondo. Therefore, this would be logically equivalent to any of the below:


```py
def _valid_parents(parents: List) -> Bool:
    for p in parents:
        if !is_mapped(p):
             return False
     return True

valid_parents = not any([not is_mapped(p) for p in parents])

valid_parents = all([is_mapped(p) for p in parents])
```

matentzn: > for in order for the parent to be obsolete in Mondo, it must be mapped in Mondo

I agree with this! I only say that because the mappings from obsolete terms may not show up in your pipeline? If they do, then I agree with you!

Just think of it like this (integrity check of your approach):

**The migration (slurp) table may only ever be empty if and only if all terms from the source are either mapped, deprecated or excluded.** (note the absence of the `parent` condition, which is only relevant for the _order_ in which migration happens).

This should ideally not wait till Thursday (urgent requirement for the paper), just solve it somehow now and we discuss it later again. Also I am OOO next week, but will respond to slack.

src/scripts/migrate.py
matentzn, line 123: I would suggest to refactor everything in the outer for loop into a method like:

```
_is_migration_candidate()
```

This isolates the logic more clearly from the rest of the pipeline, and makes it easier to maintain.

Also, add the documentation right into the code as comments as well. I will get lost if its just in the PR.

docs/developer/workflows.md
joeflack4: Hmm, alright. I just added a parameter called `parent_conditions_on`, defaulting to False, and implemented a 1 line solution. If `parent_conditions_on == False`, then it skips all of that logic.

The end result is that the terms in `slurp/%.tsv` will be exactly the same as `reports/%_unmapped_terms.tsv`, which is the same as the list of terms in `reports/%_mapping_status.tsv` where `is_mapped`, `is_deprecated`, and `is_obsolete` are False. I just ran a test and confirmed that was the case.

src/scripts/migrate.py
joeflack4: `_is_migration_candidate()` makes sense, I will do that.

Regarding the documentation, I thought the same thing and did that already, if I understand what you mean. The same exact paragraph which describes what makes a term sluprable is in the `worklfows.md` as well as a code comment where the logic is. Although I will need to update it now that we have [simplified / removed the parent logic](https://github.com/monarch-initiative/mondo-ingest/pull/316#discussion_r1199661419).

src/scripts/migrate.py
```diff
@@ -124,7 +135,7 @@ def slurp(
             terms_to_slurp.append({
                 'mondo_id': mondo_id, 'mondo_label': mondo_label, 'xref': t.curie, 'xref_source': 'MONDO:equivalentTo',
                 'original_label': t.label if t.label else '', 'definition': t.definition if t.definition else '',
-                'parents': '|'.join(qualified_parent_mondo_ids)})
+                'parents': '|'.join([p for p in t.direct_owned_parent_curies])})
```
joeflack4, line 138: @matentzn We are now removing the need for checking anything regarding a term's parents to determine migration candidacy. As such I needed to update the `parents` column in `slurp/%.tsv`. Previously, this was only for `qualified_parent_mondo_ids`. Now it will show all parents.

src/scripts/migrate.py
```diff
@@ -91,30 +118,14 @@ def slurp(
...
-            # Conversely, if any of T.parents is unmapped, T is not to be slurped
-            if parent_curie in sssom_object_ids:
-                parent: Dict = sssom_df[sssom_df['object_id'] == parent_curie].to_dict('records')[0]
-                parent_mondo_id, parent_mondo_label = parent['subject_id'], parent['subject_label']
-                if not parent_mondo_label.startswith('obsolete'):
-                    qualified_parent_mondo_ids.append(parent_mondo_id)
-            if parent_curie not in sssom_object_ids and parent_curie in owned_term_curies:
-                break
-        if qualified_parent_mondo_ids or no_parents:
+        if not parent_conditions_on or _check_parent_conditions(t, sssom_object_ids, sssom_df):
```
joeflack4, line 128: @matentzn Implemented new default logic where we do not check anything about a term's parents when determining if it is migratable. Otherwise, there is a function called `_check_parent_conditions()` where it will include that logic if we pass the `--parent-conditions-on` param. 

I was actually going to name this `_is_migration_candidate()` as you proposed, but if you look a few lines above, you will see that we're already iterating over slurp candidates: `    for t in slurp_candidates:`. This is because earlier in the script we've already filtered out any terms that are already mapped or excluded.

src/scripts/migrate.py
joeflack4: @matentzn You can hit `Resolve conversation` on this one and continue the discussion [here](https://github.com/monarch-initiative/mondo-ingest/pull/316#discussion_r1199681163) where I've added the aforementioned function.

docs/developer/workflows.md
joeflack4: @matentzn When you're done looking at this you might want to hit "Resolve conversation" on this one and continue further discussion [here](https://github.com/monarch-initiative/mondo-ingest/pull/316#discussion_r1199681163).

src/scripts/migrate.py
```diff
@@ -52,10 +52,37 @@
...
+    no_parents: bool = not t.direct_owned_parent_curies
+    for parent_curie in t.direct_owned_parent_curies:
+        if parent_curie not in sssom_object_ids:
+            all_parents_mapped = False
+        if parent_curie in sssom_object_ids:
+            parent: Dict = sssom_df[sssom_df['object_id'] == parent_curie].to_dict('records')[0]
+            parent_mondo_id, parent_mondo_label = parent['subject_id'], parent['subject_label']
+            if parent_mondo_label.startswith('obsolete'):
+                obsolete_mondo_parent_ids.append(parent_mondo_id)
+            elif parent['predicate_id'] in ['skos:exactMatch', 'skos:narrowMatch']:
```
matentzn, line 75: I think this is mixing two concerns:

1. The condition under which a record should be declared mappable/added to the migration table (not obsolete, not excluded, all parents either mapped, obsoleted or excluded) and
2. What should be included in the "parents" column (of the parent is either an exact or narrow match). 

(2) is _not_ a further condition for adding the term to the slurp.

src/scripts/migrate.py
matentzn: I don't understand this at all! Sorry :/ its better to communicate with me in output tables then in code..

The parent column in the migration table (the one processed by Sabrina) can only contain exact or narrow-matched parents, so I am not sure if this is still true after this change.

src/scripts/migrate.py
matentzn: I think your suggestion of doing (b), the refactor, makes sense, but please do not drop the parent check, and also note my comment on the exactMatch/narrowMatch condition.

The logic here has become a bit too complex for nico-style review. Try to focus on understanding what the outcome of a feature like this one should be, and then lets communicate through bug reports rather than implementation details - I don't know whats best, and our dialogue is likely misleading you many times. 

Finish the PR in whatever way you see fit, and show me the migration tables created, and we discuss it from there!

src/scripts/migrate.py
joeflack4: Ash, OK. Your comment [here](https://github.com/monarch-initiative/mondo-ingest/pull/316#discussion_r1199793432) clarifies what you mean. Will continue discussion there.

src/scripts/migrate.py
joeflack4: I'm really glad you clarified that, thank you! Will do this now.

src/scripts/migrate.py
joeflack4: Hey Nico, yes, I agree, this is getting too confusing. I think that a meeting is the most efficient thing if there are a lot of complexities or ambiguities.

I only have one question left on something unclear but has only a minimal impact, if any, so I'll save it for our next meeting.

I'll do the following things now:
i. Implement my simplification refactor.
ii. Reactivate the parent conditions, but change the logic a little as you [just described](https://github.com/monarch-initiative/mondo-ingest/pull/316#discussion_r1199793432), _all parents should either be (a) mapped, (b) obsolete, or (c, new) excluded_
iii. Move the exact/narrow match conditions out of 'ii', valid parent conditions, and add it [instead](https://github.com/monarch-initiative/mondo-ingest/pull/316#discussion_r1199794773) to the list of parents that show up in the TSV

joeflack4: @matentzn I've just finished the logic updates and run it for 3 ontologies. It's still running so I'll commit the others when it's done.

Can you merge this after face checking the outputs if they're good?

Perhaps you're right about it not being a good use of your time to check on implementation, however. I have an idea.
Rather than waiting for bug reports, this is a good opportunity to implement some testing. Sabrina could probably come up with good cases, but you would likely be better at coming up with potential tricky edge cases. It would be nice if I could get a list of terms from you guys, and then for each of those, whether or not that term would be expected in the slurp output.

src/scripts/migrate.py
```diff
@@ -52,10 +50,20 @@
...
     'xref_source': '>A oboInOwl:source SPLIT=|', 'original_label': '', 'definition': 'A IAO:0000115', 'parents': 'SC %'}
 
+def _valid_parent_conditions(
+    parents: List[CURIE], mapped: Set[CURIE], excluded: Set[CURIE], obsolete: Set[CURIE]
+) -> bool:
+    """This is an optional, stricter check on slurp candidacy / order.
+
+    For a term to be immediately migratable, it must either (a) have no parents, or (b) all of its parents must be
+    mapped, obsolete, or excluded"""
+    return not parents or [any([x in y for y in [mapped, excluded, obsolete]]) for x in parents]
```
matentzn, line 60: I can only hope this expression is correct, as I am not really able to mentally parse it :D

matentzn (approved):

src/scripts/migrate.py
joeflack4: Lol, well it is supposed to do what is written in the comment above it. The `any()` part also translates to:

`[x if (x in mapped) or (x in excluded) or (x in obsolete) for x in parents]`

`mapped`, `excluded`, and `obsolete` are lists of terms.

src/ontology/slurp/omim.tsv
```diff
@@ -1,5 +1,5 @@
 mondo_id	mondo_label	xref	xref_source	original_label	definition	parents
 ID	LABEL	A oboInOwl:hasDbXref	>A oboInOwl:source SPLIT=|		A IAO:0000115	SC %
 MONDO:0859389	knobloch syndrome	OMIMPS:267750	MONDO:equivalentTo	Knobloch syndrome		
-MONDO:0859563	hyper-ige recurrent infection syndrome 1, autosomal dominant	OMIM:147060	MONDO:equivalentTo	hyper-ige recurrent infection syndrome 1, autosomal dominant		MONDO:0018037
+MONDO:0859563	hyper-ige recurrent infection syndrome 1, autosomal dominant	OMIM:147060	MONDO:equivalentTo	hyper-ige recurrent infection syndrome 1, autosomal dominant		OMIMPS:147060
```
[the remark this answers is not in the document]
joeflack4: Ah, with the parent logic swapping back and forth I accidentally made this bug. It looks like you merged it, but I will fix it with a new commit/PR shortly.

Issue #318, opened 2023-05-24
Grouping issue: Automated testing


joeflack4:
## Overview
I felt it would be a good idea to create a general issue around automated testing (unit tests, integration tests, etc). Right now we don't have any tests at all, I believe.

## Sub-task list
These are some of the workflows that I think could most use some tests.
- [ ] 1. migration pipeline
- [ ] 2. "unmapped" artefact pipelines
- [ ] 3. exclusions pipeline

### 1. migration pipeline
Can create some static input files where we can create some mock terms that meet various conditions, and we can assert what we expect after running the workflow (e.g. whether or not the term appears in `slurp/%.tsv`).
- excluded / not
- deprecated /not
- parents deprecated (0, 1, or all)
- parents not meet narrow/exact match conditions?

joeflack4:
@matentzn We discussed testing today at our meeting. Rather than making a new issue, I updated this existing one.

The reason it's called a "grouping issue" is because I don't expect it to be something where we specifically pick a time to tackle the whole issue at once.

joeflack4:
Btw, I ran `make test` today in `mondo-ingest` and the issue here is that NCIT fails because of memory.

<details><summary>Log</summary>
<p>

```
make test
mondo-ingest.Makefile:39: warning: overriding commands for target `components/omim.owl'
Makefile:447: warning: ignoring old commands for target `components/omim.owl'
mondo-ingest.Makefile:54: warning: overriding commands for target `components/ordo.owl'
Makefile:458: warning: ignoring old commands for target `components/ordo.owl'
mondo-ingest.Makefile:73: warning: overriding commands for target `components/ncit.owl'
Makefile:436: warning: ignoring old commands for target `components/ncit.owl'
mondo-ingest.Makefile:84: warning: overriding commands for target `components/doid.owl'
Makefile:392: warning: ignoring old commands for target `components/doid.owl'
mondo-ingest.Makefile:106: warning: overriding commands for target `component-download-icd10cm.owl'
Makefile:410: warning: ignoring old commands for target `component-download-icd10cm.owl'
mondo-ingest.Makefile:112: warning: overriding commands for target `components/icd10cm.owl'
Makefile:414: warning: ignoring old commands for target `components/icd10cm.owl'
mondo-ingest.Makefile:127: warning: overriding commands for target `components/icd10who.owl'
Makefile:425: warning: ignoring old commands for target `components/icd10who.owl'
[...]

PR #319, opened 2023-05-27
Slurp / migrate

joeflack4:
- Bugfix: Robot template 'parents' column IDs were not Mondo IDs


src/scripts/migrate.py
```diff
@@ -120,13 +123,15 @@ def slurp(
         else:
             next_mondo_id, mondo_term_ids = _get_next_available_mondo_id(next_mondo_id, max_id, mondo_term_ids)
             mondo_id = 'MONDO:' + str(next_mondo_id).zfill(7)  # leading 0-padding
+        qualified_parents = [p for p in t.direct_owned_parent_curies
+                             if p in match_types and match_types[p] in ['skos:exactMatch', 'skos:narrowMatch']]
+        qualified_mondo_paents = [mondo_id_map[p] for p in qualified_parents if p in mondo_id_map]
```
joeflack4, line 128: Fixed [new bug](https://github.com/monarch-initiative/mondo-ingest/pull/316#discussion_r1200443120) where parents in robot template were not Mondo IDs.

test/test_migrate.py
```diff
@@ -0,0 +1,60 @@
...
+from pathlib import Path
+
+import pandas as pd
+
+TEST_DIR = Path(os.path.abspath(os.path.dirname(__file__)))
+PROJECT_ROOT = TEST_DIR.parent
+sys.path.insert(0, str(PROJECT_ROOT))
+from src.scripts.migrate import slurp
+
+TEST_INPUT_DIR = TEST_DIR / 'input' / 'test_migrate'
```
joeflack4, line 14: ## Static files selected for testing
Test/Slurp: For `migrate.py` tests, I decided to make special static copies in `test/input/` of the following files:
```py
mondo_mappings_path=str(TEST_INPUT_DIR / 'mondo.sssom.tsv'),
mapping_status_path=str(TEST_INPUT_DIR / 'ordo_mapping_status.tsv'),
```

And not do so for the following files:
```py
ontology_path=str(ONTOLOGY_DIR / 'components' / 'ordo.owl'),
mondo_terms_path=str(ONTOLOGY_DIR / 'tmp' / 'mirror_signature-mondo.tsv'),
onto_config_path=str(ONTOLOGY_DIR / 'metadata' / 'ordo.yml'),
```

My rationale is that the first 2 can be customized so that even when our mapping status with ORDO changes, these static files can be set up to help us test basically every possible edge case we can imagine. 

The other ones haven't been changed because there's not really a need, and if anything goes wrong with those files either, the test should catch that by erroring.

## Possible future changes to static test files
Although, there is some rationale to maybe creating a shortened, static version of `ordo.owl` which will allow the script to run faster.

## Caveats / more possible future changes
I didn't make any dependencies for the `tests` goal, so if some of those files don't exist, it'll error out. Maybe this is a  design flaw. I up the `tests` goal to conveniently automatically find and run all Python unit tests in the project. 

Perhaps a different way would be to split things up, like a `test-slurp` that has goal dependencies, and then a `tests` goal with `test-slurp` as its dependency.

src/scripts/migrate.py
matentzn: can we seperate this bugfix into new PR and merge asap? The testing thing will take a bit longer..

src/ontology/mondo-ingest.Makefile
```diff
@@ -496,6 +496,13 @@ update-jinja-sparql-queries:
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/create_sparql__ordo_replace_annotation_based_mappings.py
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/create_sparql__ordo_mapping_annotations_violation.py
 
+#################
+##### Tests #####
+#################
+.PHONY: tests
+tests:
+	cd ../../; python -m unittest discover
```
matentzn, line 504: Rather than creating many such tests, a better general strategy could be to run the whole odk test pipeline with a subset of the configured ontologies. I have never tried this, maybe something like:

```
sh run.sh make build-mondo-ingest OTHER_SRC=components/ordo.owl components/doid.owl
```

No idea if that would work, and it will almost certainly require tweaking, but at least it would run the entire pipeline rather than cherry picking some parts.

src/scripts/migrate.py
joeflack4: Oh yeah of course. Good call.

src/ontology/mondo-ingest.Makefile
joeflack4: That's a good idea. I added this to:
- #318

I think doing testing on whole build a good idea but not a 100% substitute for what can be addressed by running smaller parts in isolation with some static inputs, which has benefits of:
1. Can run quickly whenever needed
2. Static inputs can be built specifically to detect bugs and handle edge, which otherwise might not come up when running the whole pipeline with dynamic inputs.

test/test_migrate.py
```diff
@@ -0,0 +1,87 @@
...
+            slurp_dir_path=str(TEST_OUTPUT_DIR),
+            outpath=outpath,
+            min_id=850000,
+            max_id=999999,
+            use_cache=USE_CACHE)
+        cls.df = pd.read_csv(outpath, sep='\t')
+        cls.migratable = cls.df['xref'].to_list()[1:]
+        cls.mapping_status_df = pd.read_csv(mapping_status_path, sep='\t')
+
+    def test_parents_field(self):
```
joeflack4, line 49: > Test parents field to make sure these are Mondo IDs and not native ontology IDs.

test/test_migrate.py
```diff
@@ -0,0 +1,87 @@
...
+        """Test parents field to make sure these are Mondo IDs and not native ontology IDs."""
+        example_parents_str: str = self.df['parents'][1]
+        parents: List[CURIE] = example_parents_str.split('|')
+        self.assertTrue(all([x.startswith('MONDO') for x in parents]))
+
+    def test_all_expected_in_output(self):
```
joeflack4, line 55: > Tests against known passing cases in input file. All of these terms meet all conditions for inclusion in the output file, e.g. they're not mapped, excluded, or obsolete, and they either have no parents, or all of their parents are either mapped, excluded, or obsolete.

test/test_migrate.py
```diff
@@ -0,0 +1,87 @@
...
+        """Tests against known passing cases in input file. All of these terms meet all conditions for inclusion in the
+        the output file, e.g. they're not mapped, excluded, or obsolete, and they either have no parents, or all of
+        their parents are either mapped, excluded, or obsolete."""
+        expected: List[CURIE] = \
+            self.mapping_status_df[self.mapping_status_df['expected_in_output'] == True]['subject_id'].to_list()
+        self.assertTrue(all([x in self.migratable for x in expected]))
+
+    def test_all_unexpected_not_in_output(self):
```
joeflack4, line 63: > Tests against known failing cases in input file. None of these terms meet all conditions for inclusion in the output file, e.g. they're either mapped, excluded, or obsolete, or not all of their parents are mapped, excluded, or obsolete.

test/test_migrate.py
```diff
@@ -0,0 +1,87 @@
...
+        """Tests against known failing cases in input file. None of these terms meet all conditions for inclusion in the
+        the output file, e.g. they're either mapped, excluded, or obsolete, or not all of their parents are mapped,
+        excluded, or obsolete."""
+        unexpected: List[CURIE] = \
+            self.mapping_status_df[self.mapping_status_df['expected_in_output'] == False]['subject_id'].to_list()
+        self.assertTrue(all([x not in self.migratable for x in unexpected]))
+
+    def test_deprecated_excluded_mapped_not_in_output(self):
```
joeflack4, line 71: > Tests subset cases of test_all_unexpected_not_in_output(), specifically one case of each where a term is either mapped, excluded, or obsolete/deprecated.

test/test_migrate.py
```diff
@@ -0,0 +1,87 @@
...
+        """Tests subset cases of test_all_unexpected_not_in_output(), specifically one case of each where a term is
+        either mapped, excluded, or obsolete/deprecated."""
+        for field in ['is_mapped', 'is_excluded', 'is_deprecated']:
+            df = self.mapping_status_df[self.mapping_status_df[field] == True]
+            unexpected: CURIE = df['subject_id'].to_list()[0]
+            self.assertNotIn(unexpected, self.migratable)
+
+    def test_no_parents_in_output(self):
```
joeflack4, line 79: > Tests subset case of test_all_expected_in_output(), specifically a case where the term is not mapped, excluded, or obsolete, and also that it has no parents.

joeflack4 (commented):
@matentzn Now ready for review.
@hrshdhgd @souzadevinicius Finally added some unit testing to `mondo-ingest`, starting with the migration pipeline. The tests are short and have some comments. If you glance through them you can probably grok what we're checking for when it comes to determining whether a term from an ontology is migratable or not into Mondo.

joeflack4 (commented):
@matentzn Now ready for review.
@hrshdhgd @souzadevinicius Finally added some unit testing to `mondo-ingest`, starting with the migration pipeline. The tests are short and have some comments. If you glance through them you can probably grok what we're checking for when it comes to determining whether a term from an ontology is migratable or not into Mondo.

test/test_migrate.py
```diff
@@ -0,0 +1,87 @@
...
+import os
+import sys
+import unittest
+from pathlib import Path
+from typing import List
+
+import pandas as pd
+from oaklib.types import CURIE
+
+TEST_DIR = Path(os.path.abspath(os.path.dirname(__file__)))
```
hrshdhgd, line 11: Same refactor suggestion as before.

test/test_migrate.py
```diff
@@ -0,0 +1,87 @@
...
+PROJECT_ROOT = TEST_DIR.parent
```
hrshdhgd, line 12: You are actually using this here! Awesome!

hrshdhgd (commented):
The tests in general look good to me. Nico's suggestion about adopting the odd test pipeline would be awesome and a learning experience for me through you @joeflack4 ! lol!

src/scripts/migrate.py
```diff
@@ -20,8 +22,11 @@
 from oaklib.implementations import ProntoImplementation
 from oaklib.types import CURIE, URI
 
-from utils import CACHE_DIR, DOCS_DIR, PREFIX, PROJECT_DIR, Term, _get_all_owned_terms, _get_next_available_mondo_id, \
-    get_mondo_term_ids, _load_ontology, SLURP_DIR
+SCRIPTS_DIR = Path(os.path.abspath(os.path.dirname(__file__)))
```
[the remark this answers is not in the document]
joeflack4: Thanks for the suggestion! I should have probably switched to `Path` long ago. I will likely make this change.

souzadevinicius:
Nice.I would be happy to contribute 😊

test/__init__.py
```diff
@@ -0,0 +1 @@
+"""Unit tests"""
```
joeflack4, line 1: ## Incorporate QC checks
The checks added in #385 should be incorporated here as tests as part of this PR ideally.

PR #325, opened 2023-06-05
ODK update, `%-db` removal

joeflack4:
- Update: ODK files after pulling latest image
- Delete: %-db goal as it is no longer necessary after latest update.


joeflack4 (commented):
@matentzn Note that we also moved the `unmapped-%.owl` files out of `components/` and into `unmapped/` for this very same issue. But I guess we can keep it that way? If you want to change it back I can.

PR #345, opened 2023-07-15
Bug fixes

hrshdhgd:
This PR has only 1 change (scripts/match-mondo-sources-all-lexical.py):
```
mapping_msdf.df[PREDICATE_MODIFIER] = "" if PREDICATE_MODIFIER not in mapping_msdf.df.columns else mapping_msdf.df[PREDICATE_MODIFIER]
```

Once we install the latest `sssom-py`, this line and the one above can be removed. But ran ` make matches -B` and it completed successfully.


src/scripts/match-mondo-sources-all-lexical.py
```diff
@@ -135,6 +135,7 @@ def run(input: str, config: str, rules: str, rejects: str, output: str):
         df=msdf.df, filter_prefixes=prefix_of_interest, features=[SUBJECT_ID, OBJECT_ID]
     )
     msdf.df[PREDICATE_MODIFIER] = "" if PREDICATE_MODIFIER not in msdf.df.columns else msdf.df[PREDICATE_MODIFIER]
+    mapping_msdf.df[PREDICATE_MODIFIER] = "" if PREDICATE_MODIFIER not in mapping_msdf.df.columns else mapping_msdf.df[PREDICATE_MODIFIER]
```
matentzn, line 138: I trust you on this, but it seems this should not be necessary. 

```
msdf.remove_mappings(mapping_msdf)
```

Should be resilient to data frames that don't have a `PREDICATE_MODIFIER` column!

matentzn (commented):
Can you make a new PR with only the relevant changes? There seem to be many changed files here, and I only think 3 or 4 need to be included, right?

hrshdhgd:
Continued by #347

PR #348, opened 2023-07-27
release 25.07.2023

matentzn:


src/ontology/lexmatch/mondo-only/unmapped_ncit_mondo.tsv
matentzn, line 1: @hrshdhgd I need a quick theory of what could cause such a huge drop of lines in this file..

src/ontology/lexmatch/split-mapping-set/mondo_exactmatch_omim.tsv
matentzn, line 1: Why this big drop ? @hrshdhgd

matentzn (commented):
As all the files I want investigated are non-essential, I will merge this.

src/ontology/lexmatch/mondo-only/unmapped_ncit_mondo.tsv
hrshdhgd: I looked at a handful and they all seemed to have been matched. e.g. 
 - `NCIT:C4587` => `exactMatch`
 - `NCIT:C2991` => `exactMatch`
 - `NCIT:C3611` => `exactMatch`
 - `NCIT:C6162` => `exactMatch`

All of the above can be found in `mondo_exactmatch_ncit.sssom.tsv`

matentzn: I saw you made a comment in my email, but I cant find that comment anymore. The key is that this is in the "mondo-only" folder - I don't really know what `mondo-only/unmapped*` means exactly. 

My question is a bit more: Why did this change happen _now_  - we did not align 1000 NCIT terms, not at least as far as I know (cc @sabrinatoro)

hrshdhgd: Initially I thought it was because the `unmapped` terms were mapped and hence they were removed from the `unmapped` file. But then I was looking at the file locally which wasn't up-to-date. So I deleted the comment.

>  I don't really know what mondo-only/unmapped* means exactly.

It means unmapped mappings with `MONDO:` as the subject. We had this decision made months ago.

matentzn: Are you certain? I seem to remember that `mondo-only` meant that the mapping was only in Mondo, but not in lexmatch?

src/ontology/lexmatch/README.md
```diff
@@ -3,48 +3,48 @@
 * mondo-only: Positive mappings in MONDO not caught by the lexical mapping pipeline
```
hrshdhgd, line 3: Here's the answer . You're correct.

matentzn: Ok, but what does mondo-only + unmapped mean? I mean, the word "unmapped"?

hrshdhgd: `unmapped` means mappings that are in one but not the other. For example: 

| subject_id   | subject_label              | object_id    | predicate_id          | object_label       | mapping_justification     | mapping_tool | confidence | subject_match_field | object_match_field | match_string |
|--------------|----------------------------|--------------|-----------------------|---------------------|---------------------------|--------------|------------|---------------------|--------------------|--------------|
| ID           | A oboInOwl:hasDbXref       | >A oboInOwl:source      |                       |                     |                           |              |            |                     |                    |              |
| MONDO:0000275 | obsolete monogenic disease | DOID:0050177 | MONDO:equivalentTo    | monogenic disease   | semapv:UnspecifiedMatching |              |            |                     |                    |              |

This mapping from `unmapped_doid_mondo.tsv` is in `mondo.sssom.tsv` but was not discovered in `lexmatch`. Hence `mondo-only`.

src/ontology/lexmatch/mondo-only/unmapped_ncit_mondo.tsv
hrshdhgd: So drop in `mondo-only/unmapped_ncit_mondo.tsv` means `lexmatch` had not caught this before but now it does.

src/ontology/lexmatch/split-mapping-set/mondo_exactmatch_ncit.tsv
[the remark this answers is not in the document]
hrshdhgd: `split-mapping-set` dir elements are the subset of `mondo-only` mappings `split` by `subject`,  `predicate` and `object` combination. So terms missing/dropped above will be the same here.

Issue #353, opened 2023-08-18
`lexmatch` is missing umapped_ordo tables


matentzn:
For some reason the ordo lexmatch table is missing.

https://github.com/monarch-initiative/mondo-ingest/blob/main/src/ontology/slurp/ordo.tsv#L60

- perrault syndrome type 1 should have been a lexmatch (same exact synonym)

@hrshdhgd high priority issue

Is it somehow possible to create an empty file if there are no matches, and not let it disappear?

I assume this is some kind of prefix mismatch again.

matentzn:
Please also check the following files:

src/ontology/reports/ordo_mapped_deprecated_terms.robot.template.tsv
src/ontology/reports/ordo_unmapped_terms.tsv
docs/reports/unmapped_ordo.md

hrshdhgd:
Could something from [this commit](https://github.com/monarch-initiative/mondo-ingest/commit/47acb634be73568c82708c54b9ff6baf0755f8d7) be responsible?

matentzn:
How? It does not affect Ordo/orphanet as far as I can see?

hrshdhgd:
Oops! I mistook `OMIMPS` for `ORDO`

matentzn:
Will you try to rerun and see if the problem lies with merged.db or mondo.sssom.tsv perhaps?

hrshdhgd:
It was correct until [release 25.07.2023](https://github.com/monarch-initiative/mondo-ingest/commit/f93dfc5c85da588dcdc0086168f1aef74377ab61) It changed in [release 2023-08-20](https://github.com/monarch-initiative/mondo-ingest/commit/a5bc5c5a75a8d59c6d8b075ecb523ff4258308f8)

matentzn:
Iiiiiiii!!!!! tHaaaanks! I will fix it.

matentzn:
Ok, now that I see this; why is this specific file you sent relevant (the ordo.sssom.tsv)? Lexmatch does not really care about that file? Is the prefix changes in mondo-ingest.db?

matentzn:
@hrshdhgd can you try and make sure that our internal prefixmap always takes precedence over the SSSOM py one?

hrshdhgd:
I think it already does that.

matentzn:
But why is the Ordo file empty then if we provide an override for the prefix?

hrshdhgd:
Ok, here's what I found:

The incorrect prefix occurs in 
[...]

PR #354, opened 2023-08-20
Mondo Ingest release 2023-08-20

matentzn:


matentzn (commented):
There is clearly something badly wrong with the ORDO, but no need to keep this from being merged. This issue needs solving #353

PR #356, opened 2023-08-22
Code to create merged_lex_exact_df

### LINKED ISSUE #355: Automatically update exact synonyms in Mondo
Create a GitHub Action in mondo (not mondo-ingest!) that runs once per week, and does the following:

0. (just thought of this, maybe easier) Provide a  merged version of lexmatch/*_exact.robot.tsv (lexmatch/all_exact.robot.tsv)
1. pull lexmatch/all_exact.robot.tsv from mondo-ingest repo using the `make TEMPLATE_URL... merge_template` workflow in `src/ontology/mondo.Makefile`. 
1. Make PR and assign to Sabrine

hrshdhgd:
Fixes #355


src/ontology/mondo-ingest.Makefile
```diff
@@ -444,8 +444,16 @@ lexmatch/README.md: $(SCRIPTSDIR)/lexmatch-sssom-compare.py $(MAPPINGSDIR)/mondo
 
 extract-unmapped-matches: lexmatch/README.md
 
+###################################
+#### Lexmatch-combine #######
+###################################
+lexmatch/all_exact.robot.tsv: $(SCRIPTSDIR)/lexmatch-sssom-compare.py
```
joeflack4, line 450: @hrshdhgd Personally I find Python way easier to do stuff in than ShellScript. But some months ago I was doing something similar for the exclusions pipeline, concatenating robot templates together but needing to leave out the 2 header lines. Stackoverflow or ChatGPT helped me create this `awk` command on 270 that worked for me.

https://github.com/monarch-initiative/mondo-ingest/blob/331e04a9ff6cba8aa7d41b439d011124ac54796d/src/ontology/mondo-ingest.Makefile#L268-L270

I don't know if this will work for you though because I see that you're also doing `merged_lex_exact_df.drop_duplicates(inplace=True)`. I don't know how I would do that in ShellScript.

src/scripts/lexmatch-sssom-compare.py
```diff
@@ -54,6 +60,28 @@ def main(verbose: int, quiet: bool):
...
+def combine_unmapped_lex_exacts() -> None:
+    file_list = [file for file in listdir(LEXMATCH_DIR) if file.endswith("_lex_exact.tsv")]
+    # Initialize an empty DataFrame to store the merged data
+    merged_lex_exact_df = pd.DataFrame()
+
+    # Concatenate all the files into a single DataFrame
+    for file in file_list:
+        file_path = join(LEXMATCH_DIR, file)
+        df = pd.read_csv(file_path, delimiter="\t")
+        merged_lex_exact_df = pd.concat([merged_lex_exact_df, df])
```
joeflack4, line 73: I don't know if it is any more performant (not that it really matters), but I tend to do `for file in file_list`, and then for each `file`, I read the `df` and append to a `df_list`. Then after the `for` loop, I do a `merged_df = pd.concat(df_list)`.

src/scripts/lexmatch-sssom-compare.py
```diff
@@ -54,6 +60,28 @@ def main(verbose: int, quiet: bool):
...
+
+    merged_lex_exact_df = pd.concat(
+        [pd.DataFrame.from_dict(ROBOT_ROW_DICT, orient="columns"), merged_lex_exact_df],
```
joeflack4, line 76: Perhaps this is ever so slightly simpler; here's an example of how I concat in the robot row:
```py
ROBOT_TEMPLATE_HEADER = {'mondo_id': 'ID', 'mondo_label': 'LABEL', 'xref': 'A oboInOwl:hasDbXref'}
final_df = pd.concat([pd.DataFrame([ROBOT_TEMPLATE_HEADER]), df])
```

joeflack4 (approved):

src/scripts/lexmatch-sssom-compare.py
matentzn: I dont think performance matters here much, but I agree @joeflack4 I think collecting in list and then concatenation all at once is cleaner. No need to block PR for this, just good to keep in mind.

matentzn (approved):
LGTM but no in depth review. I assume the code is tested.

src/ontology/mondo-ingest.Makefile
hrshdhgd: Yeah since the python script file was already in the project I overloaded it with another CLI command. Makes it easier to maintain and debug. But thanks for this, I always wondered what `awk` did.

PR #362, opened 2023-09-05
Subprocess output

joeflack4:
- Bugfix: On Python 3.7+, additional flags needed to be passed to subprocess.run() in order to capture output, which some of our code was depending on in order to check if any errors occurred, etc.

### DIFF
--- a/src/analysis/problematic_exclusions.py
+++ b/src/analysis/problematic_exclusions.py
@@ -133,7 +133,7 @@ def populate_labels(
             f.write(instantiated_str)
         with open(command_save_path, 'w') as f:
             f.write(command_str)
-        subprocess.run(command_str.split())
+        subprocess.run(command_str.split(), capture_output=True, text=True)
 
     # Read results and return
     try:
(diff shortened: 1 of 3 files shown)


src/analysis/problematic_exclusions.py
```diff
@@ -133,7 +133,7 @@ def populate_labels(
             f.write(instantiated_str)
         with open(command_save_path, 'w') as f:
             f.write(command_str)
-        subprocess.run(command_str.split())
+        subprocess.run(command_str.split(), capture_output=True, text=True)
```
joeflack4, line 136: Just 3 changes exactly like this

joeflack4, line 136: I didn't see this as an issue before. But this came up while I was doing an update in OAK, and I saw that I also had done this in `mondo-ingest`. This should be an issue for Python 3.7+.

@hrshdhgd I see you already have some code that passes these flags to capture output.

@matentzn I'm just fixing this / standardizing it so that all such calls are correctly capturing output.

hrshdhgd: I don't remember using these flags. Could you show me where? The only flags I've used for `subprocess` is `shell=True` which seems to solve most indues I've had in the past.

joeflack4: Hm, looks like it must have been me that used it previously. I've used `shell=True` before too, maybe with `Popen`. 

In any case, previously these flags were being passed only in some instances of `subprocess.run()`; now it's beeing used in all instances.

matentzn (approved):
I trust this does not break anything; lgtm

PR #367, opened 2023-09-29
Makefile package management updates

joeflack4:
Makefile package management
- Update: Moved goals 'sssom' and 'dependencies' into a new makefile section 'package management'.
- Update: Made 'dependencies' goal depend on specific dependencies. Refactored so there are less go
als.

### DIFF
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -14,6 +14,21 @@ MAPPINGSDIR=../mappings
 SKIP_HUGE=false
 
 ####################################
+### Package management #############
+####################################
+.PHONY: pip-%
+pip-%:
+	 python3 -m pip install --upgrade $*
+
+.PHONY: dependencies
+dependencies:
+	$(MAKE) pip-pip
+	$(MAKE) pip-setuptools
+	$(MAKE) pip-oaklib
+	$(MAKE) pip-sssom
+	$(MAKE) pip-semsql
+
+####################################
 ### Relevant signature #############
 ####################################
 # This section is concerned with identifiying the entities of interest that should be imported from the source.
@@ -159,14 +174,6 @@ ALL_COMPONENT_IDS=$(strip $(patsubst $(COMPONENTSDIR)/%.owl,%, $(OTHER_SRC)))
 #################
 ### Mappings ####
 #################
-.PHONY: sssom
-sssom:
-	python3 -m pip install --upgrade pip setuptools && python3 -m pip install --upgrade sssom
-
-.PHONY: dependencies
-dependencies:
-	python3 -m pip install --upgrade pip setuptools && python3 -m pip install --upgrade semsql==0.3.2 oaklib
-
 ALL_MAPPINGS=$(foreach n,$(ALL_COMPONENT_IDS), $(MAPPINGSDIR)/$(n).sssom.tsv)
 
 $(TMPDIR)/component-%.json: $(COMPONENTSDIR)/%.owl


src/ontology/mondo-ingest.Makefile
```diff
@@ -13,6 +13,28 @@
...
+.PHONY: sssom
+sssom: pip
+	python3 -m pip install --upgrade $@
+
+.PHONY: semsql
+semsql: pip
+	python3 -m pip install --upgrade semsql>=0.3.2
+
+.PHONY: dependencies
+dependencies: oaklib sssom semsql
```
joeflack4, line 36: Previously was only installing `semsql` and `oaklib`.
Currently no goal depends on this. I think it is just for ad hoc use.

src/ontology/mondo-ingest.Makefile
```diff
@@ -13,6 +13,28 @@
 MAPPINGSDIR=../mappings
 SKIP_HUGE=false
 
+####################################
+### Package management #############
+####################################
+.PHONY: pip
+pip:
```
joeflack4, line 20: To make things DRY I moved repetitive `python3 -m pip install --upgrade pip setuptools` out of each individual dependency goal, and made this a prerequisite to those goals.

Only disadvantage is that if you run `make dependencies`, it'll run this for each command dependency listed, but it shouldn't take up much time.

src/ontology/mondo-ingest.Makefile
```diff
@@ -13,6 +13,28 @@
...
+	python3 -m pip install --upgrade pip setuptools
+
+.PHONY: oaklib
+oaklib: pip
```
joeflack4, line 24: Added this, as I have need for latest `oaklib` this right now in some of my work.

src/ontology/mondo-ingest.Makefile
```diff
@@ -13,6 +13,28 @@
...
+	 python3 -m pip install --upgrade $@
+
+.PHONY: sssom
+sssom: pip
+	python3 -m pip install --upgrade $@
+
+.PHONY: semsql
+semsql: pip
+	python3 -m pip install --upgrade semsql>=0.3.2
```
joeflack4, line 33: @hrshdhgd Previously was `pip install --upgrade semsql==0.3.2`. Do we need to lock into that version specifically? Or can we just do `semsql` or `semsql>=0.3.2`?

joeflack4 (commented):
@matentzn @hrshdhgd Take a look. @matentzn Unless any large issues you foresee, do you mind if @hrshdhgd is the one to make the final call on this one and merge this when/if he sees fit?

src/ontology/mondo-ingest.Makefile
```diff
@@ -13,6 +13,28 @@
...
+####################################
+### Package management #############
+####################################
+.PHONY: pip
+pip:
+	python3 -m pip install --upgrade pip setuptools
+
+.PHONY: oaklib
+oaklib: pip
+	 python3 -m pip install --upgrade $@
```
joeflack4, line 25: Also changed these goals to use `$@` (\*shrug\*)

src/ontology/mondo-ingest.Makefile
matentzn: Exactly, you can use it like this:

```
sh run.sh make dependencies whateverothergoalyouneed
```

src/ontology/mondo-ingest.Makefile
matentzn: I dont remember, sorry.

src/ontology/mondo-ingest.Makefile
matentzn: This is fine. Note that an alternative pattern which I have only started to use recently is this (typed without checking for errors but you will get the picture). The reason for this is that it is never a good idea to have a .PHONY goal being a dependency, because it really defeats the purpose of `make`:

```
pip:
    python3 -m pip install --upgrade pip setuptools
 
sssom:
     pip install -U sssom

dependencies:
     $(MAKE) pip
     $(MAKE) sssom
````

File: src/ontology/mondo-ingest.Makefile (hunk 3 above)
Comment (matentzn) [reply src/ontology/mondo-ingest.Makefile:24]: ok

File: src/ontology/mondo-ingest.Makefile (hunk 5 above)
Comment (matentzn) [reply src/ontology/mondo-ingest.Makefile:25]: I dont think this is needed/useful for `.PHONY` goals.. This makes more sense when $@ corresponds to a filename.

File: src/ontology/mondo-ingest.Makefile (hunk 4 above)
Comment (hrshdhgd) [reply src/ontology/mondo-ingest.Makefile:33]: Latest should suffice.

Comment (joeflack4) [reply src/ontology/mondo-ingest.Makefile:33]: Makes sense to me. Changed!

File: src/ontology/mondo-ingest.Makefile (hunk 2 above)
Comment (joeflack4) [reply src/ontology/mondo-ingest.Makefile:20]: Sounds good. I changed the whole thing to:

```make
.PHONY:
pip:
	python3 -m pip install --upgrade pip setuptools

.PHONY: oaklib
oaklib:
	 python3 -m pip install --upgrade $@

.PHONY: sssom
sssom:
	python3 -m pip install --upgrade $@

.PHONY: semsql
semsql:
	python3 -m pip install --upgrade $@

.PHONY: dependencies
dependencies:
	$(MAKE) pip
	$(MAKE) oaklib
	$(MAKE) sssom
	$(MAKE) semsql
```

I kind of don't like the repetitiveness of `python3 -m pip install --upgrade $@` in multiple goals there, but I have tried things like creating functions in `make` before and it's just a whole realm of voodoo that I'm not good at / not worth the time for this.

matentzn: If you want to reduce the code you can refactor it like this:

```
.PHONY: pip-%
pip-%:

.PHONY: dependencies
dependencies:
	$(MAKE) pip-pip
	$(MAKE) pip-setuptools
	$(MAKE) pip-oaklib
	$(MAKE) pip-sssom
	$(MAKE) pip-semsql
```

joeflack4: Ash, so that's one way of functionalizing things.

I don't know if Harshad still has need for depending on just one of these packages, but the only one I currently need to depend on is `oaklib`. But it's probably not a harm to upgrade them all if I just want my "sync subclass" PR to depend on this `dependencies` goal instead.

PR #363, opened 2023-09-10
Synchronization: SubClassOf

joeflack4:
## Updates
Synchronization: SubClassOf
Pipeline to keep subclass relationships in sync between Mondo and sources.
Currently handled: Case 1, 3, 5
- Add: makefile goals: sync, sync-subclassof, and more.
- Update: makefile goal: build-mondo-ingest
- Add: src/scripts/sync_subclassof.py
- Add: Outputs to reports/

General
- Bugfix: utils.py remove_angle_brackets(): Now correctly returns a str if receives a str.
- Update: config/prefixes.csv: Add: New entries
- Add: utils.py: get_monarch_curies_converter()
- Update: Python requirements: oaklib upgrade
- Update: Minor refactor in _get_all_owned_terms()
- Update: .gitignore: entries, ordering, and comments

Related to other PRs
- Update: metadata/icd10who.yml: The 2 prefix maps in this file were inconsistent with each other. Corrected here and also in https://github.com/monarch-initiative/mondo-ingest/pull/377

PR #374, opened 2023-11-20
SemanticSQL `.db` goal updates

joeflack4:
## Updates
    Artefact & makefile updates
    - Rename: mondo-ingest.db --> merged.db: It was coming from merged.owl, and this was causing a conflict, since mondo-ingest.db (from mondo-ingest.owl) is needed separately.

    Makefile updates
    - Update: %.db: Replaces all other *.db goals, unless they have a specific goal that overrides.
    - Add: tmp/mondo-ingest.db

## Context
- https://github.com/monarch-initiative/mondo-ingest/pull/363#discussion_r1396484410

### DIFF
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -29,6 +29,18 @@ dependencies:
 	$(MAKE) pip-semsql
 
 ####################################
+### General ########################
+####################################
+%.db: %.owl
+	@rm -f .template.db
+	@rm -f .template.db.tmp
+	@rm -f tmp/merged-relation-graph.tsv.gz
+	RUST_BACKTRACE=full semsql make tmp/$*.db -P config/prefixes.csv
+	@rm -f .template.db
+	@rm -f .template.db.tmp
+	@rm -f tmp/merged-relation-graph.tsv.gz
+
+####################################
 ### Relevant signature #############
 ####################################
 # This section is concerned with identifiying the entities of interest that should be imported from the source.
@@ -326,7 +338,7 @@ documentation: j2 $(ALL_DOCS) unmapped-terms-docs mapped-deprecated-terms-docs s
 
 .PHONY: build-mondo-ingest
 build-mondo-ingest:
-	$(MAKE) refresh-imports exclusions-all mondo-ingest.db slurp-all mappings matches \
+	$(MAKE) refresh-imports exclusions-all slurp-all mappings matches \
 		mapped-deprecated-terms mapping-progress-report \
 		recreate-unmapped-components documentation
 	$(MAKE) prepare_release
@@ -412,19 +424,6 @@ $(MAPPINGSDIR)/rejected-mappings-sssom.tsv: $(MAPPINGSDIR)/rejected-mappings.tsv
 tmp/merged.owl: tmp/mondo.owl mondo-ingest.owl tmp/mondo.sssom.ttl
 	$(ROBOT) merge -i tmp/mondo.owl -i mondo-ingest.owl -i tmp/mondo.sssom.ttl --add-prefixes config/context.json \
 			 remove --term "http://purl.obolibrary.org/obo/mondo#ABBREVIATION" --preserve-structure false -o $@
-
-tmp/merged.db: tmp/merged.owl
-	@rm -f .template.db
-	@rm -f .template.db.tmp
-	@rm -f tmp/merged-relation-graph.tsv.gz
-	RUST_BACKTRACE=full semsql make $@ -P config/prefixes.csv
-	@rm -f .template.db
-	@rm -f .template.db.tmp
-	@rm -f tmp/merged-relation-graph.tsv.gz
-
-mondo-ingest.db: tmp/merged.db
-	cp $< $@
-.PRECIOUS: mondo-ingest.db
 
 $(MAPPINGSDIR)/mondo-sources-all-lexical.sssom.tsv: $(SCRIPTSDIR)/match-mondo-sources-all-lexical.py tmp/merged.db $(MAPPINGSDIR)/rejected-mappings.tsv
 	rm -f $(MAPPINGSDIR)/mondo-sources-all-lexical.sssom.tsv
--- a/.gitignore
+++ b/.gitignore
@@ -61,7 +61,6 @@ src/ontology/imports/*.owl
 !src/ontology/imports/ro_import.owl
 src/ontology/imports/*.json
 src/ontology/imports/*_terms_combined.txt
-src/ontology/mondo-ingest.db
 
 src/ontology/components/*
 


src/ontology/mondo-ingest.Makefile
```diff
@@ -413,18 +413,17 @@ tmp/merged.owl: tmp/mondo.owl mondo-ingest.owl tmp/mondo.sssom.ttl
...
 	@rm -f .template.db.tmp
 	@rm -f tmp/merged-relation-graph.tsv.gz
-	RUST_BACKTRACE=full semsql make $@ -P config/prefixes.csv
+	RUST_BACKTRACE=full semsql make tmp/$*.db -P config/prefixes.csv
 	@rm -f .template.db
 	@rm -f .template.db.tmp
 	@rm -f tmp/merged-relation-graph.tsv.gz
 
-mondo-ingest.db: tmp/merged.db
+merged.db: tmp/merged.db
```
joeflack4, line 425: ## Rename: `mondo-ingest.db` --> `merged.db`
Justification: (i) `merged.db` is a literal copy of `tmp/merged.db`, which comes from `merged.owl`, (ii) Synchronization pipeline has need of `mondo-ingest.db`, which comes from `mondo-ingest.owl`.

Personally I don't even thing we need `merged.db`. We have goals which depend on `tmp/merged.db` (which is exactly the same) but no goals that depend on `merged.db`.

src/ontology/mondo-ingest.Makefile
```diff
@@ -413,18 +413,17 @@ tmp/merged.owl: tmp/mondo.owl mondo-ingest.owl tmp/mondo.sssom.ttl
...
 	cp $< $@
-.PRECIOUS: mondo-ingest.db
```
joeflack4, line 427: ## Removed `.PRECIOUS` on `merged.db`
I read up again on `.PRECIOUS` and intermediate files, and how they are deleted after a goal is run (conversing with ChatGPT helped a lot). However, `merged.db` should not be considered an intermediate file because it has a goal. Intermediate files are usually/always referenced in the prerequisites as a variable expansion ([example](https://stackoverflow.com/questions/15189704/makefile-removes-object-files-for-no-reason)), not as something that has a goal of its own. If it has a goal of its own, it shouldn't get deleted after running.

hrshdhgd (approved):

src/ontology/mondo-ingest.Makefile
```diff
@@ -413,18 +413,20 @@ tmp/merged.owl: tmp/mondo.owl mondo-ingest.owl tmp/mondo.sssom.ttl
...
+
+tmp/mondo-ingest.db: mondo-ingest.db
```
joeflack4, line 428: ## Add: `tmp/mondo-ingest.db`
This variation of the goal actually comes from `mondo-ingest.owl` and not from `tmp/merged.owl`. Putting it in  `src/ontology/tmp/` instead of `src/ontology/` (which is where `mondo-ingest.owl` is) is to be standard with where the other `.db` files are. It has no reason to be in `src/ontology/`. I'm not sure why `mondo-ingest.owl` is going there TBH.

src/ontology/mondo-ingest.Makefile
```diff
@@ -413,18 +413,20 @@ tmp/merged.owl: tmp/mondo.owl mondo-ingest.owl tmp/mondo.sssom.ttl
 	$(ROBOT) merge -i tmp/mondo.owl -i mondo-ingest.owl -i tmp/mondo.sssom.ttl --add-prefixes config/context.json \
 			 remove --term "http://purl.obolibrary.org/obo/mondo#ABBREVIATION" --preserve-structure false -o $@
 
-tmp/merged.db: tmp/merged.owl
+%.db: %.owl
```
joeflack4, line 416: ## Add: `%.db`
We previously had a lot of duplicate code (literally exact same goal recipe/commands) for the creation of these `.db` files. It seems overdue that this be captured with a wildcard.

This will also properly handle `tmp/*.db` cases, not just `*.db`.

src/ontology/mondo-ingest.Makefile
```diff
@@ -326,7 +326,7 @@ documentation: j2 $(ALL_DOCS) unmapped-terms-docs mapped-deprecated-terms-docs s
 
 .PHONY: build-mondo-ingest
 build-mondo-ingest:
-	$(MAKE) refresh-imports exclusions-all mondo-ingest.db slurp-all mappings matches \
+	$(MAKE) refresh-imports exclusions-all merged.db slurp-all mappings matches \
```
joeflack4, line 329: ## `merged.db` within `build-mondo-ingest`
@hrshdhgd Nico doesn't know; do you know why `merged.db` (previously called `mondo-ingest.db` even though it comes from `tmp/merged.owl`) is part of `build-mondo-ingest`?

@matentzn  @hrshdhgd I'm guessing maybe this was added because we just want to mark it as some required output / something that is needed for release / analysis. But I personally think this should be removed from `build-mondo-ingest`; it stands out as the only artefact that is declared in that goal, and I just don't see any reason for it.

hrshdhgd: I think we had that there during a meeting when Nico was dictating recipes to me. I don't recall me being a decision maker for this one.

src/ontology/mondo-ingest.Makefile
```diff
@@ -28,6 +28,21 @@ dependencies:
 	$(MAKE) pip-sssom
 	$(MAKE) pip-semsql
 
+####################################
+### General ########################
```
joeflack4, line 32: ## `%.db` to new "General" section
The previous `*.db` goals were in the "lexmatch" section even though not all of them were lexmatch related. I kept the `merged.db` goal there since lexmatch is all that depends on it, but I moved `%.db` here to this new general section, as there didn't look like there was anywhere else appropriate to put it.

src/ontology/mondo-ingest.Makefile
```diff
@@ -28,6 +28,21 @@ dependencies:
...
+%.db: %.owl
+	@rm -f .template.db
+	@rm -f .template.db.tmp
+	@rm -f tmp/merged-relation-graph.tsv.gz
+	RUST_BACKTRACE=full semsql make tmp/$*.db -P config/prefixes.csv
+	@rm -f .template.db
+	@rm -f .template.db.tmp
+	@rm -f tmp/merged-relation-graph.tsv.gz
+
+tmp/mondo-ingest.db: mondo-ingest.db
```
joeflack4, line 43: ## Add: `tmp/mondo-ingest.db`
This is needed by the synchronization pipeline, which hasn't been merged yet. But since we're sorting out other `.db` goal idiosyncrasies in this PR, I think it'd good to do here and I can rebase into my PR after.

The reason this one needs a specific goal and can't be handled by `%.db` is because the `%.db` goal (and SemanticSQL by design) expect that the directory of the output `.db` and its input be in the same directory. But that's not the case here. I think `mondo-ingest.db` should go into `tmp/` with the other `.db` files, but its source `mondo-ingest.owl` is not in `tmp/` for some reason.

How it works: So this goal basically just moves `mondo-ingest.db` to `tmp/mondo-ingest.db` after `mondo-ingest.db` is created through the `%.db` wildcard goal.

src/ontology/mondo-ingest.Makefile
matentzn: If merged.db is not needed anywhere can we renamed it mondo-ingest-merged.db, because I think this file is a file that Chris was asking for in the past (mondo+all ingests) merged together.

src/ontology/mondo-ingest.Makefile
matentzn: If you are positive that there are no _hidden_ dependencies on this file (thinking, python scripts, use grep to find out), then I am fine removing it for now.

joeflack4: Sounds good. Based on what you and Harshad are saying, and my investigations, and us not finding anything about it in the release during our meeting, it looks like it was put there by mistake, so I will delete it.

PR #376, opened 2023-11-21
Add MedGen Prefixes

joeflack4:
## Updates
- Add: MedGen prefixes to prefixes.csv
- Add: config/medgen.yml

## Sub-tasks 
- [x] Add: MedGen prefixes to `prefixes.csv`
- [x] Add: `config/medgen.yml`
- [ ] Add: MedGen prefixes to `mondo.sssom.config.yml` `curie_map` and `extended_prefix_map`
- [ ] Add anywhere else that may be missing as outlined in `docs/developer/add-new-source.md`
- [ ] Make sure nothing was added that shouldn't be, in compliance with #378

## Related
- #375
- #377 
- #146 
- https://github.com/monarch-initiative/mondo-ingest/pull/363#discussion_r1396540100

## Addresses
- #414

PR #375, opened 2023-11-21
Update `prefixes.csv`

joeflack4:
## Updates
- Update: prefixes.csv: Missing prefixes

## Related stuff
- #376 
- #377 
- #146 
- https://github.com/monarch-initiative/mondo-ingest/pull/363#discussion_r1396540100

Issue #146, opened 2023-01-31
Automating `prefix_map`s into (a) `metadata/ONTO.yml`, (b) a single registry


joeflack4:
## Overview
Original conversation: https://github.com/monarch-initiative/mondo-ingest/pull/127#discussion_r1055902211

Summary
mondo-ingest prefix_maps for native & foreign URIs *
- current situation
  - in mondo-ingest we have 
  - 'base_prefix_map' for native
  - 'prefix_map' for foreign/everything
- Nico suggestion
  - rather than have a 'prefix_map' for every ontology, 
  - we should have a master 'prefix_map' that we can use for all ontologies
- Joe: Potential issues w/ suggestion & possible solution
  - Potential issues
    - That sounds like a good idea. Although, what if two different ontologies (let's just say ORDO and DO) are using 2 different versions / URIs for their prefixes for a given external ontology (let's just say RO for sake of example).
  - Possible solution
    - A way to fix this would be, for `prefix_map` to not be of type `Dict[str, str]`, but of type `Dict[str, List[str]` to accommodate for various different versions / URIs for each of these prefixes.

matentzn:
I think prefix maps should usually be bi-directional; but you are right, we need to reconcile prefixes in mondo ingest.. Tough nut. Let me know what solution you are thinking!

joeflack4:
I chatted with [Chris on Slack](https://obo-communitygroup.slack.com/archives/C03D93DEALA/p1675205568176069) about this, and I think I'm going to defer to his knowledge/advice here, and probably you will agree the same:

1. For `base_prefix_map` (the ones that are 'owned' by the ontology, e.g. `OMIM:` and `OMIMPS:` for OMIM), we should use what's in the ontology itself, and not automate that; just set it manually.
2. For everything else, we can share a single `prefix_map`. Options (a) I haven't checked out https://github.com/linkml/prefixmaps/ yet, but we may just be able to use this. I assume it queries Bioregistry. If 'a' doesn't meet our needs, than (b) we could somehow get from Bioregistry directly via a `make` goal and cache it in `metadata/` and re-use it.

matentzn:
> For base_prefix_map (the ones that are 'owned' by the ontology, e.g. OMIM: and OMIMPS: for OMIM), we should use what's in the ontology itself, and not automate that; just set it manually.

Yes, agreed!

[...]

PR #377, opened 2023-11-21
ICD10WHO prefix updates

joeflack4:
## Updates
- Update: prefixes.csv: Add ICD10WHO prefix map
- Update: icd10who.yml: Fixed for consistency

## Related stuff
- #375
- #376 
- #146 
- https://github.com/monarch-initiative/mondo-ingest/pull/363#discussion_r1396540100


joeflack4 (commented):
@matentzn I made this its own PR because I feel this one is more important than the other prefix PRs. If this is not addressed there could be bugs.

Context/related: 
- #375 
- #376 
- #146 
- https://github.com/monarch-initiative/mondo-ingest/pull/363#discussion_r1396540100

matentzn (approved):
Thanks! We should eventually use a better "PURL" for ICD10WHO, but for now, this is ok

PR #380, opened 2023-11-30
Python requirements for Apple Silicon

joeflack4:
## Updates
- Add: Python requirements for Apple Silicon


python-requirements-apple-silicon.txt
```diff
@@ -0,0 +1,123 @@
+airium==0.2.6
```
joeflack4, line 1: ## Alternate file for Apple Silicon
This issue explains the problem and the solution implemented:
- #381

hrshdhgd:
If my opinion matters here, I strongly recommend to convert to poetry without using yet another requirements.txt file. Poetry is architecture agnostic as far as I know.

PR #382, opened 2023-11-30
Bugfix: `%.db`

joeflack4:
## Updates
- Update: %.db goal: Was always creating within (potentially nested) …


src/ontology/mondo-ingest.Makefile
```diff
@@ -35,7 +35,7 @@ dependencies:
 	@rm -f .template.db
 	@rm -f .template.db.tmp
 	@rm -f tmp/merged-relation-graph.tsv.gz
-	RUST_BACKTRACE=full semsql make tmp/$*.db -P config/prefixes.csv
```
joeflack4, line 38: The bugfix

PR #383, opened 2023-11-30
RO/OMO import & `.gitignore` updates

joeflack4:
## Updates
- Update: imports: RO & OMO
- Update: .gitignore: (i) Added relation-graph file pattern, (ii) reorganized and added comments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,18 +1,25 @@
+# General
+# - directories
+__pycache__/
 /.idea/
 /.run/
 bin/
 cache/
 semantic.cache/
-
+# - patterns
 *.code-workspace
 *.tmp
 *.tmp.obo
 *.tmp.owl
 *.tmp.json
 *.db
+.vscode/*
 
+# - static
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -1,18 +1,25 @@
...
 *.tmp.owl
 *.tmp.json
 *.db
+.vscode/*
 
+# - static
 .DS_Store
 
+# Specific files, patterns, and directories
+*relation-graph*
```
joeflack4, line 22: ## `relation-graph` ignore
After running to create SemanticSQL DBs, I was surprised to see it wanted to commit one of the relation-graph intermediaries (which ideally would be deleted automatically). So I added it to the `.gitignore`.

.gitignore
```diff
@@ -1,18 +1,25 @@
+# General
```
joeflack4, line 1: ## `.gitignore` organization
I took this time to organize a little more.

PR #385, opened 2023-12-08
QC checks

joeflack4:


joeflack4:
@hrshdhgd @souzadevinicius We want to actually use a proper testing framework and setup as in #319, but we want to discuss setting that up in a standard, approved way before merging it, so we're getting the ball rolling now with some QC that involve simply running scripts.

tests/__init__.py
```diff
@@ -0,0 +1 @@
+"""Tests"""
```
joeflack4, line 1: ## Little bit of general setup required
There are 2 checks that came out of my sync subclass PR that I wanted to preserve. As such, this entails ideally multiple files, which further entails some setup. An `__init__.py` and a `config.py` for allowing imports for DRYness, etc.

tests/check_dupe_exact_mappings.py
```diff
@@ -0,0 +1,81 @@
+"""Report duplciate exact mappings in mondo.sssom.tsv"""
```
joeflack4, line 1: ## Check for duplicate exact mappings from multiple Mondo IDs to single source ID

tests/check_sssom_in_sync.py
```diff
@@ -0,0 +1,90 @@
+"""Checks that SSSOM is in sync with Mondo by ensuring that all SSSOM doesn't have any novel IDs."""
```
joeflack4, line 1: ## Check for any source IDs in mappings that aren't in the ontology

src/ontology/mondo-ingest.Makefile
```diff
@@ -519,6 +519,24 @@ $(REPORTDIR)/%_mapped_deprecated_terms.robot.template.tsv: $(REPORTDIR)/%_mappin
 	--mapping-status-path $(REPORTDIR)/$*_mapping_status.tsv \
 	--outpath $@
 
+####################################
+### QC / test ######################
```
joeflack4, line 523: ## QC section in makefile
This setup will be changed when #319 is decided on and implemented, and these checks will go into that setup.

joeflack4 (commented):
@matentzn Mergeable unless you find any mistakes or have any suggestions.

matentzn (commented):
We need to discuss this in a call, as there are two architectural questions attached:

1. All mondo related python code should be in mondolib so it gets the same code quality practices associated and shared code can be levaraged
1. IMO the check you have here is Mondo-level, not Mondo-ingest level, so there is a question of where it should be deployed.

Issue #386, opened 2023-12-11
Mondo ingest SSSOM 0.4 migration


matentzn:
```
make ../mappings/doid.sssom.tsv
```

(and probably by extension all `../mappings/%.sssom.tsv`) is now broken because conflicting prefixes due to the new sssom py version:

```
http://identifiers.org/snomedct/:
        prefix='SNOMEDCT_US_2022_10_31' uri_prefix='http://identifiers.org/snomedct/' prefix_synonyms=[] uri_prefix_synonyms=[] pattern=None
        prefix='SNOMEDCT_US_2023_02_28' uri_prefix='http://identifiers.org/snomedct/' prefix_synonyms=[] uri_prefix_synonyms=[] pattern=None
```

The proper solution here is most likely to use EPMs instead of prefixmaps, and declare the `SNOMEDCT_US_2022_10_31` prefixes etc as synonym prefixes in the EPM. 

Check also if these are not in sssom.context.get_converter() already, then we can remove them from the prefixmap managed in mondo-ingets.

matentzn:
This has been fixed for a while!

PR #389, opened 2023-12-16
Ran `make mappings -B`

### LINKED ISSUE #372: The ORDO ingest is still broken
@hrshdhgd 

I dont know what needs doing here, but this needs to go to the top of the pile.

Right now, _all_ of ORDO appears as unmapped:

https://github.com/monarch-initiative/mondo-ingest/blob/main/src/ontology/lexmatch/unmapped_ordo_lex_exact.tsv

This is most likely because still, the mondo.sssom.tsv uses the wrong ordo prefix. Can you hack the mondo pipeline in a away that we get the right prefix? You said it worked for you locally, but you may have used local versions of sssom-py that were never published. The problem happens here (most likely):

https://github.com/monarch-initiative/mondo/blob/master/src/ontology/mondo.Makefile#L575

and you need to find a way with the version of sssom-py in ODK that it uses the correct ORDO prefix.

hrshdhgd:
Fixes #372 

Continued from #386


matentzn:
Hey @hrshdhgd dont worry about this for now; I will take this part over until I hit a wall, I am running this `mappings` part in another context.

matentzn:
Superceded by #390

PR #391, opened 2024-01-07
Add dependency: bioregistry

joeflack4:


joeflack4:
@twhetzel @matentzn Perhaps this is wrong to merge without review but it is a small update. We're already using `bioregistry`. It's being used by one of Harshad's scripts. And it also resides in the lockfile `python-requirements.txt`. It just wasn't in the unlocked file nor the apple silicon file.

Also in any event these files will be migrated to mondolib sooner or later and we'll be using poetry instead of this dependency management system.

Issue #390, opened 2023-12-17
Bug in match-mondo-sources-all-lexical.py


matentzn:
```
Traceback (most recent call last):
  File "/work/src/ontology/../scripts/match-mondo-sources-all-lexical.py", line 174, in <module>
    main()
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/work/src/ontology/../scripts/match-mondo-sources-all-lexical.py", line 121, in run
    msdf = lexical_index_to_sssom(oi, lexical_index, ruleset=ruleset, meta=meta)
  File "/usr/local/lib/python3.10/dist-packages/oaklib/utilities/lexical/lexical_indexer.py", line 316, in lexical_index_to_sssom
    Converter.from_prefix_map((meta or {}).pop(CURIE_MAP, {})),
AttributeError: 'tuple' object has no attribute 'pop'
make[1]: *** [mondo-ingest.Makefile:432: ../mappings/mondo-sources-all-lexical.sssom.tsv] Error 1
```

matentzn:
After updating all dependencies the issue persists:

```
Traceback (most recent call last):
  File "/work/src/ontology/../scripts/match-mondo-sources-all-lexical.py", line 174, in <module>
    main()
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1434, in invoke
[...]

PR #394, opened 2024-01-12
Bug in `match-mondo-sources-all-lexical.py`

joeflack4:
## Updates
Bugfixes: match-mondo-sources-all-lexical.py
- Bugfix: AttributeError: 'tuple' object has no attribute 'pop': wrong datatype for metadata was being passed to lexical_index_to_sssom()
- Bugfix: Several other bugs in mondo-ingest, and upgrading OAK/sssom-py/curies to fix other bugs related to prefix maps.
- Update: mondo.sssom.config.yml: Commented out duplicate prefix 'oio'
- Update: prefixes.csv: Removed duplicate prefix oio
- Update: Python requirements: Upgraded curies for bugfix involving get_prefixes(include_synonyms)
- Update: Not bugfixes, but did some codestyle fixes: (i) converted regular comments to docstring, (ii) added missing docs
tring, (iii) renamed built-in 'input' param

## Sub-tasks
- [x] 1. Determine if OAK changes needed or not & PR if so _(not needed)_
- [x] 2. Fulfill all requests / tasks in `sssom-py` PR: https://github.com/mapping-commons/sssom-py/pull/485
  - [x] Remove passing of non-bijective map into `sssom-py` if possible
  - [x] Upgrade `sssom-py` and `curies` and use those recent changes instead of my PR changes, if needed _(wasn't necessary for this PR but upgraded `curies` anyway. may need in near future)_
- [x] 3. If root problem needs to be addressed (`oio` and `oboInOwl` both in Mondo), make PR in `mondo` repo too _(doesn't seems so. seems to be related to `oio` in `config/prefixes.csv`)_
- [x] 4. Fix original bug in issue: `AttributeError: 'tuple' object has no attribute 'pop'`
- [x] 5. Add EPM to `mondo-ingest`
- [x] 6. Remove `oio` from `mondo.sssom.config.yaml` `curie_map` and `prefixes.csv`

## Related
- https://github.com/mapping-commons/sssom-py/pull/485
  - Has necessary bugfix to `clean_prefix_map()` (`.get_prefix(include_synonyms=True)`)
- https://github.com/INCATools/ontology-access-kit/pull/697
  - This one is not necessary to merge in order to fix anything but was inspired by what I was working on here.
- https://github.com/INCATools/ontology-access-kit/issues/698
  - We can probably avoid addressing this issue by changing our `oio` to `oboInOwl` in part of our pipeline.

---

@souzadevinicius FYI. Harshad is out ATM so I'm helping out with a bug that appeared here, but it opened a can of worms, involving conflicting prefixes that are causing trouble when `mondo-ingest` interacts with SSSOM and OAK. And we've been decided how to fix this, involving using extended prefix maps instead of plain, flat, bijective prefix maps (key: val), etc.

### DIFF
--- a/src/ontology/config/prefixes.csv
+++ b/src/ontology/config/prefixes.csv
@@ -3,7 +3,7 @@ rdf,http://www.w3.org/1999/02/22-rdf-syntax-ns#
 rdfs,http://www.w3.org/2000/01/rdf-schema#
 xsd,http://www.w3.org/2001/XMLSchema#
 owl,http://www.w3.org/2002/07/owl#
-oio,http://www.geneontology.org/formats/oboInOwl#
+oboInOwl,http://www.geneontology.org/formats/oboInOwl#
 dce,http://purl.org/dc/elements/1.1/
 dct,http://purl.org/dc/terms/
 foaf,http://xmlns.com/foaf/0.1/
--- a/src/ontology/metadata/mondo.sssom.config.yml
+++ b/src/ontology/metadata/mondo.sssom.config.yml
@@ -34,7 +34,7 @@ curie_map:
   semapv: https://w3id.org/semapv/vocab/
   rdf: http://www.w3.org/1999/02/22-rdf-syntax-ns#
   sssom: https://w3id.org/sssom/
-  oio: http://www.geneontology.org/formats/oboInOwl#
+#  oio: http://www.geneontology.org/formats/oboInOwl#
   GTR: "http://purl.obolibrary.org/obo/mondo/mappings/unknown_prefix/GTR/"
   NCI: "http://purl.obolibrary.org/obo/mondo/mappings/unknown_prefix/NCI/"
   NIFSTD: "http://purl.obolibrary.org/obo/mondo/mappings/unknown_prefix/NIFSTD/"
--- a/src/scripts/match-mondo-sources-all-lexical.py
+++ b/src/scripts/match-mondo-sources-all-lexical.py
@@ -1,17 +1,19 @@
-# Basic matching pipeline that takes in
+"""Match mondo sources, all lexical
+Basic matching pipeline that takes in
 
-# Input:
-# 1. MERGED_ONTOLOGY = tmp/merged.owl
-# 2. SSSOM_CONFIG = metadata/mondo.sssom.config.yml
-# 3. OUTPUT_SSSOM = mapping/mondo-sources-all-lexical.sssom.tsv
+Input:
+1. MERGED_ONTOLOGY = tmp/merged.owl
+2. SSSOM_CONFIG = metadata/mondo.sssom.config.yml
+3. OUTPUT_SSSOM = mapping/mondo-sources-all-lexical.sssom.tsv
 
-# I would try some basic things first:
+I would try some basic things first:
 
-# Use synonymiser
-# Use oak.mapping() pipeline
-
+Use synonymiser
+Use oak.mapping() pipeline
+"""
 import logging
 from pathlib import Path
+from curies import Converter
 from oaklib.resource import OntologyResource
 from oaklib.implementations.sqldb.sql_implementation import SqlImplementation
 from oaklib.utilities.lexical.lexical_indexer import (
@@ -25,11 +27,11 @@ import click
 import yaml
 import pandas as pd
 
-from sssom.constants import SUBJECT_ID, OBJECT_ID, PREDICATE_MODIFIER
+from sssom.constants import SUBJECT_ID, OBJECT_ID
 from sssom.util import filter_prefixes, is_curie, is_iri
 from sssom.parsers import parse_sssom_table
 from sssom.writers import write_table
-from sssom.io import get_metadata_and_prefix_map, filter_file
+from sssom.io import filter_file
 from bioregistry import curie_from_iri
 
 SRC = Path(__file__).resolve().parents[1]
@@ -49,6 +51,7 @@ output_option = click.option(
 )
 
 
+# todo: duplicated code fragment w/ lexmatch-sssom-compare: solution, move to a lexmatch_utils.py and import to both
 @click.group()
 @click.option("-v", "--verbose", count=True)
 @click.option("-q", "--quiet")
@@ -83,11 +86,11 @@ def main(verbose: int, quiet: bool):
 )
 @output_option
 def run(input: str, config: str, rules: str, rejects: str, output: str):
-    # Implemented `meta` param in `lexical_index_to_sssom`
-
-    meta = get_metadata_and_prefix_map(config)
+    """Run the script"""
+    # Get metadata config
     with open(config, "r") as f:
         yml = yaml.safe_load(f)
+    converter = Converter.from_extended_prefix_map(yml.pop('extended_prefix_map', {}))
 
     # Get mondo.sssom.tsv
     mapping_msdf = parse_sssom_table(SSSOM_MAP_FILE)
@@ -108,8 +111,6 @@ def run(input: str, config: str, rules: str, rejects: str, output: str):
     #     .reset_index(drop=True)
     # )
 
-    prefix_of_interest = yml["subject_prefixes"]
-
     resource = OntologyResource(slug=f"sqlite:///{Path(input).absolute()}")
     oi = SqlImplementation(resource=resource)
     ruleset = load_mapping_rules(rules)
@@ -118,9 +119,9 @@ def run(input: str, config: str, rules: str, rejects: str, output: str):
     save_lexical_index(lexical_index, OUT_INDEX_DB)
 
     if rules:
-        msdf = lexical_index_to_sssom(oi, lexical_index, ruleset=ruleset, meta=meta)
+        msdf = lexical_index_to_sssom(oi, lexical_index, ruleset=ruleset, prefix_map=converter)
     else:
-        msdf = lexical_index_to_sssom(oi, lexical_index, meta=meta)
+        msdf = lexical_index_to_sssom(oi, lexical_index, prefix_map=converter)
 
     # msdf.prefix_map = sssom_yaml['curie_map']
     # msdf.metadata = sssom_yaml['global_metadata']
@@ -131,8 +132,9 @@ def run(input: str, config: str, rules: str, rejects: str, output: str):
     # msdf.df[OBJECT_ID] = msdf.df[OBJECT_ID].apply(
     #     lambda x: iri_to_curie(x) if x.startswith("<http") else x
     # )
+    prefixes_of_interest = yml["subject_prefixes"]
     msdf.df = filter_prefixes(
-        df=msdf.df, filter_prefixes=prefix_of_interest, features=[SUBJECT_ID, OBJECT_ID]
+        df=msdf.df, filter_prefixes=prefixes_of_interest, features=[SUBJECT_ID, OBJECT_ID]
     )
     msdf.remove_mappings(mapping_msdf)
 
--- a/python-requirements-apple-silicon.txt
+++ b/python-requirements-apple-silicon.txt
@@ -15,7 +15,7 @@ charset-normalizer==3.3.2
 class-resolver==0.4.2
 click==8.1.7
 colorama==0.4.6
-curies==0.7.4
+curies==0.7.6
 Deprecated==1.2.14
 deprecation==2.1.0
 distlib==0.3.7
--- a/python-requirements.txt
+++ b/python-requirements.txt
@@ -22,7 +22,7 @@ class-resolver==0.4.2
 click==8.1.7
 colorama==0.4.6
 commonmark==0.9.1
-curies==0.6.4
+curies==0.7.6
 decorator==5.1.1
 Deprecated==1.2.13
 deprecation==2.1.0


src/ontology/imports/omo_import.owl
```diff
@@ -7,9 +7,9 @@ Prefix(rdfs:=<http://www.w3.org/2000/01/rdf-schema#>)
 
 
 Ontology(<http://purl.obolibrary.org/obo/mondo-ingest/imports/omo_import.owl>
-<http://purl.obolibrary.org/obo/mondo-ingest/releases/2023-11-30/imports/omo_import.owl>
+<http://purl.obolibrary.org/obo/mondo-ingest/releases/2024-01-11/imports/omo_import.owl>
```
joeflack4, line 10: ## `OMO` & `RO` (FYI)
<p>

I needed to run the `make` pipeline to get the prereqs to run the lexmatch script. These files got updated.

I pushed to main already, and when I do the diff locally, it doesn't show any diff for these 2 files, so this feels like a GitHub bug to me.

</p>
</details>

src/ontology/metadata/mondo.sssom.config.yml
```diff
@@ -34,7 +34,7 @@ curie_map:
   semapv: https://w3id.org/semapv/vocab/
   rdf: http://www.w3.org/1999/02/22-rdf-syntax-ns#
   sssom: https://w3id.org/sssom/
-  oio: http://www.geneontology.org/formats/oboInOwl#
+#  oio: http://www.geneontology.org/formats/oboInOwl#
```
joeflack4, line 37: ## New bug: Duplicate URI stems @joeflack4 @matentzn 
@matentzn Any ideas are welcome.
<p>
@joeflack4 Address issue

We have duplicate URI stems for the following CURIEs:
- oio: http://www.geneontology.org/formats/oboInOwl#
- oboInOwl: http://www.geneontology.org/formats/oboInOwl#

I know that there is a way to use `curies` to set up aliases. However I am not sure where the fix for this needs to be; it might need to go in several places:
1. `mondo.sssom.config.yml`: Had the duplicate entry; may/may not need updates
  For now, I commented one entry out hoping for a quick temporary fix. Did not work; errors in OAK (see below).
  Also, I need to figure out if this file should be changes, and if that is within the scope of this PR. Perhaps there is a way for us to update this file so that it is reflective of how we would set up the `Converter`, e.g. we could declare which prefixes are authoritative and which ones are aliases here. And for that, I would need to finally learn a bit more about these [extended prefix maps](https://curies.readthedocs.io/en/latest/struct.html#extended-prefix-maps).
2. SSSOM: Probably doesn't need a PR
  The `mondo-ingest` lexmatch calls SSSOM, which then creates the `Converter` object, but it doesn't set up aliases. Seems like this needs to be updated at some point to handle these situations. This may not be a blocker though. No error is produced; I just do a little patch right now on the metadata that SSSOM is returning back in `mondo-ingest`. That is so far sufficient.
3. OAK: Probably needs a PR
  Where I'm currently stuck (as of Jan 11) is in `lexical_indexer.py`, in `lexical_index_to_sssom()`. At the end, it does `msdf.clean_prefix_map(strict=ensure_strict_prefixes)` and fails because it detects the same duplicate prefixes (`oio` and `oboInOwl`). As with SSSOM, there is a `Converter` object being created her which does not handle prefix aliases. I think I'll likely have to change the code to fix this. I'm currently getting:
```
ValueError: new record already exists and merge=False: {('oboInOwl', 'http://www.geneontology.org/formats/oboInOwl#', 'oio', ''): ['prefix match']}
```

</p>
</details>

src/scripts/match-mondo-sources-all-lexical.py
```diff
@@ -83,18 +83,21 @@ def main(verbose: int, quiet: bool):
 )
 @output_option
 def run(input: str, config: str, rules: str, rejects: str, output: str):
-    # Implemented `meta` param in `lexical_index_to_sssom`
-
-    meta = get_metadata_and_prefix_map(config)
+    sssom_meta = get_metadata_and_prefix_map(config)
+    oak_meta = {'curie_map': sssom_meta[0].prefix_map}
```
joeflack4, line 87: ## Bugfix for `AttributeError: 'tuple' object has no attribute 'pop'` @matentzn @hrshdhgd 
@hrshdhgd I don't think is going to be back for a while so, @matentzn should just review.

<p>

The error was being thrown in OAK, in `lexical_indexer.py`, in `lexical_index_to_sssom()`. However the problem is that the wrong thing was being passed. `lexical_index_to_sssom()` asks for a `meta` object of type `Dict`, but the `meta` that was being passed was a `tuple[Converter, Dict]`. It seemed that what `lexical_index_to_sssom()` really wanted was a dictionary with a `curie_map` in it. 

## Possible side effects
`meta` is used in 2 places in `lexical_index_to_sssom():
1. `Converter.from_prefix_map((meta or {}).pop(CURIE_MAP, {})),`
2. `msdf = MappingSetDataFrame.from_mappings(mappings=mappings, metadata=meta, converter=converter)`

My fix above fixes the problem for (1). But for (2), I don't yet know how `meta` ends up getting used, so I do not know if it would benefit more from just `curie_map` being in `meta`. I don't know how this was when Harshad originally set it up. And when I interrogate the code, it gets difficult to inspect, so my decision for now is: If (2) executes without error, I'm considering that fine for now. My other failsafe for this is that when this PR is done, I'll provide some outputs and have Nico check them. Hopefully, if there's any negative side effects caused by my necessary changes to `meta`, we will notice at that stage.

</p>
</details>

src/scripts/match-mondo-sources-all-lexical.py
```diff
@@ -83,18 +83,21 @@ def main(verbose: int, quiet: bool):
...
     with open(config, "r") as f:
         yml = yaml.safe_load(f)
 
+    # TODO: PR comment: This seems slow
+    #  - if can't speed up, maybe move it down to where it is needed?
     # Get mondo.sssom.tsv
+    # TODO: <uncomment
     mapping_msdf = parse_sssom_table(SSSOM_MAP_FILE)
```
joeflack4, line 95: ## `parse_sssom_table()` (FYI)
<p>

I didn't time it strictly but it took a while to do this. I'm not sure what's taking so long.

Addressing this isn't really within scope for this PR, just bringing attention.

</p>
</details>

src/scripts/match-mondo-sources-all-lexical.py
```diff
@@ -109,18 +112,25 @@ def run(input: str, config: str, rules: str, rejects: str, output: str):
...
+    # TODO: uncomment
+    # lexical_index = create_lexical_index(oi=oi, mapping_rule_collection=ruleset)
+    # save_lexical_index(lexical_index, OUT_INDEX_DB)
+    # TODO: /uncomment
+
+    # TODO temp delete after
+    import pickle
+    pp = '/Users/joeflack4/Desktop/lexical_index.pickle'
+    # pickle.dump(lexical_index, open(pp, "wb"))
+    lexical_index = pickle.load(open(pp, "rb"))
```
joeflack4, line 128: ## Caching `.pickle` > `.yml` for `tmp/merged.db.lexical.yaml`? @matentzn @hrshdhgd 
@hrshdhgd I don't think will be able to review this but @matentzn if you could consider my suggestions.

<p>

### Add caching
For development, caching helps. These two lines for `create_lexical_index()` and `save_lexical_index()` are taking 20 minutes. 

### Cache `.pickle` > `.yml`?
The `tmp/merged.db.lexical.yaml` currently being saved is 266MB, and I don't see it being used anywhere. Maybe it is for manual inspection / troubleshooting purposes?

Also, it takes a long time to save. I'm not sure how long. The way I executed it, PyCharm crashed.

If we add caching, I would suggest pickle. It both saved and loaded quickly, and was only 750KB. I was shocked at the size difference from the 266MB `yml`. I wonder if there was some lossiness in the pickling, but I don't yet know how to even check for that. Either way, I've been running this and reloading from pickle, and it has not seemed to cause any issues. The script still continues to run for another 45 minutes.

</p>
</details>

src/ontology/metadata/mondo.sssom.config.yml
matentzn: As for (1), please also note https://github.com/monarch-initiative/mondo-ingest/pull/388

For (3), is there no way we can avoid conflicting prefixes going in, e.g. (1)?

Else I am not sure how to deal with this atm..

src/scripts/match-mondo-sources-all-lexical.py
matentzn: Yes, dont worry about this here. Latest version of sssom-py?

src/scripts/match-mondo-sources-all-lexical.py
matentzn: Maybe make a separate issue about this and discuss with Trish and Harshad when you see them?

src/scripts/match-mondo-sources-all-lexical.py
matentzn: I think it is used - but maybe this is not relevant for the mondo-ingest pipeline. We can use other postprocessing methods to append any metadata if needed. Probably your fix is the right direction.

src/ontology/metadata/mondo.sssom.config.yml
joeflack4: On #388 I took a look; noted!
For (3), the OAK code. I need more time to look into it. I removed conflicting prefixes from (1) but still got the error.

src/scripts/match-mondo-sources-all-lexical.py
joeflack4: Sounds good. I'll leave this open until we've reviewed the outputs to make sure they look good then I'll mark resolved

src/scripts/match-mondo-sources-all-lexical.py
joeflack4: Good point I probably need to upgrade, I'll check.

src/scripts/match-mondo-sources-all-lexical.py
```diff
@@ -83,18 +83,21 @@ def main(verbose: int, quiet: bool):
 )
 @output_option
 def run(input: str, config: str, rules: str, rejects: str, output: str):
-    # Implemented `meta` param in `lexical_index_to_sssom`
-
-    meta = get_metadata_and_prefix_map(config)
+    sssom_meta = get_metadata_and_prefix_map(config)
```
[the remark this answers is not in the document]
joeflack4: I like this suggestion. Rather than a new commit, I'll amend it in.

src/ontology/metadata/mondo.sssom.config.yml
matentzn: Make sure you try this with all changes in #388 as well! The error may be coming from a different prefix map with the same conflicts.

joeflack4, line 37: ## Partial or merged EPM?
@cthoyt Right now we only have `oio` and `oboInOwl` in conflict. I can think of several ways to address this. I could convert our entire flat prefix map into an EPM, and I may well do that. But I wonder about your thoughts. Consider:


Plain prefix map:
```yaml
  MONDO: http://purl.obolibrary.org/obo/MONDO_
  UMLS: http://linkedlifedata.com/resource/umls/id/
  NCIT: http://purl.obolibrary.org/obo/NCIT_
  DOID: http://purl.obolibrary.org/obo/DOID_
```

EPM:
```python
{
   'prefix': 'oboInOwl',
   'prefix_synonyms': [],
   'uri_prefix': 'http://www.geneontology.org/formats/oboInOwl#',
   'uri_prefix_synonyms': []
}
```

Let's say I wanted to merge the two programmatically. What do you think would be the best option:
a. A new complex data structure that allows inclusion of both a plain flat prefix map as well as an EPM
b. Keep these two structures separate, but use some code that merges them together, e.g.
  b.1. by converting the flat prefix map to an EPM and merges that
  b.2. a method that merges 2 `Converter` objects (`curies.chain()`?)
c. something else?

matentzn: Making sure you saw my comment

cthoyt: I would highly suggest converting the flat prefix map into an EPM as a first step and have a single source of truth. You're asking good questions that all boil down to: how do I manage my prefix map, given there are so many different streams from which definitions of CURIE prefix-URI prefix pairs come from?


This problem is so complicated that we built a solution into the Bioregistry https://bioregistry.io/context. My preferred solution is to bake in the Monarch community rules into a "context" definition, then all of the merging and reconciliation get taken care of by the Bioregistry and give you an EPM. @matentzn has already done this to make an "OBO" context.

matentzn: Just as an aside this here has nothing at all to do with any "Monarch" prefix maps - this is all ETL, build long before curies and EPMs; I agree with constructing an EPM, if it is necessary at all - I just want to make sure why understand the problem and _that_ it is necessary, or we can get away with simply deleting the conflicting entires in the prefix maps.

joeflack4: Thanks for getting back, @cthoyt.

@matentzn I explained a few times in https://github.com/mapping-commons/sssom-py/pull/485 that deleting the conflicting entry doesn't work. A prefix map is getting dynamically generated based on what OAK is interrogating from Mondo. Even if I were to delete `oio` or `oboInOwl` after it was dynamically generated, it will throw an error later because it can't resolve.

joeflack4: I saw it, but as I explained in [this comment](https://github.com/monarch-initiative/mondo-ingest/pull/394#discussion_r1456679001) and in other comments in the [other PR](https://github.com/mapping-commons/sssom-py/pull/485), removing the a duplicate prefix from these files won't fix the issue. For clarification, the only config #388 the `mondo.sssom.config.yml`, and I tried deleting the duplicate in there already (as discussed in other PR) and it didn't work.

src/ontology/metadata/mondo.sssom.config.yml
```diff
@@ -75,6 +75,304 @@ curie_map:
   GC_ID: "http://purl.obolibrary.org/obo/mondo/mappings/unknown_prefix/GC_ID/"
   SNOMEDCT_2010_1_31: "http://purl.obolibrary.org/obo/mondo/mappings/unknown_prefix/SNOMEDCT_2010_1_31/"
 
+extended_prefix_map:
```
joeflack4, line 78: ## `extended_prefix_map` in `mondo.sssom.config.yml`
@matentzn Here's the new EPM. I added it within `mondo.sssom.config.yml`. I have no idea however if this could have unintended side effects, so another consideration is that I could make a separate file for the EPM entirely and load that instead.

src/ontology/config/prefixes.csv
```diff
@@ -3,7 +3,7 @@ rdf,http://www.w3.org/1999/02/22-rdf-syntax-ns#
 rdfs,http://www.w3.org/2000/01/rdf-schema#
 xsd,http://www.w3.org/2001/XMLSchema#
 owl,http://www.w3.org/2002/07/owl#
-oio,http://www.geneontology.org/formats/oboInOwl#
```
joeflack4, line 6: ## `oio` removed from `prefixes.csv`

src/ontology/metadata/mondo.sssom.config.yml
```diff
@@ -34,7 +34,7 @@ curie_map:
   semapv: https://w3id.org/semapv/vocab/
   rdf: http://www.w3.org/1999/02/22-rdf-syntax-ns#
   sssom: https://w3id.org/sssom/
-  oio: http://www.geneontology.org/formats/oboInOwl#
```
joeflack4, line 37: ## `oio` removed from `curie_map` in `mondo.sssom.config.yml`

src/scripts/match-mondo-sources-all-lexical.py
```diff
@@ -145,6 +194,9 @@ def run(input: str, config: str, rules: str, rejects: str, output: str):
     kwargs = {"subject_id": ("MONDO:%",), "object_id": prefix_args}
     with open(str(Path(output.replace("lexical", "lexical-2"))), "w") as f:
         filter_file(input=str(Path(output)), output=f, **kwargs)
+    t1 = datetime.now()  # todo: temp
+    print('match-mondo-sources-all-lexical complete in seconds:', (t1 - t0).seconds)  # todo temp
```
joeflack4, line 198: ## Outputs
@matentzn @twhetzel I will be uploading [outputs here](https://docs.google.com/spreadsheets/d/1topWKAf4-biD_dbmies-P3KQB53MKlovxIpUw0EYfDI/edit#gid=1579419867).

src/ontology/config/prefixes.csv
matentzn: > The negative is that now OAK is not checking for lexical matches on synonyms identified by the oio:hasExactSynonym field.

This makes no sense. 

I will look into it...

src/ontology/metadata/mondo.sssom.config.yml
matentzn: `HGNC__2` should be merged into HGNC to be a proper EPM!

src/ontology/metadata/mondo.sssom.config.yml
```diff
@@ -75,6 +75,304 @@ curie_map:
...
+    prefix_synonyms: []
+    uri_prefix: http://purl.obolibrary.org/obo/mondo/mappings/unknown_prefix/IEDB/
+    uri_prefix_synonyms: []
+  - prefix: PMID
+    prefix_synonyms: []
+    uri_prefix: http://purl.obolibrary.org/obo/mondo/mappings/unknown_prefix/PMID/
+    uri_prefix_synonyms: []
+  - prefix: KEGG
+    prefix_synonyms: []
+    uri_prefix: http://purl.obolibrary.org/obo/mondo/mappings/unknown_prefix/KEGG/
```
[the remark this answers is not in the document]
matentzn: Dont worry, none of this is related to prefixmaps, bioregistry or anything else for that matter. These are internally used for ETL, nothing else.

src/ontology/config/prefixes.csv
joeflack4: @matentzn If I understand you correctly, that's what I'm saying. If `prefixes.csv` contains `oio`, then when I run the pipeline, it does matches on just on `rdfs:label` but also `oio:hasExactSynonym`, and you can see that reflected in my [subsample outputs](https://docs.google.com/spreadsheets/d/1topWKAf4-biD_dbmies-P3KQB53MKlovxIpUw0EYfDI/edit#gid=1579419867). And if I remove `oio` from `prefixes.csv` I no longer see those in the outputs.
So if we want to keep matching on `oio:hasExactSynonym`, it seems that we should keep `oio` in `prefixes.csv`, even though that causes it to be non-bijective. It's only passed to SemanticSQL anyway. Alternative is to ask OAK to rename `oio` to `oboInOwl`.

src/ontology/metadata/mondo.sssom.config.yml
joeflack4: Yep, no worries, I'm still on it!

joeflack4: Moved this work into #396, merging that, and doing a rebase here.

src/scripts/match-mondo-sources-all-lexical.py
```diff
@@ -108,19 +112,17 @@ def run(input: str, config: str, rules: str, rejects: str, output: str):
...
+    resource = OntologyResource(slug=f"sqlite:///{Path(_input).absolute()}")
     oi = SqlImplementation(resource=resource)
     ruleset = load_mapping_rules(rules)
     # syn_rules = [x.synonymizer for x in ruleset.rules if x.synonymizer]
     lexical_index = create_lexical_index(oi=oi, mapping_rule_collection=ruleset)
     save_lexical_index(lexical_index, OUT_INDEX_DB)
 
     if rules:
-        msdf = lexical_index_to_sssom(oi, lexical_index, ruleset=ruleset, meta=meta)
+        msdf = lexical_index_to_sssom(oi, lexical_index, ruleset=ruleset, prefix_map=converter)
```
joeflack4, line 123: ## `meta` -> `converter`
Along with removing `oio` from `config/prefixes.csv`, this is the essential bugfix. This allows us to use the new extended prefix map, and solves the `.pop()` error in the original issue #390.

joeflack4 (commented):
@matentzn @twhetzel I thought this was ready to merge because it ran successfully through the debugger. However I'm getting an error in OAK now: `AttributeError: 'Converter' object has no attribute 'items'`. I upgraded OAK within ODK but that didn't work. I need more time to fix this; hopefully can do before Monday.

src/ontology/config/prefixes.csv
joeflack4: _As of today here's where we are:_
Nico has looked into further and opened: https://github.com/INCATools/ontology-access-kit/issues/698

Joe looked to see other places in which `merged.db` is used for identifying possible side effects of changing `oio` to `oboInOwl` in `prefixes.csv` when constructing `merged.db`. These other usages were found, but do not know if unintended side effects could occur:
- `lexmatch-sssom-compare.py`
- `mappings.ipynb`
- `oak-lexmatch-sssom-compare.ipynb`
- `oak-lexmatch-sssom-compare-checkpoint.ipynb`
- `remove-mappings-test.ipynb`
- `synchronizeMappingsWithMondo-checkpoint.ipynb`

Nico said he would put in some time to solve this, so Joe is pausing activity on this.

src/scripts/match-mondo-sources-all-lexical.py
```diff
@@ -83,11 +86,11 @@ def main(verbose: int, quiet: bool):
 )
 @output_option
 def run(input: str, config: str, rules: str, rejects: str, output: str):
-    # Implemented `meta` param in `lexical_index_to_sssom`
-
-    meta = get_metadata_and_prefix_map(config)
+    """Run the script"""
```
joeflack4, line 89: ## `AttributeError: 'Converter' object has no attribute 'items'` _(fixed)_
You can mark this resolved. It's fixed.
<p>

I thought this was ready to merge because it ran successfully through the debugger. However I'm got error in OAK now: `AttributeError: 'Converter' object has no attribute 'items'`.

```
ls Traceback (most recent call last):
  File "/work/src/ontology/../scripts/[match-mondo-sources-all-lexical.py](http://match-mondo-sources-all-lexical.py/)", line 176, in <module>
    main()
  File "/usr/local/lib/python3.10/dist-packages/click/[core.py](http://core.py/)", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/click/[core.py](http://core.py/)", line 1078, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.10/dist-packages/click/[core.py](http://core.py/)", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.10/dist-packages/click/[core.py](http://core.py/)", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.10/dist-packages/click/[core.py](http://core.py/)", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/work/src/ontology/../scripts/[match-mondo-sources-all-lexical.py](http://match-mondo-sources-all-lexical.py/)", line 122, in run
    msdf = lexical_index_to_sssom(oi, lexical_index, ruleset=ruleset, prefix_map=converter)
  File "/usr/local/lib/python3.10/dist-packages/oaklib/utilities/lexical/[lexical_indexer.py](http://lexical_indexer.py/)", line 317, in lexical_index_to_sssom
    {k: v for k, v in prefix_map.items() if k not in meta.prefix_map.keys()}
AttributeError: 'Converter' object has no attribute 'items'
```

I upgraded OAK within ODK but that didn't work. However, upgrading my ODK image did.

</p>
</details>

src/scripts/match-mondo-sources-all-lexical.py
```diff
@@ -118,9 +119,9 @@ def run(input: str, config: str, rules: str, rejects: str, output: str):
     save_lexical_index(lexical_index, OUT_INDEX_DB)
 
     if rules:
-        msdf = lexical_index_to_sssom(oi, lexical_index, ruleset=ruleset, meta=meta)
+        msdf = lexical_index_to_sssom(oi, lexical_index, ruleset=ruleset, prefix_map=converter)
```
joeflack4, line 122: ## Redundancy with #397
The main difference is that #397 does passes `meta=None` to `lexical_index_to_sssom()`, whereas #394 passes a `Converter`. Supposedly #397 approach is fine.

Everything else in that #394 is now cosmetic, so i can merge just those changes afterwards or ignore and close.

joeflack4:
I'm going to close this, as this is essentially supplanted by #397. There's just some minor stylistic stuff in here that we can re-open and merge if we really want.

PR #397, opened 2024-01-22
Lexmatch revamp

matentzn:
This PR 

- adds a few missing prefixes to merged.db (through prefixes.csv) that were previously missed
- replaces the `prefix_map` with `converter` parameter in `src/scripts/lexmatch-sssom-compare.py`
- skips `meta` in `src/scripts/match-mondo-sources-all-lexical.py` which is not urgently needed and was causing all the problems @joeflack4 is fixing
- amends to preprocessing queries in the pipeline that added junk xrefs to the pipeline
- Updates `src/ontology/config/prefixes.csv` to cover all relevant prefixes for the matching
- updates `src/ontology/metadata/*` configs to avoid conflicting namespaces.

@joeflack4 I am not reading your answers etc - please review this PR with @twhetzel and see if you have any concern with these changes.

To test this PR if you like:

```
docker pull obolibrary/odkfull:dev
ODK_TAG=dev ./run.sh make lexmatch/README.md -B IMP=false PAT=false MIR=false
```

I ran it locally - now testing the whole pipeline (I circumvent the main problem you are trying to solve by simply ditching `meta` here, hoping its not needed).
```

PR #403, opened 2024-01-26
Reinstate sync pipeline

matentzn:
@joeflack4 @twhetzel 

This PR fixes the sync pipeline - in the end, the error was small, but it took quite some needle in the haystack searching.. 

I would suggest to review this, merge it into #397 (this is where this PR is directed towards), and then merge #397 as a whole.

Once its done, I will rerun everything again.


src/scripts/sync_subclassof.py
```diff
@@ -281,6 +281,10 @@ def sync_subclassof(
     logging.info('- Collecting direct subclass relations')
     rels_direct_source_source: Set[RELATIONSHIP] = _get_direct_scr_rels(  # from source
         source_ids, ingest_db, owned_prefix_map)
+    
+    if not rels_direct_source_source or len(rels_direct_source_source) < 1:
+        raise ValueError("Source has no relationships. Something went wrong")
```
matentzn, line 286: I add this here because during debugging, I had an issue where one of the sources was not present in the mondo-ingest.db (for some unknown reason), and it would have helped me if they pipeline had failed at this point if no relationships were identified.

There are some corner cases, for example, when zero concepts are mapped to a new source, where this error is false alarm, but I think in the grand scheme of things this indicates a really important issue with one of the inputs.

src/scripts/sync_subclassof.py
```diff
@@ -304,11 +308,13 @@ def sync_subclassof(
...
     missing_ancestor_rels = rels_indirect_mondo_mondo.union(rels_direct_mondo_mondo).difference(ancestors_mondo_mondo)
     if missing_ancestor_rels:
-        raise RuntimeError(
+        rels_for_printing =  [rel for rel in list(missing_ancestor_rels)[:5]]
+        logging.error(
             'Detected error in consistency of sets of terms gathered from Mondo.\n'
             f'\n 1. Mondo SCR ancestors: {len(ancestors_mondo_mondo)}'
             f'\n 2. Mondo direct SCR relationships: {len(rels_direct_mondo_mondo)}'
             f'\n 3. Mondo indirect SCR relationships: {len(rels_indirect_mondo_mondo)}'
+            f'\n Intersection (Top 5): {rels_for_printing}'
```
matentzn, line 317: Added this here just so it can be debugged.

src/scripts/sync_subclassof.py
```diff
@@ -304,11 +308,13 @@ def sync_subclassof(
     #  after. It could be that I invalidated the cache somehow without realizing it. Rerunning cache fixed it though.
     missing_ancestor_rels = rels_indirect_mondo_mondo.union(rels_direct_mondo_mondo).difference(ancestors_mondo_mondo)
     if missing_ancestor_rels:
-        raise RuntimeError(
```
matentzn, line 307: Can you @joeflack4 review the current case? There are seemingly 11 relationships including

```
 Intersection (Top 5): [('MONDO:0014937', 'rdfs:subClassOf', 'MONDO:0003008'), ('MONDO:0014938', 'rdfs:subClassOf', 'MONDO:0002466'), ('MONDO:0011790', 'rdfs:subClassOf', 'MONDO:0100500'), ('MONDO:0014937', 'rdfs:subClassOf', 'MONDO:0002466'), ('MONDO:0100534', 'rdfs:subClassOf', 'MONDO:0006260')]
```

in that list. I would suggest:

1. Create a new issue about this, which suggests changing this back to `raise RuntimeError`.
2. Determining why the 11 cases occur and fix the route cause.

Fixes in new PR.

src/scripts/sync_subclassof.py
```diff
@@ -378,7 +384,7 @@ def sync_subclassof(
 
     # Case 3: SCR is direct in the source, but not at all in Mondo
     subheader = deepcopy(ROBOT_SUBHEADER)
-    subheader[0]['object_mondo_id'] = 'AI obo:mondo#:excluded_subClassOf'
+    subheader[0]['object_mondo_id'] = 'AI obo:mondo#excluded_subClassOf'
```
matentzn, line 387: Typo.

src/ontology/mondo-ingest.Makefile
```diff
@@ -515,7 +522,7 @@ sync-subclassof-%:
 	$(MAKE) $(REPORTDIR)/$*.subclass.direct-in-mondo-only.tsv
 
 # Side effects: Deletes SOURCE.subclass.direct-in-mondo-only.tsv's from which the combination is made.
-$(REPORTDIR)/sync-subClassOf.direct-in-mondo-only.tsv: $(foreach n,$(ALL_COMPONENT_IDS), sync-subclassof-$(n))
+$(REPORTDIR)/sync-subClassOf.direct-in-mondo-only.tsv: $(foreach n,$(ALL_COMPONENT_IDS), sync-subclassof-$(n)) tmp/mondo.db
```
matentzn, line 525: This was not the problem, but since `sync_subclassof_collate_direct_in_mondo_only` depends on `tmp/mondo.db` its better to spell it out.

In general, try always to declare the inputs of a python script, dont assume presence of files etc in a python script.

src/ontology/mondo-ingest.Makefile
```diff
@@ -422,6 +426,9 @@ $(MAPPINGSDIR)/rejected-mappings.tsv:
 $(MAPPINGSDIR)/rejected-mappings-sssom.tsv: $(MAPPINGSDIR)/rejected-mappings.tsv
 	sssom parse $< -m metadata/mondo.sssom.config.yml --no-strict-clean-prefixes -o $@
 
+tmp/mondo-ingest.owl: mondo-ingest.owl
+	cp $< $@
```
matentzn, line 430: This was the source of the problem.

1. The sync pipeline depends on tmp/mondo-ingest.db
2. tmp/mondo-ingest.db depends ofn tmp/mondo-ingest.owl
3. There was no goal for tmp/mondo-ingest.owl

src/ontology/mondo-ingest.Makefile
```diff
@@ -32,13 +32,16 @@ dependencies:
...
 %.db: %.owl
+	@rm $*.db
 	@rm -f .template.db
 	@rm -f .template.db.tmp
 	@rm -f $*-relation-graph.tsv.gz
 	RUST_BACKTRACE=full semsql make $*.db -P config/prefixes.csv
 	@rm -f .template.db
 	@rm -f .template.db.tmp
 	@rm -f $*-relation-graph.tsv.gz
+	@test -f $*.db || (echo "Error: File not found!" && exit 1)
```
matentzn, line 43: The above pipeline (`semsql make`) does not fail hard, so this goal (`%.db`) always succeeds. This way, we make it fail (by first deleting the db file, and after the pipeline has finished checking if it is there.

.gitignore
```diff
@@ -110,3 +110,6 @@ src/scripts/mondo_unmapped.tsv
 # Test
 tests/output/
 src/scripts/dataframes/*
+src/ontology/reports/gard.subclass.added-obsolete.robot.tsv
+src/ontology/reports/gard.subclass.added.robot.tsv
+src/ontology/reports/gard.subclass.confirmed.robot.tsv
```
matentzn, line 115: Removed the gard ones as they are meaningless for this process.

matentzn (commented):
@joeflack4 I made comments in all the relevant sections.

.gitignore
joeflack4: Sounds good. I have it high on my todo list to do teardown of GARD.

src/ontology/mondo-ingest.Makefile
joeflack4: Hmm, that's unfortunate that it doesn't fail hard when it fails. Should I make an issue about that in the 
Your `test -f` solution is good.

However, `rm $*.db` will fail if you run this goal and the file doesn't already exist, causing `make` to error out. `rm -f $*.db` should solve.

src/ontology/mondo-ingest.Makefile
```diff
@@ -32,13 +32,16 @@ dependencies:
 ### General ########################
 ####################################
 %.db: %.owl
+	@rm $*.db
```
joeflack4, line 35:
```suggestion
	@rm -f $*.db
```

src/ontology/mondo-ingest.Makefile
joeflack4: Ugh, I am thinking about this and I thought this was something I'd addressed. But perhaps what I changed was actually moving `merged.db` to `tmp/merged.db`. 

Sorry I never caught this!

Do I need to update my development workflow to the following?
Make sure code works in the following ways, in order.
1. Debugging in my IDE
2. `sh run.sh make ...`
3. `sh run.sh make ... -B`
4. `build-mondo-ingest`

I've only been doing (1) and (2). (4) is something we're all thinking about making easier, but I don't think we can expect to do 100% of the time. Should I be adding (3) to my workflow? Would that have caught this problem? If so, do I do (3) and just never need to do (2)?

src/ontology/mondo-ingest.Makefile
joeflack4: Ah, an oversight. Thanks for correcting!

src/scripts/sync_subclassof.py
joeflack4: Ash, yes that sounds like a good idea. Maybe I didn't add this initially because the GARD files had no rows. But then when GARD became no longer needed, I didn't think to add this.

src/scripts/sync_subclassof.py
joeflack4: Hmm, I was hoping this error would never happen again.

Sure, I'll do that. "Fixes in new PR" sounds like part of (2). Is "new PR" one that is currently open or not one that exists yet?

Are you asking me just to review the case, which seems like tasks (1). I assume you are also asking me to take a stab at (2) as well?

src/scripts/sync_subclassof.py
joeflack4: Typo and bug! Sorry.

src/ontology/mondo-ingest.Makefile
```diff
@@ -422,6 +426,9 @@ $(MAPPINGSDIR)/rejected-mappings.tsv:
 $(MAPPINGSDIR)/rejected-mappings-sssom.tsv: $(MAPPINGSDIR)/rejected-mappings.tsv
 	sssom parse $< -m metadata/mondo.sssom.config.yml --no-strict-clean-prefixes -o $@
 
+tmp/mondo-ingest.owl: mondo-ingest.owl
```
joeflack4, line 429: ## Suggestion -> `tmp/mondo-ingest.db: mondo-ingest.db`
<p>

### Changes
(1) Change the prereq here to `mondo-ingest.db`
```make
$(REPORTDIR)/%.subclass.added.robot.tsv $(REPORTDIR)/%.subclass.confirmed.robot.tsv $(REPORTDIR)/%.subclass.direct-in-mondo-only.tsv $(REPORTDIR)/%.subclass.added-obsolete.robot.tsv: tmp/mondo-ingest.db tmp/mondo.db tmp/mondo.sssom.tsv
```

(2) For the same target as above, remove `$(TMPDIR)` from this line
```make
--mondo-ingest-db-path $(TMPDIR)/mondo-ingest.db
```

(2) Delete goal for `tmp/mondo-ingest.owl`

### Rationale
We already have a `mondo-ingest.owl`. We can just make a `mondo-ingest.db`. There is not need I think for a duplicate file in `tmp/`. Also, the `.gitignore` is already set up for `*.db` so no worry about it getting committed.

I felt like we had already made this change months ago, but I think what happened was we had a similar situation for `mergd.owl`/`.db` and `tmp/merged.owl`/`.db`. We decided to stick with only 1 copy of those files (the `tmp/` version).

</p>
</details>

src/ontology/mondo-ingest.Makefile
matentzn: Great suggestion, merged.

src/ontology/mondo-ingest.Makefile
matentzn: I like as few files as possible in the src/ontology dir -> hence that suggestion..

joeflack4: ## Suggestion -> `mondo-ingest.owl`/`.db` -> `tmp/`
How about this, then? `mondo-ingest.owl` is actually in the `.gitignore`. We could just move it and its `.db` into `tmp/`.

src/ontology/mondo-ingest.Makefile
twhetzel: At some point I'd include `sh run.sh make ... -B` so that everything is rebuilt, but I am also wondering how versions of all targets are being captured between runs to know that the inputs are the same from one run to the next.

src/scripts/sync_subclassof.py
```diff
@@ -308,7 +308,7 @@ def sync_subclassof(
     #  after. It could be that I invalidated the cache somehow without realizing it. Rerunning cache fixed it though.
     missing_ancestor_rels = rels_indirect_mondo_mondo.union(rels_direct_mondo_mondo).difference(ancestors_mondo_mondo)
     if missing_ancestor_rels:
-        rels_for_printing =  [rel for rel in missing_ancestor_rels[:5]]
+        rels_for_printing =  [rel for rel in list(missing_ancestor_rels)[:5]]
```
twhetzel, line 311: Curious, why is it limited to the first 5?

twhetzel (approved):
looks good to me

src/scripts/sync_subclassof.py
matentzn: Will answer in PR later

src/scripts/sync_subclassof.py
matentzn: There could be 10000s..

src/ontology/mondo-ingest.Makefile
matentzn: `mondo-ingest.owl` is in "gitignore", but it is used by "make deploy-release". We can reconsider this when we decide that we want to publish a db version of Mondo ingest for easier access in OAK by other parties.

joeflack4: (1) I'm not talking about publishing a `.db` version. I'm just think we have needless duplication of goals / files here.

(2) By `deploy-release` (I don't see it) do you mean `deploy-mondo-ingest`? It's NP to me that it's in the `.gitignore` but also in a release. Usually that's the case for release artefacts.

(3) There are no prereqs for `deploy-mondo-ingest`. Not a huge deal but I suppose all the release artefacts should be listed.

matentzn: Yes, we decided to decouple deploy commands from everything as they are dependent on files after the are moved to the release directory. I can explain in a call how this works if you like details!

joeflack4: That answers (3). I can just take your word for it.

That doesn't answer my questions (1) and (2) above. But we have bigger fish to fry if you don't wanna worry about it.

matentzn: (2) `deploy-mondo-ingest` -> Yes, that is what I meant. It picks the release files from the release directory and uploads them to github (just did it yesterday in #417)

(1) Its a value judgement. I kinda dont like (at all), that ODK puts _anything_ (any intermediate file) in `src/ontology`. That was a design mistake. It makes that folder, which has key confguration, feel cramped an unwieldy. I would prefer all files, or at least as much as possible, to be generated in `tmp/` to be less cramped. I am not taking a super hard stance here though and can be convinced otherwise.

joeflack4: thanks for providing me those answers!

Issue #406, opened 2024-01-28
Bug: Sync subclass: ancestor inconsistencies


joeflack4:
## Overview
Nico was trying to fix some bugs in the pipeline (#403) when he noticed [an issue regarding ancestor inconsistencies](https://github.com/monarch-initiative/mondo-ingest/pull/403/files#r1467619776). Basically, we expect that `ancestor()` edges should give the same results as the indirect edges + direct edges. But we're getting an error where that is not the case.

## Sub-tasks
- [ ] 1. Find root cause of issue & fix
  - May include #408
- [ ] 2. Consider changing back: `logging.error()` -> `raise ValueError()`
- [ ] 3. Change logger to print info?

### 3. Change logger to print info?
<details><summary>Details</summary>
<p>

This is unrelated to this issue, but basically I was expecting `logging.info()` to print, but it is not. It would help me to debug if this would print. Should we print these? Should we change these to `logging.debug()` and maybe I change my code so that if I'm running what I typically run in the debugger, `run_defaults()`, it sets the logging to debug mode?

Shamefully I don't have enough experience with logging. I tried to fix with some suggestions [here](https://stackoverflow.com/questions/7016056/python-logging-not-outputting-anything), but most of what I tried didn't work (nor did ChatGPT work). One suggestion did work, but it ended up printing everything twice.

</p>
</details>

joeflack4:
@matentzn Frustratingly, I can't replicate your issue. I probably screwed up somewhere along the way. Maybe you can figure it out? Here's what I tried.

## What I did
### Setup 1: My current `mondo-ingest` clone
1. Branch off of your [`reinstate-sync` PR](https://github.com/monarch-initiative/mondo-ingest/pull/403/)
2. Make new branch / [PR `bug-sync-ancestor`](https://github.com/monarch-initiative/mondo-ingest/pull/408)
3. Add [`%.db` bugfix I suggested](https://github.com/monarch-initiative/mondo-ingest/pull/403/files#r1468197406). Otherwise it was erroring out for me as expected.
4. `rm src/ontology/reports/*subclass*`
5. `Deleted and remade subclass prereqs `tmp/mondo-ingest.db tmp/mondo.db tmp/mondo.sssom.tsv`
  - Pretty sure I did this without also adding `IMP=false COMP=false MIR=false`. I know at least they changed somehow because I backed them up first and I just checked `md5` and they are different.
[...]

PR #408, opened 2024-01-28
Bug: Sync subclass: ancestor inconsistencies

joeflack4:
## Changes
Bugfix sync ancestors
- Bugfix: Fixed bug in .ancestors() results were not consistent with direct + indirect results.

## Status
Not ready for review yet. Still trying to replicate issue. See: [comment](https://github.com/monarch-initiative/mondo-ingest/issues/406#issuecomment-1913801438).

## Addresses
Addresses part of:
- #408

### DIFF
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -345,13 +345,17 @@ documentation: j2 $(ALL_DOCS) unmapped-terms-docs mapped-deprecated-terms-docs s
 build-mondo-ingest:
 	$(MAKE) refresh-imports exclusions-all slurp-all mappings matches \
 		mapped-deprecated-terms mapping-progress-report \
-		recreate-unmapped-components sync documentation
-	$(MAKE) prepare_release
+		recreate-unmapped-components sync documentation \
+		prepare_release
 	echo "Mondo Ingest has been fully completed"
 
 .PHONY: build-mondo-ingest-no-imports
 build-mondo-ingest-no-imports:
-	$(MAKE_FAST) build-mondo-ingest
+	$(MAKE_FAST) exclusions-all slurp-all mappings matches \
+		mapped-deprecated-terms mapping-progress-report \
+		recreate-unmapped-components sync documentation \
+		prepare_release
+	echo "Mondo Ingest has been fully completed"
 
 DEPLOY_ASSETS_MONDO_INGEST=$(OTHER_SRC) $(ALL_MAPPINGS) ../../mondo-ingest.owl ../../mondo-ingest.obo
 
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -32,7 +32,7 @@ dependencies:
 ### General ########################
 ####################################
 %.db: %.owl
-	@rm $*.db
+	@rm -f $*.db
 	@rm -f .template.db
 	@rm -f .template.db.tmp
 	@rm -f $*-relation-graph.tsv.gz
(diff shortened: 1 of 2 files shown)


src/scripts/sync_subclassof.py
```diff
@@ -499,3 +500,4 @@ def run_defaults(use_cache=True):  # todo: #remove-temp-defaults
 
 if __name__ == '__main__':
     cli()
+    #run_defaults()  # TODO temp
```
joeflack4, line 503: I see that Nico is trying to replicate this issue since I have been unable to ([comment](https://github.com/monarch-initiative/mondo-ingest/issues/406#issuecomment-1916227053)).

FYI Nico this `run_defaults()` is my alternative preferred way to debug / do development on this. You probably already know but I would have changed it back before undrafting this.

src/ontology/mondo-ingest.Makefile
```diff
@@ -437,6 +437,7 @@ tmp/merged.owl: tmp/mondo.owl mondo-ingest.owl tmp/mondo.sssom.ttl
 $(MAPPINGSDIR)/mondo-sources-all-lexical.sssom.tsv: $(SCRIPTSDIR)/match-mondo-sources-all-lexical.py tmp/merged.db $(MAPPINGSDIR)/rejected-mappings.tsv
 	rm -f $(MAPPINGSDIR)/mondo-sources-all-lexical.sssom.tsv
 	rm -f $(MAPPINGSDIR)/mondo-sources-all-lexical-2.sssom.tsv
+	pip install bioregistry
```
joeflack4, line 440: Hmm, missing bioregistry, or a specific version of it, I see.
```suggestion
	pip-bioregistry
```

I suggest this because I take it that it's not simply that `bioregistry` is missing, but maybe you want the latest version of it? In which case you want:
```make
pip-%:
```

@twhetzel The reason we have some `pip` install commands scattered throughout the makefile, and why they are for specific packages, rather than doing the a full install (e.g. `pip install -r python-requirements-unlocked.txt` or `poetry install`), is because for the particular piece of work we're working on, there's sometimes a new upgraded version of that package that has not yet made its way into the versioning for the latest release ODK yet. This is also partially alternatively solved by using the ODK dev container, which is what I think we're pretty much always using now.

@matentzn Are we using the ODK dev container for builds too now? I haven't seen an ODK release since September.

src/scripts/sync_subclassof.py
matentzn: I know - best not commit development related changes so others can just run the code after checking out the branch. just keep change in "staging"?

matentzn:
(I regularly update `docker pull obolibrary:odkfull:dev`)

src/ontology/mondo-ingest.Makefile
```diff
@@ -345,13 +345,17 @@ documentation: j2 $(ALL_DOCS) unmapped-terms-docs mapped-deprecated-terms-docs s
 build-mondo-ingest:
 	$(MAKE) refresh-imports exclusions-all slurp-all mappings matches \
 		mapped-deprecated-terms mapping-progress-report \
-		recreate-unmapped-components sync documentation
-	$(MAKE) prepare_release
```
matentzn, line 349: This $(MAKE) caused the imports to be run twice.. No idea why, but this fixes it for now.

src/scripts/sync_subclassof.py
```diff
@@ -309,7 +309,8 @@ def sync_subclassof(
     missing_ancestor_rels = rels_indirect_mondo_mondo.union(rels_direct_mondo_mondo).difference(ancestors_mondo_mondo)
     if missing_ancestor_rels:
         rels_for_printing =  [rel for rel in list(missing_ancestor_rels)[:5]]
-        logging.error(
+        # TODO: temp:
+        raise ValueError(
```
matentzn, line 313: @joeflack4 it seems you are right - when I reran it now, with -B, the error was not thrown. Must have been a syncing issue!

matentzn (approved):

matentzn:
@twhetzel This is now ready for review, and contains some minor fixes of the pipeline. We are getting close to stable.

After this and #417 is merged, I will run again. Thank you!

src/scripts/sync_subclassof.py
joeflack4: well, others don't typically need to run my draft code. also i sometimes use more than 1 computer so its good for me to push my state.  but it's not a big deal.

also you too can run it using run_defaults() for much easier debugging if you want ut to automatically pause whenever it encounters an exception, if your setup works like that (I know how to do this in pycharm, not in VS code)

twhetzel: Sounds like some good topics to discuss as we are a larger team now.

PR #409, opened 2024-01-30
Update omim_exclusions.tsv

sabrinatoro:
Added new term to the OMIM exclusion list


matentzn (approved):

twhetzel:
lgtm

@sabrinatoro @matentzn since these are gene symbols, is this something we want to scan and update automatically, ie to find and exclude genes?

matentzn:
> @sabrinatoro @matentzn since these are gene symbols, is this something we want to scan and update automatically, ie to find and exclude genes?

Good idea, would be easy enough. Could make issue.

sabrinatoro:
>since these are gene symbols, is this something we want to scan and update automatically, ie to find and exclude genes?
>Good idea, would be easy enough. Could make issue.

I don't know what OMIM file we ingest, but OMIM file indicates what entry corresponds to a gene. I thought we were already doing this (?)

twhetzel:
Good to know, I'll make a ticket to check where and why that did not work in this round. @sabrinatoro I am assuming based on the entries in `src/ontology/config/omim_exclusions.tsv` that you added these manually, is that correct?

sabrinatoro:
yes, I added the genes manually here, which suggests that we might not be doing what I mentioned above.

sabrinatoro:
can we merge this PR?

twhetzel:
PR - yeah, that's fine with me... I haven't sync'd up with @matentzn as to that process in this repo but will merge now.

Issue #405, opened 2024-01-28
SOP for development workflow


joeflack4:

matentzn:
Great initiative to get this started.

I would also suggest that even minor (non-functional) PRs (eg [this](https://github.com/monarch-initiative/mondo-ingest/pull/413) should get a review from now on.. We didnt do this so far, but at least until we have our development SOP sorted, maybe its a good way to ensure knowledge transfer (@twhetzel).

joeflack4:
@twhetzel I added a step (5) to the workflow for merging `develop` --> `main`.

The other day, we were merging in ICD11 and I had to run the build twice. But I've proposed a variation that I think is smarter; it will require that the build only be run once.

twhetzel:
To follow-up here at meeting, let's keep with the original plan to run the build on both the feature and develop branch, and then merge only the code changes into main.

joeflack4:
Just following up on some previous meeting notes; I don't think I jotted down well enough, but I think the crux of it was:
- As soon as a build on `main` passes (runs without fail), we can merge it without review, and just notify on Slack.

twhetzel:
Another note on merging `develop` into `main` is to do this when a new pipeline is complete, e.g. synonym sync, new resource like ICD11 Foundation, etc.

joeflack4:
@twhetzel For convention on _drafts_, do we want to undraft when (a) all requirements for the PR have been completed to the author's understanding, but the build has not yet been run / passed, or (b) build has also been run/passed.

I think I've been doing / prefer 'a', because it lets people know 'this is ready to review', even if we don't have word on the build yet.

twhetzel:
This development workflow should only apply to new features and managing feature branches since there is already a SOP in place for managing code in this repo related to the Mondo release cycle.

joeflack4:
OK that sounds good to dedupe this SOP. I think it came out around the same time but now the Mondo release cycle is very fleshed out.

Guess I'll change name to 'SOP for feature branch development workflow'. Will delete the `main` and `develop` parts from the OP now. Also linking release cycle to 'related' and linking our 'build stats' doc which is now part of the SOP.

joeflack4:
## Regarding the 'build' step
I have written: `Run build, e.g. sh run.sh make build-mondo-ingest -B`

So this is NP. However, I have actually been doing `build-mondo-ingest` without` -B` on fresh clones, which @matentzn and I had thought was equivalent.

[...]

PR #411, opened 2024-01-31
- Delete: Goal for mondo.sssom.config.yml

joeflack4:
## Changes
Delete: Goal for mondo.sssom.config.yml
It was previously getting from the 'mondo' repository, but the use cases / contents of the mondo-ingest config is significantly different, so it is now simply created statically.


src/ontology/mondo-ingest.Makefile
```diff
@@ -196,11 +196,6 @@ $(TMPDIR)/component-%.json: $(COMPONENTSDIR)/%.owl
 	$(ROBOT) convert -i $< -f json -o $@
 .PRECIOUS: $(TMPDIR)/component-%.json
 
-MONDO_SSSOM_CONFIG_URL=https://raw.githubusercontent.com/monarch-initiative/mondo/master/src/ontology/metadata/mondo.sssom.config.yml
-
-metadata/mondo.sssom.config.yml:
```
joeflack4, line 201: Delete

matentzn (commented):
Approved. But best to get explicit approvals even for small changes for now, see #405

PR #413, opened 2024-01-31
ODK yaml formatting

joeflack4:
## Updates
ODK yaml formatting
- Standardized codestyle for key/val and spaces around colons

### DIFF
--- a/src/ontology/mondo-ingest-odk.yaml
+++ b/src/ontology/mondo-ingest-odk.yaml
@@ -41,13 +41,13 @@ components:
       source: http://www.orphadata.org/data/ORDO/ordo_orphanet.owl
 robot_report:
   release_reports: False
-  fail_on : None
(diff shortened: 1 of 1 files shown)


src/ontology/mondo-ingest-odk.yaml
```diff
@@ -41,13 +41,13 @@ components:
       source: http://www.orphadata.org/data/ORDO/ordo_orphanet.owl
 robot_report:
   release_reports: False
-  fail_on : None
```
joeflack4, line 44: ## Remove space
For internal consistency, I removed the space in front of some of the `:` in this file.

I double checked on Google (no quick results) and [w/ ChatGPT](https://chat.openai.com/share/ed90a666-35ed-448e-9d32-7ef23118a675) as to whether this was syntactically the same and that this wouldn't break anything.

Issue #415, opened 2024-01-31
Stabelise serialisation of certain files


matentzn:
When rerunning frequently, it would be nice to avoid spurious diffs (diffs that are due to for example order changes in the lines). @joeflack4 already did a great job stabelising the serialisation of 90% of the files generated in Mondo ingest. I am listing here all the remaining ones to make it perfect.

- [ ] src/ontology/lexmatch/all_exact.robot.tsv
- [ ] docs/reports/mapped_deprecated.md
- [ ] docs/reports/unmapped.md
- [ ] src/ontology/lexmatch/README.md

Example:

<img width="846" alt="image" src="https://github.com/monarch-initiative/mondo-ingest/assets/7070631/dae16b5c-2ec7-4e84-8068-203cc0bb0584">

@twhetzel I am not automatically assigning anyone, as I want to make sure you can apply your own system of prioritisation/assignment.

joeflack4:
@twhetzel I wrote the code for (2) and (3) so I can easily commit to fixing those. I believe Harshad wrote the code for (1) and (4) but I should be able to figure out what's needed if I'm assigned to work on those as well.

PR #417, opened 2024-01-31
Update release files

matentzn:
Nothing surprising, can be merged without a big review.

See #415 to reduce spurious diffs.

This release was run based on #408 which is now also ready for review.


joeflack4 (approved):
This isn't the kind of PR that I can normally offer much review for, as it's more about the data and not about code. I looked through to see if there were any changes in output fields / columns -- I could at least help with that if so. But I saw no such changes. I did glance through anyway but nothing looks unusual to me.

PR #431, opened 2024-02-06
Updated files 2024.02.05

matentzn:
@twhetzel I hope this settles it; I ran the pipeline on a much lower spec machine (ubuntu, 4 core) and it also finished within 4 hours (not quite 2.5, like on my M3, but good enough compared to the 20+ hours it took before). 

```
### DEBUG STATS ###
Elapsed time: 3:41:08
Peak memory: 25709916 kb
```

From now on, I will run the pipeline only once per week, does this sound ok?

Note for review: I looked at every change and see no reason for concern.

🎉


matentzn:
PRs like this need to be merged immediately, else they have too many conflicts..

@twhetzel can we agree that just for updating the data files, we dont need a positive review (as there are no "changes" to the pipeline)? Else, I will have to chase every time a PR is open which creates an uncomfortable situation for me..

twhetzel:
I never saw this PR.

matentzn:
Oooo that is bad.. Somewhere your GitHub emails must be swallowed up, maybe by a spam filter of the institution.. Worth looking into!

twhetzel:
Nothing in spam. I was out sick Tuesday when this was submitted so this along with the plethora of weekend github emails and new ones, this was lost. If there is something time sensitive, like these pipeline reviews, ping me on Slack.

As far as the frequency to run the pipeline, I think that needs to be coordinated with Sabrina.

Issue #16, opened 2022-02-24
ICD11 ingest: grouping issue


matentzn:
## Overview
Ingest ICD11 into Mondo.

## Sub-task list
- [x] 1. Ingest ICD11 source  #434 @joeflack4 
- [x] 2. https://github.com/monarch-initiative/icd11/issues/10 / https://github.com/monarch-initiative/icd11/pull/12 @joeflack4 
- [x] 3. #140 @hrshdhgd 
- [x] 4. Close https://github.com/monarch-initiative/mondo/issues/1107 when all tasks above are complete

## 1. Ingest ICD11 source
Add ICD 11 foundation as a regular sync source in ODK / `mondo-ingest.Makefile`.

<details><summary>How to get data via REST API (defunct)</summary>
<p>
Defunct because we now have an `.owl` to work with.

From Can Celik (WHO):
> As Chris mentioned, we have a REST API for the ICD11 which gives access to the foundation component of the ICD-11 as well the linearizations (i.e. statistical classification). It contains additional functionality on top of providing full access to the classification. For example, the API contains search endpoints powered by our search engine which could be useful for mapping projects. The API supports JSON-LD, it is multilingual which contains 5 languages at the moment, and also provides access to earlier releases of the classification. More information on it could be found in our ICD-API Home Page https://icd.who.int/icdapi

And: 
> At this point, we don’t have a file that would be suitable for you. We have basic files with codes, titles, etc. but not something that includes more than that.
Reading everything won’t be a problem. The API is rather efficient. Also, you could easily deploy a local version if you prefer. With Docker, it’s just one command or several clicks to install it as a Windows or a Linux service.
We’ve just released new versions yesterday.

</p>
</details> 

## 2. ICD11::ORDO mappings
Get ahold of all ICD11 - ORDO mappings from a respectable source and derive an initial set of Mondo - ICD11 mappings from that (with the mapping set id as the source)

From Julie at Orphanet:
[...]

PR #412, opened 2024-01-31
ICD11 config & docs

joeflack4:
## Overview
I decided to make a smaller PR for the ICD11 ingest and try to merge just the configuration and docs before I even start writing the main ingest code.

---

Partially addresses:
- #16

<details><summary>Pre-merge checklist</summary>
<p>

- [x] Docs
   - `docs/` have been added/updated **OR**
   - No updates to the docs necessary after careful consideration.

- [x] QC
   - `sh run.sh make build-mondo-ingest` has been run on this branch (after `docker pull obolibrary/odkfull:dev), and no errors occurred **OR**
   -  No functional (code-related) changes to the pipeline are suggested, so no re-run is necessary.

- [x] Reviewed
   - Has been sufficiently reviewed by at least one review from a different team member of the Mondo Technical team.

</p>
</details> 

<details><summary>Changes</summary>
<p>

ICD11 config & docs
- Rename: ICD11 -> ICD11Foundation
- Add prefixes: icd11.foundation, icd11.schema, icd11.z
- Add: intensional exclusions TSV (currently empty)
- Add: SPARQL query for selecting all diseases

ICD11 config & docs
- Update: mondo-ingest-odk.yaml: New entry for ICD11 in 'components'
- Updates from running 'make update_repo':
  - modified: docs/odk-workflows/ManageDocumentation.md
  - modified: docs/odk-workflows/RepositoryFileStructure.md
  - modified: src/ontology/Makefile
  - modified: src/ontology/run.sh
  - new file: src/scripts/run-command.sh
  - modified: src/scripts/update_repo.sh

</p>
</details> 

## Sub-tasks
- [x] Add entry to `mondo-ingest-odk.yaml`
- [x] Update prefix map(s) in `metadata/*.yaml`, `metadata/mondo.sssom.config.yml`, and `config/prefixes.csv`
- [x] Add/update `mkdocs.yaml,` `docs/sources/*.md`, `docs/sources.md`, `docs/metrics/*.md`, and `docs/metrics.md`
- [x] https://github.com/monarch-initiative/mondo-ingest/issues/2 standard guidance (see: [thread](https://github.com/monarch-initiative/mondo-ingest/pull/412#discussion_r1485633424))
  - [x] `src/ontology/config/icd11_exclusions.tsv`?
  - [x] An exclude query for terms we don't want (sparql)?
  - [x] Any other considerations from that issue or otherwise?
- [x] Make sure pipeline passes

---

[...]

PR #434, opened 2024-02-08
ICD11 Ingest

joeflack4:
Partially addresses:
- #16 

## Overview
Adds source ICD11 to Mondo.
This will be deemed a success if `build-mondo-ingest` is run and ICD11 artefacts are present and look good (e.g. lexmatch with >1k rows).

## Updates
    ICD11 Ingest
    - Update: mondo-ingest.Makefile
      - Add: $(COMPONENTSDIR)/icd11.owl
    - Add: config/icd11foundation-property-map.sssom.tsv
    - Update: ICD11 docs
    - Update: config/properties.txt
    - Update: config/context.json

## Pre-merge checklist
- [x] Docs
- [x] QC
   - `sh run.sh make build-mondo-ingest` has been run on this branch (after `docker pull obolibrary/odkfull:dev), and no errors occurred **OR** https://github.com/monarch-initiative/mondo-ingest/pull/471
- [x] Account for any new packages
- [x] Reviewed

## Primary sub-tasks
- [x] Complete the goal
  - [x] SPARQL queries _(deferred)_
  - [x] Mapping TSV
- [x] [Analyze unmapped entities](https://github.com/monarch-initiative/mondo-ingest/pull/434#discussion_r1493431883)
- [x] Docs updates


src/ontology/mondo-ingest.Makefile
```diff
@@ -163,6 +163,23 @@ $(COMPONENTSDIR)/icd10who.owl: $(TMPDIR)/icd10who_relevant_signature.txt | compo
 		remove -T config/properties.txt --select complement --select properties --trim true \
 		annotate --ontology-iri $(URIBASE)/mondo/sources/icd10who.owl --version-iri $(URIBASE)/mondo/sources/$(TODAY)/icd10who.owl -o $@; fi
 
+# TODO: do
+# TODO: remove any unnecessary queries / lines here
+#  - fix_hgnc_mappings
+#  - maybe make a comment about other ones i thought to delete
+$(COMPONENTSDIR)/icd11.owl: $(TMPDIR)/icd11_relevant_signature.txt | component-download-icd11.owl
```
joeflack4, line 170: ## Discussion: Annotation props to remap
@matentzn Trish came up with the good idea of first simply listing out the predicates that we might need to remap.

@twhetzel also further refined my language on this. Is it correct to say that the predicates to map will all be annotation properties? If not, then I'm not capturing everything I might have to remap.

<details><summary>Full list of annotation props</summary>
<p>

browserUrl
codingHint
exclusion
foundationReference
fullySpecifiedName
icd10code
inclusion
isObsolote
longDefinition
narrowerTerm
note
releaseDate
releaseId
http://www.w3.org/2004/02/skos/core#altLabel
http://www.w3.org/2004/02/skos/core#broaderTransitive
http://www.w3.org/2004/02/skos/core#definition
http://www.w3.org/2004/02/skos/core#narrowerTransitive
http://www.w3.org/2004/02/skos/core#prefLabel

</p>
</details> 

<details><summary>Standard remappings already handled</summary>
<p>

We already have `config/property-map-2.sssom.tsv` which will be run. The entries intersection between this and what's in ICD11 is:
- http://www.w3.org/2004/02/skos/core#prefLabel --> http://www.w3.org/2000/01/rdf-schema#label
- http://www.w3.org/2004/02/skos/core#altLabel	--> oboInOwl:hasExactSynonym


</p>
</details> 

## Decisions to be made
<details><summary>Can probably ignore / filter out</summary>
<p>

- releaseDate
- releaseId
- browserUrl
- codingHint

</p>
</details> 

<details><summary>New cases to handle</summary>
<p>

#### `isObsolote` --> `owl:deprecated`
Note that they misspelled 'obsolete'.

#### `note` --> `rdfs:comment`

#### `fullySpecifiedName` --> `oboInOwl:hasExactSynonym`
Open to other suggestions. Note that they're also already using `skos:prefLabel` and `skos:altLabel`, which we will be remapping.

#### Mapping related
Props they have:
- http://www.w3.org/2004/02/skos/core#broaderTransitive
- http://www.w3.org/2004/02/skos/core#narrowerTransitive
- narrowerTerm

I'm not sure if these map to anything outside of ICD11. If they don't, should we simply ignore?
I imagine `narrowerTerm` is either `skos:narrowMatch` or `skos:broadMatch`.

The `skos:broaderTransitive` and `skos:narrowerTransitive` don't have any usages.

#### Definition related
Props they have:
- http://www.w3.org/2004/02/skos/core#definition
- longDefinition

Possible solutions:
a. Combine both and put into IAO:0000115
- Add subheaders?: "short definition" and "long definition"?
b. Put one into IAO:0000115, and include the other using its original predicate, or add it to rdfs:comment

#### Other
`foundationReference`
- include but also put inside rdfs:comment?

`icd10code`
- include but also put inside rdfs:comment?

`inclusion`
- take a look. include but also put inside rdfs:comment?

`exclusion`
- take a look. include but also put inside rdfs:comment?

</p>
</details> 

### Questions
1. In SPARQL, how do I map that which has no URI?
For some props, instead of URIs I'm seeing this `:PRED` syntax, e.g. `Declaration(AnnotationProperty(:isObsolote))` instead of `Declaration(AnnotationProperty(<http://www.w3.org/2004/02/skos/core#altLabel>))`

src/ontology/metadata/icd11.yml
```diff
@@ -0,0 +1,13 @@
+id: ICD11
+label: International Classification of Diseases 11th Revision
+prefix_map:
+  ICD11: http://purl.obolibrary.org/obo/mondo/mappings/unknown_prefix/ICD11/
```
matentzn, line 4: does not seem correct as per our conversation on ICD11 prefices

src/ontology/metadata/icd11.yml
```diff
@@ -0,0 +1,13 @@
...
+description: >
+  The International Classification of Diseases (ICD) provides a common language that allows health professionals to 
+  share standardized information across the world. The eleventh revision contains around 17 000 unique codes, more than 
+  120 000 codable terms and is now entirely digital.Feb 11, 2022
+  
+  This data source in particular is the ICD11 foundation, not one of its linearizations.
+homepage: https://icd.who.int/
+base_prefix_map:
+  ICD11: http://purl.obolibrary.org/obo/mondo/mappings/unknown_prefix/ICD11/
```
matentzn, line 13: See above

src/ontology/mondo-ingest.Makefile
matentzn: Very good effort to do this properly.

isObsolote --> owl:deprecated --> YES (important)
note --> rdfs:comment --> OK but we will never sync this. May help with debugging though

> The skos:broaderTransitive and skos:narrowerTransitive don't have any usages.

Well then, ditch.

> narrowerTerm

--> ignore for now and make separate issue with examples

> icd10code 

--> hasDbXref: "ICD10WHO:xyz" (needs sparql update)

> fullySpecifiedName --> oboInOwl:hasExactSynonym

Cool, but do in sparql, not remap, and add synonym type "fully qualified names"

> Definition related

Very important, and hard to answer. **I think separate issue**, for now mak skos:definition to IAO:0000115 but in issue, suggest what you want for the longDefinition.

> inclusion/exclusion

Make separate issue with examples. (PRs are not great for this)

> In SPARQL, how do I map that which has no URI?

In the prefixes there should be something like

```
Prefix(:=<http://purl.obolibrary.org/obo/mondo-ingest.owl#>)
```

Which denotes the base iri.

If there is no such thing, I dont know, I guess you will have to try.

matentzn (commented):
Shaping up!

src/ontology/metadata/icd11.yml
joeflack4: This is an artefact of this being branched off of #412, which is still WIP!

src/ontology/metadata/icd11.yml
joeflack4: This is an artefact of this being branched off of #412, which is still WIP!

src/ontology/mondo-ingest.Makefile
joeflack4: `icd10code`
Ah, yes I agree it should be `hasDbXref`. I actually read it as 11 and not 10.

_In SPARQL, how do I map that which has no URI?_
Ahh, I see your answer. Excellent; I found: `Prefix(:=<http://id.who.int/icd/schema/>)`

"Will do!" and "Good to hear!" regarding all the other topics!

docs/odk-workflows/RepositoryFileStructure.md
```diff
@@ -34,6 +34,7 @@ These are the components in MONDO-INGEST
 | gard.owl | https://github.com/monarch-initiative/gard/releases/latest/download/gard.owl |
 | icd10cm.owl | https://data.bioontology.org/ontologies/ICD10CM/submissions/23/download?apikey=8b5b7825-538d-40e0-9e9e-5ab9274a9aeb |
 | icd10who.owl | https://github.com/monarch-initiative/icd10who/releases/latest/download/icd10who.ttl |
+| icd11.owl | https://icd11files.blob.core.windows.net/tmp/whofic-2023-04-08.owl.gz |
```
joeflack4, line 37: ## Docs
_[Nico wrote](https://github.com/monarch-initiative/mondo-ingest/pull/412#discussion_r1485633597)_:
> add what you know and we will iterate (where we got the file from, for instance, and maybe a link to an external page about what the ICD11 foundation is)
- [ ] @joeflack4 By the time I've undrafted PR, add anything I can think of to `docs/sources/icd11.md`
- [ ] @joeflack4 Perhaps also add: where we got the file, link to what ICD11 foundation is

src/ontology/mondo-ingest.Makefile
joeflack4, line 170: ## Analyze unmapped entities
- [ ] @joeflack4

I wonder if the easiest way to do this is by creating a SSSOM first when the ingest is working correctly.

_[Nico wrote](https://github.com/monarch-initiative/mondo-ingest/pull/412#discussion_r1485632153)_:
> It would be good to get a sense of which entities are not mapped after these 7 namespaces are applied.

<p>

```
rdf: http://www.w3.org/1999/02/22-rdf-syntax-ns#
rdfs: http://www.w3.org/2000/01/rdf-schema#
owl: http://www.w3.org/2002/07/owl#
skos: http://www.w3.org/2004/02/skos/core#
icd11.foundation: http://id.who.int/icd/entity/
icd11.schema: http://id.who.int/icd/schema/
icd11.z: http://who.int/icd#Z_
```

</p>
</details>

src/ontology/metadata/icd11.yml
joeflack4: Resolved now following `rebase`.

src/ontology/mondo-ingest.Makefile
```diff
@@ -163,6 +163,22 @@ $(COMPONENTSDIR)/icd10who.owl: $(TMPDIR)/icd10who_relevant_signature.txt | compo
...
```
joeflack4, line 170: ## Comparison to other components goals
### Removals
1. **`rename --mappings config/property-map-2.sssom.tsv `**: None of the annotation props in this file appear in ICD11. Honestly I think this may be the case for some other sources as well, even though this step appears in those goals (no harm because of `--allow-missing-entities true`)

### Kept
- **`rename --mappings config/property-map-1.sssom.tsv `**: 2/3 (`prefLabel`, `altLabel` of the annotation props appear in ICD11. However, even after the removal, the `<owl:AnnotationProperty>` declaration still remains. IDK if this is an oversight on `robot`'s part

joeflack4, line 170: ## Source has `isObsolete` and `owl:deprecated`
@matentzn @twhetzel I initially found `isObsolete` and was going to rename that to `owl:deprecated`. However, I now see that the source has both. Not sure what I intend to do yet. Perhaps there is some nuance difference between these two. I'm not sure if we care about that nuance and should do some retention of both of these properties, or if I should rename them all to `owl:deprecated`.

Perhaps one thing I can do is an analysis of the cases where a class is both `isObsolete` and `owl:deprecated`, and also look at classes that only have one of these properties.

src/ontology/mondo-ingest.Makefile
```diff
@@ -163,6 +163,21 @@ $(COMPONENTSDIR)/icd10who.owl: $(TMPDIR)/icd10who_relevant_signature.txt | compo
...
+#  - fix_hgnc_mappings
+#  - maybe make a comment about other ones i thought to delete
+$(COMPONENTSDIR)/icd11foundation.owl: $(TMPDIR)/icd11foundation_relevant_signature.txt | component-download-icd11foundation.owl
+	if [ $(COMP) = true ] ; then $(ROBOT) remove -i $(TMPDIR)/component-download-icd11foundation.owl.owl --select imports \
+		rename --mappings config/property-map-1.sssom.tsv --allow-missing-entities true \
+		remove -T $(TMPDIR)/icd11foundation_relevant_signature.txt --select complement --select "classes individuals" --trim false \
+		remove -T $(TMPDIR)/icd11foundation_relevant_signature.txt --select individuals \
+		query \
+			--update ../sparql/fix-labels-with-brackets.ru \
+			--update ../sparql/fix_complex_reification.ru \
```
joeflack4, line 177: ## Question: `fix_complex_reification.ru`
Does this file only apply to ORDO?

<details><summary>Details & suggestions</summary>
<p>

I've since removed this line from the `icd11foundation` goal, but it's still on several other goals.

I ask because it at the end, it has:
```sparql
  FILTER (isIRI(?entity) && (regex(str(?entity), "http://www.orpha.net/ORDO/")))
  BIND(IF(STRSTARTS(str(?precision_value), "NTBT"), skos:narrowMatch, IF(STRSTARTS(str(?precision_value), "BTNT"), skos:broadMatch, oboInOwl:hasDbXref)) as ?property)
```

### Suggestions
1. If it indeed only applies to ORDO, then I suggest we remove this line from all the goals except for ORDO, as done in PR with similar situation: #459.
2. If that's the case, I suggest we also prepend `ordo_` to the filename (like I've suggest in similar situation https://github.com/monarch-initiative/mondo-ingest/pull/459#discussion_r1510414581).

</p>
</details>

src/ontology/config/icd11foundation-property-map.sssom.tsv
```diff
@@ -0,0 +1,4 @@
+subject_id	object_id
+http://purl.org/dc/elements/1.1/source	oboInOwl:source
+http://id.who.int/icd/schema/isObsolote    owl:deprecated
+http://id.who.int/icd/schema/note    rdfs:comment
```
joeflack4, line 4: ## Added `icd11foundation-property-map.sssom.tsv`
**Filename**: The only 2 other such files are called `property-map-1.sssom.tsv` and `property-map-2.sssom.tsv`. But I felt that since this is only being used for 1 source, it would be good to prepend the source name.

**Context**: property mappings decision thread

src/ontology/mondo-ingest.Makefile
```diff
@@ -163,6 +163,18 @@ $(COMPONENTSDIR)/icd10who.owl: $(TMPDIR)/icd10who_relevant_signature.txt | compo
 		remove -T config/properties.txt --select complement --select properties --trim true \
 		annotate --ontology-iri $(URIBASE)/mondo/sources/icd10who.owl --version-iri $(URIBASE)/mondo/sources/$(TODAY)/icd10who.owl -o $@; fi
 
+# TODO: complete
+$(COMPONENTSDIR)/icd11foundation.owl: $(TMPDIR)/icd11foundation_relevant_signature.txt | component-download-icd11foundation.owl
```
joeflack4, line 167: ## `foundationReference`
- [ ] @matentzn or @twhetzel  advise
<p>

In the [property discussion thread](https://github.com/monarch-initiative/mondo-ingest/pull/434#discussion_r1483728680), I wrote:

> include but also put inside `rdfs:comment`?

**What to do?**
What should I do in this case? Should I just ignore it? Should I integrate it into Mondo somehow, e.g. by including in `rdfs:comment`?

**Analysis?**
Perhaps I need to show some cases of how this is used / what it looks like? Is this plan correct?:
- [ ] Show an example @joeflack4 
- [ ] Then, Nico/Trish will advise

</p>
</details>

joeflack4, line 167: ## `skos:broaderTransitive` and `skos:narrowerTransitive`
- [ ] @matentzn advise
<p>

**Context from [property discussion thread](https://github.com/monarch-initiative/mondo-ingest/pull/434#discussion_r1483728680)**
Joe:
> The skos:broaderTransitive and skos:narrowerTransitive don't have any usages.

Nico:
> Well then, ditch.

**Question**
What do you mean by ditch?
It sounds like you want me to remove something, but there are no usages to remove. Do you want me to remove the declarations? But I don't know if that's worth the time / trouble.
Maybe by "ditch" you meant to just not worry about these?

</p>
</details>

joeflack4, line 167: ## Which annotation prop issues in scope for this PR?
- [ ] @matentzn / @twhetzel advise

I created these 3 issues, and added them as checkboxes to the OP, and I set the issue priority as well as added my comment to the side in accordance to how important / timely it is / whether I think it might be in scope for this PR. Personally I don't mind just leaving all of these until after the PR, though. Thoughts?
- #461 (possibly / probably in scope)
- #462 (possibly in scope)
- #463 (possibly / probably not in scope)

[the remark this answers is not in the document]
matentzn: The synchronisation pipeline (you did this for subclass of and deprecation) deliberatly picks and chooses content that we want synchronised. I would prefer you to add a robot remove step that gets rid of all of these properties as part of the ingest to keep the component file smaller.

matentzn: Examples?

matentzn: robot `remove --term skos:broaderTransitive --term skos:narrowerTransitive`

matentzn: Thanks, great move. I think you can leave this to a subsequent PR.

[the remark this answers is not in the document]
joeflack4: Synchronisation, OK thought so. So when we integrate a source, we do a mapping process. And then there's synchronisation. I just want to give myself clarity; these two processes cover like 100% of our source ingestion, you would say?

- [ ] `robot remove` these 4 properties @joeflack4

joeflack4, line 167: ## Question: Why `xmlns:schema` created?
- [ ] @matentzn or @twhetzel

I noticed that `components/icd11foundation.yml` has this in its namespace: `xmlns:schema="http://id.who.int/icd/schema/">`. I'm curious as to why this namespace was assigned when (i) I didn't instruct `robot` to do so, (ii) it didn't decide to automatically create any other namespaces (e.g why did it create this namespace, but not `xmlns:entity="http://id.who.int/icd/entity/">`?

matentzn: Are you sure that the original file from WHO does not have that NS declaration?

joeflack4: Ah, good idea. Yes it does! It's the one that was set to `:`: `Prefix(:=<http://id.who.int/icd/schema/>)`

src/ontology/mondo-ingest.Makefile
joeflack4: @matentzn I'm just coming back to to this after a few weeks. Can you clarify again for me what you mean by the above quote? What is my action item?

src/ontology/mondo-ingest.Makefile
joeflack4: Reviewed with Trish today.

joeflack4: Can maybe write a SPARQL query or use OAK for the usage of each, or OAK. And see if there's an interesection.

src/ontology/config/properties.txt
```diff
@@ -19,6 +19,8 @@ http://www.w3.org/2004/02/skos/core#narrowMatch
 http://www.w3.org/2004/02/skos/core#relatedMatch
 http://www.w3.org/2004/02/skos/core#exactMatch
 http://www.w3.org/2004/02/skos/core#closeMatch
+rdfs:comment
 rdfs:label
 rdfs:seeAlso
 owl:deprecated
+dcterms:description
```
joeflack4, line 26: ## Updated `config/properties.txt`
There are some [properties we decided to map to](https://github.com/monarch-initiative/mondo-ingest/pull/434#discussion_r1483728680) that aren't included in this file, so I've added them.
[_Related: Added `icd11foundation-property-map.sssom.tsv`_](https://github.com/monarch-initiative/mondo-ingest/pull/434#discussion_r1510429828)

Based on my understanding, any properties not present in this file will be removed.

src/ontology/mondo-ingest.Makefile
```diff
@@ -163,6 +163,19 @@ $(COMPONENTSDIR)/icd10who.owl: $(TMPDIR)/icd10who_relevant_signature.txt | compo
...
+	if [ $(COMP) = true ] ; then $(ROBOT) remove -i $(TMPDIR)/component-download-icd11foundation.owl.owl --select imports \
+		rename --mappings config/property-map-1.sssom.tsv --allow-missing-entities true \
+		rename --mappings config/icd11foundation-property-map.sssom.tsv \
+		remove -T $(TMPDIR)/icd11foundation_relevant_signature.txt --select complement --select "classes individuals" --trim false \
+		remove -T $(TMPDIR)/icd11foundation_relevant_signature.txt --select individuals \
+		remove --term http://id.who.int/icd/schema/narrowerTerm --select "annotation-properties" \
```
joeflack4, line 173: ## `remove` `narrowerTerm
Per:
- #461

src/ontology/mondo-ingest.Makefile
joeflack4: Resolving as out of scope for this PR. See: 
- #469

src/ontology/mondo-ingest.Makefile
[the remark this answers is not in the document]
joeflack4: - [ ] Advise @matentzn @twhetzel 
In addition to just clarifying the above question, I want to know if it really matters to remove `releaseId` and ` releaseDate`. Because they only exist on the `Ontology` itself, not any class. And the command I ran to remove them didn't work, so I undid it.

<p>

I ran: `remove --term http://id.who.int/icd/schema/narrowerTerm --term http://id.who.int/icd/schema/browserUrl --term http://id.who.int/icd/schema/codingHint --term http://id.who.int/icd/schema/releaseDate --term http://id.who.int/icd/schema/releaseId --select "annotation-properties"`

For `narrowerTerm`,  `browserUrl`, and `codingHint`, it removed usages and declarations as I intended. But for `releaseId` and `releaseDate`, it only removed the declarations.

`cat tmp/component-download-icd11foundation.owl.owl | grep releaseDate`
```owl
        <schema:releaseDate>2023-04-09</schema:releaseDate>
    <!-- http://id.who.int/icd/schema/releaseDate -->
    <owl:AnnotationProperty rdf:about="http://id.who.int/icd/schema/releaseDate"/>
```

`cat tmp/component-download-icd11foundation.owl.owl | grep releaseId`
```owl
        <schema:releaseId xml:lang="en">beta</schema:releaseId>
    <!-- http://id.who.int/icd/schema/releaseId -->
    <owl:AnnotationProperty rdf:about="http://id.who.int/icd/schema/releaseId"/>
```

As far as I see it, my choices are:
a. Actually just leave these in, they're harmless and maybe even nice to keep.
b. Stay with [robot remove](https://robot.obolibrary.org/remove#selectors)'ing them, but how? I can try `--select "properties"` instead of `--select "annotation-properties"`.

</p>
</details>

src/sparql/icd11foundation-icd10-xrefs.ru
```diff
@@ -0,0 +1,17 @@
...
+prefix owl: <http://www.w3.org/2002/07/owl#>
+prefix oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
+prefix sssom: <https://w3id.org/sssom/>
+
+
+DELETE {
+  ?cls http://id.who.int/icd/schema/icd10code ?xref .
+} INSERT {
+  # TODO: add prefix declaration for ICD10WHO or wrap in URI() around CONCAT()
+  ?cls oboInOwl:hasDbXref ICD10WHO?xref CONCAT("ICD10WHO", REPLACE(?xref, "NEED_EXAMPLE_SHOWING_URI_STEM", ""))
```
joeflack4, line 15: ## `icd10code`
- [ ] Advise @matentzn / @twhetzel

I started writing a SPARQL query to remap `icd10code` annotation properties to `oboInOwl:hasDbXref ICD10WHO:<CODE>`. However I just realized that there exists only a property declaration for `icd10code`, no usages.

Decisions:
1. What to do with the annotation prop declaration?
  a. remove
  b. keep
2. What to do with my existing SPARQL file?
  a. Delete it since it's not needed
  b. Keep it for the future in case we need it; keep it blank
  c. Keep it for the future in case we need it; leave it as is: unfinished w/ TODO's

src/ontology/config/properties.txt
matentzn: If the command is redundant, remove to save CPU cycles. I think we should prefer using config/properties.txt to have a common solution for all ingests, rather than configuring this on an ingest by ingest basis.

src/ontology/mondo-ingest.Makefile
joeflack4: - [ ] Find: Are there any entities in the ontology that don't fall within these 7 namespaces. @joeflack4 

What is the quickest way to find such an entity?
One hacky way, 5 min without pain: Get all the entities in the file, OAK --terms, etc. Do a regex remove from each of these namespaces. or file sort it like as a TSV and look through.

src/ontology/config/properties.txt
joeflack4: - [ ] @joeflack4 
Sounds good. Verify that when I convert the file with and without the `robot remove ... --annotation-properties` line, it is the same effect as just leaving the `properties.txt` line by itself. If that's the case, remove the `robot remove` one.

src/ontology/mondo-ingest.Makefile
joeflack4: Marking resolved as the only stuff left here will be done in different PR.

src/ontology/config/properties.txt
```diff
@@ -20,6 +20,8 @@ http://www.w3.org/2004/02/skos/core#narrowMatch
 http://www.w3.org/2004/02/skos/core#relatedMatch
 http://www.w3.org/2004/02/skos/core#exactMatch
 http://www.w3.org/2004/02/skos/core#closeMatch
+rdfs:comment
 rdfs:label
 rdfs:seeAlso
 owl:deprecated
+http://purl.org/dc/terms/description
```
joeflack4, line 27: ## `xmlns:terms="http://purl.org/dc/terms/"`
- [ ] Advise @matentzn or @twhetzel

I originally added an entry `dcterms:description` to `config/properties.txt`, but it was not a recognized namespace, so when I converted `longDefinition` to this, it got removed. I changed to `http://purl.org/dc/terms/description`, but I got an unexpected namespace `terms` added.

I actually had the same thing happen for an OMOP ingest I was doing for TIMS. I was doing `robot template ... --prefix "OMOP: https://athena.ohdsi.org/search-terms/terms/"`, and it ignored my namespace and did `xmnls:terms` instead.

Does this bother you at all or cause any problems for us? Should I keep it as is? Make an issue to handle this later? Or handle this before merging this PR?

Any ideas what might be causing the issue?
I know that we just had the same problem with it creating `xmlns:schema="http://id.who.int/icd/schema/"`. When that happened, you asked me to check the original `.owl` that we got from WHO and see if that namespace was already there, and indeed it was. However, I just checked it and this 'terms' namespace is not there.

src/ontology/config/icd11foundation-property-map.sssom.tsv
```diff
@@ -0,0 +1,5 @@
+subject_id	object_id
+http://id.who.int/icd/schema/isObsolote	owl:deprecated
+http://id.who.int/icd/schema/longDefinition	http://purl.org/dc/terms/description
+http://id.who.int/icd/schema/note	rdfs:comment
+skos:definition	IAO:0000115
```
joeflack4, line 5: ## `obo:IAO_0000115`
- [ ] Advise @matentzn  or @twhetzel 
I'm getting an `<obo:IAO_0000115 xml:lang="en">` tag. I was expecting `<IAO:0000115 xml:lang="en">`. I checked our other components though, and I see that the same thing is also happening for them.

Am I correct that this is unexpected? Is this potentially causing any problems for us? Should I create an issue or do anything in this PR?

src/ontology/config/properties.txt
matentzn: If it's just namespace declarations don't worry about it. In RDF they are purely cosmetic. Now in semsql all of this may matter more; but then you simple extend the prefixmap for that correctly. So yes, you can leave it as is.

src/ontology/config/icd11foundation-property-map.sssom.tsv
matentzn: No problem you can leave this. In RDF prefixes are syntactic sugar.

src/ontology/config/context.json
```diff
@@ -31,6 +31,8 @@
 		"NCBITaxon": "http://purl.obolibrary.org/obo/NCBITaxon_",
 		"ICD10CM": "http://purl.bioontology.org/ontology/ICD10CM/",
 		"ICD10WHO": "http://apps.who.int/classifications/icd10/browse/2010/en#/",
+		"icd11.foundation": "http://id.who.int/icd/entity/",
```
joeflack4, line 34: ## FYI: `config/context.json`
- [ ] Review & mark resolved @matentzn @twhetzel 

I needed to update this file as well. I found that docs about it were missing, so I made a PR for that:
- #470

I think this is a file that probably should be programmatically generated. I updated:
- https://github.com/monarch-initiative/mondolib/issues/8

src/ontology/mondo-ingest.Makefile
joeflack4: @matentzn LGTM:
I looked at this today. I ran `runoak -i INPUT terms > OUTPUT.csv`, then I just opened in Excel and sorted the column and looked at the unique namespaces that way.

- `icd11.foundation`: ~94k
- `icd11.schema`: 82
- `icd11.z:RetiredClasses`: 1
- `obo:mondo-ingest/components/icd11foundation.owl`: 1 _(seems weird but I'm not worried about it. I also don't see anything like this in `icd11foundation.owl`)_
- `skos`: 5 _(IDK why this appears here and e.g. `rdfs` doesn't, but I'm not worried about this either)_

docs/odk-workflows/RepositoryFileStructure.md
joeflack4: @matentzn @twhetzel I just added the following.

> **Comments about this source:**  
> _Data source_
> _Original source URL_: https://icd11files.blob.core.windows.net/tmp/whofic-2023-04-08.owl.gz
> 
> _Preprocessing_
> In the [monarch-initiative/icd11](https://github.com/monarch-initiative/icd11) repo, We remove unicode characters and then remove equivalent class statements as discussed below. 
> 
> _Equivalent classes_  
> We remove all equivalent class statements as they are not unique and result in unintended node merges. For example  `icd11.foundation:2000662282` (_Occupant of pick-up truck or van injured in collision with car, pick-up truck or van: person on outside of vehicle injured in traffic accident_) has the same exact equivalent concept expression as `icd11.foundation:1279712844` (_Occupant of pick-up truck or van injured in collision with two- or three- wheeled motor vehicle: person on outside of vehicle injured in traffic accident_).

Marking resolved.

src/ontology/config/properties.txt
joeflack4: @matentzn @twhetzel Indeed, effect was the same! Marking resolved.

src/ontology/mondo-ingest.Makefile
```diff
@@ -159,6 +159,17 @@ $(COMPONENTSDIR)/icd10who.owl: $(TMPDIR)/icd10who_relevant_signature.txt | compo
 		remove -T config/properties.txt --select complement --select properties --trim true \
 		annotate --ontology-iri $(URIBASE)/mondo/sources/icd10who.owl --version-iri $(URIBASE)/mondo/sources/$(TODAY)/icd10who.owl -o $@; fi
 
+$(COMPONENTSDIR)/icd11foundation.owl: $(TMPDIR)/icd11foundation_relevant_signature.txt | component-download-icd11foundation.owl
```
joeflack4, line 162: ## `build-mondo-ingest` _@finally_
- [ ] Run `build-mondo-ingest` and considered passing if [lexmatch table](https://github.com/monarch-initiative/mondo-ingest/blob/main/src/ontology/lexmatch/unmapped_doid_lex_exact.tsv) has more than 1k rows. @joeflack4

src/ontology/mondo-ingest.Makefile
joeflack4: @matentzn @twhetzel Alright, so it looks like `foundationReference` works something like this. It exists only on annotation axioms for `annotatedProperty` `icd11.foundation:inclusion` and `icd11.foundation:exclusion`.

<details><summary>How I determined it was annotating only these 2</summary>
<p>

1. `cat tmp/component-download-icd11foundation.owl.owl | grep foundationReference -B 2 > ~/Desktop/foundationRef.txt`
2. `cat ~/Desktop/foundationRef.txt| grep annotatedProperty > ~/Desktop/foundationrefprops.txt`
3. Looked at the unique values.

</p>
</details> 

As I wrote in #462:
> I'm not 100% sure, but at face these appear to me to be a way, to add additional modifications beyond a normal subclass relationship. You mark something a subclass of another, but then it might inherit some other relationships which then you can mark with `exclusion`. And then perhaps you can mark `inclusion` for some other classes that do not have a subclass relationship.

<details><summary>Example usage in OWL</summary>
<p>

For the first class, I left all properties just for reference. It actually had 2 axioms annotating its exclusions. I left one out. The axiom that I left in actually refers to the 2nd class I've shown an example for. It also has 1 exclusion and an axiom annotation on it, referenced by the 3rd class I've put as an example, which itself has 4 exclusions, each with an annotation axiom.

```owl
    <owl:Class rdf:about="http://id.who.int/icd/entity/1000337196">
        <rdfs:subClassOf rdf:resource="http://id.who.int/icd/entity/1632259883"/>
        <schema:browserUrl rdf:resource="https://icd.who.int/dev11/f/en#/http%3A%2F%2Fid.who.int%2Ficd%2Fentity%2F1000337196"/>
        <schema:exclusion xml:lang="en">Catatonia</schema:exclusion>
        <schema:exclusion>Delirium</schema:exclusion>
        <schema:inclusion xml:lang="en">Semicoma</schema:inclusion>
        <skos:altLabel xml:lang="en">Semicoma</skos:altLabel>
        <skos:altLabel xml:lang="en">stuporous</skos:altLabel>
        <skos:definition xml:lang="en">Total or nearly total lack of spontaneous movement and marked decrease in reactivity to environment.</skos:definition>
        <skos:prefLabel xml:lang="en">Stupor</skos:prefLabel>
    </owl:Class>
    <owl:Axiom>
        <owl:annotatedSource rdf:resource="http://id.who.int/icd/entity/1000337196"/>
        <owl:annotatedProperty rdf:resource="http://id.who.int/icd/schema/exclusion"/>
        <owl:annotatedTarget xml:lang="en">Catatonia</owl:annotatedTarget>
        <schema:foundationReference rdf:resource="http://id.who.int/icd/entity/486722075"/>
    </owl:Axiom>
    ...



    <owl:Class rdf:about="http://id.who.int/icd/entity/486722075">
        ...
        <schema:exclusion>Harmful effects of drugs, medicaments or biological substances, not elsewhere classified</schema:exclusion>
        <skos:prefLabel xml:lang="en">Catatonia</skos:prefLabel>
    </owl:Class>
    <owl:Axiom>
        <owl:annotatedSource rdf:resource="http://id.who.int/icd/entity/486722075"/>
        <owl:annotatedProperty rdf:resource="http://id.who.int/icd/schema/exclusion"/>
        <owl:annotatedTarget>Harmful effects of drugs, medicaments or biological substances, not elsewhere classified</owl:annotatedTarget>
        <schema:foundationReference rdf:resource="http://id.who.int/icd/entity/1868408442"/>
    </owl:Axiom>



    <owl:Class rdf:about="http://id.who.int/icd/entity/1868408442">
        ...
        <schema:exclusion xml:lang="en">Alcohol intoxication</schema:exclusion>
        <schema:exclusion>Allergic or hypersensitivity conditions</schema:exclusion>
        <schema:exclusion>Disorders due to substance use or addictive behaviours</schema:exclusion>
        <schema:exclusion>Reactions or intoxications due to drugs administered to fetus or newborn</schema:exclusion>
        <schema:inclusion xml:lang="en">overdose of these substances</schema:inclusion>
        <schema:inclusion xml:lang="en">wrong substance given or taken in error</schema:inclusion>
        ...
        <skos:prefLabel xml:lang="en">Harmful effects of drugs, medicaments or biological substances, not elsewhere classified</skos:prefLabel>
    </owl:Class>
    <owl:Axiom>
        <owl:annotatedSource rdf:resource="http://id.who.int/icd/entity/1868408442"/>
        <owl:annotatedProperty rdf:resource="http://id.who.int/icd/schema/exclusion"/>
        <owl:annotatedTarget xml:lang="en">Alcohol intoxication</owl:annotatedTarget>
        <schema:foundationReference rdf:resource="http://id.who.int/icd/entity/1339202943"/>
    </owl:Axiom>
    <owl:Axiom>
        <owl:annotatedSource rdf:resource="http://id.who.int/icd/entity/1868408442"/>
        <owl:annotatedProperty rdf:resource="http://id.who.int/icd/schema/exclusion"/>
        <owl:annotatedTarget>Allergic or hypersensitivity conditions</owl:annotatedTarget>
        <schema:foundationReference rdf:resource="http://id.who.int/icd/entity/642618805"/>
    </owl:Axiom>
    <owl:Axiom>
        <owl:annotatedSource rdf:resource="http://id.who.int/icd/entity/1868408442"/>
        <owl:annotatedProperty rdf:resource="http://id.who.int/icd/schema/exclusion"/>
        <owl:annotatedTarget>Disorders due to substance use or addictive behaviours</owl:annotatedTarget>
        <schema:foundationReference rdf:resource="http://id.who.int/icd/entity/1602669465"/>
    </owl:Axiom>
    <owl:Axiom>
        <owl:annotatedSource rdf:resource="http://id.who.int/icd/entity/1868408442"/>
        <owl:annotatedProperty rdf:resource="http://id.who.int/icd/schema/exclusion"/>
        <owl:annotatedTarget>Reactions or intoxications due to drugs administered to fetus or newborn</owl:annotatedTarget>
        <schema:foundationReference rdf:resource="http://id.who.int/icd/entity/142627676"/>
    </owl:Axiom>
```

</p>
</details> 

So, `inclusion` and `exclusion` hold labels for classes. The URI to the foundation class (`foundationReference`) is then annotated on those properties.

Why not just make `inclusion` and `exclusion` point to URIs for the classes instead of their labels? I suppose it probably has something to do with linearization. I assume new class IDs get generated at linearization-time, but the labels remain static.

src/ontology/mondo-ingest.Makefile
joeflack4: Been running for 3 hours so far. I'll check on it later tonight or tomorrow.

joeflack4: Ok, it just finished. Unfortunately there were no `lexmatch/unmapped_icd11foundation*.txv` generated. I'll look into it this week, hopefully tomorrow. I am supposing there is some configuration missing somewhere. I wonder if there might be a hardcoding of ontology names in the lexmatch scripts.

hrshdhgd: > Fix issues where lexmatch table is not getting generated

The `components` dir does not seem to have any `icd11.owl` file , that may be why there are no lexmatch files for icd11

joeflack4: Ah yeah, it should be `icd11foundation`. We changed the name.

src/ontology/metadata/mondo.sssom.config.yml
```diff
@@ -367,6 +367,7 @@ subject_prefixes:
   - EFO
   - ICD10CM
   - ICD10WHO
+  - icd11.foundation
```
joeflack4, line 370: ## `mondo.sssom.config.yml`: `subject_prefixes`
I'm doing another quick check and re-running the build, but I added this missing namespace to `subject_prefixes`, ran a lexmatch script, and got what I expected.

I've added this as a step to `docs/developer/add-new-source.md`.

joeflack4, line 370: ## `source_metadata`
- [ ] Any need, benefit, or future desire to add something here for ICD11? @matentzn @hrshdhgd 

There's also a `source_metadata` field in `mondo.sssom.config.yml` with 9 entries in it.

<details><summary>Example</summary>
<p>

```yml
  - from: MONDO
    to: OMIM
    metadata:
      object_source: https://www.omim.org/
      object_source_version: http://www.orphadata.org/data/ORDO/ordo_orphanet_3.1.owl
      mapping_date: "2022-02-11"
      comment: "The latest OMIM is integrated with every Mondo release."
```

</p>
</details> 

But I don't think this is strictly necessary for ICD11 at this time. But I don't know for sure what the benefits of adding an entry here would be. I looked around for references for some of these fields, e.g. `object_source_version`, in `mondo-ingest`, but didn't find any outside of this `mondo.sssom.config.yml`.

src/scripts/lexmatch-sssom-compare.py
```diff
@@ -173,11 +173,14 @@ def extract_unmapped_matches(input: str, matches: TextIO, output_dir: str, summa
     ont_df_list = []
 
     for _, ont in enumerate(input):
+        # Map ontology filenames to prefixes
         ont2 = ont.upper()
         if ont == "omim":
             ont2 = "|".join((["OMIM", "OMIMPS"]))
         elif ont == "ordo":
             ont2 = "|".join((["ORDO", "Orphanet"]))
+        elif ont == "icd11foundation":
```
joeflack4, line 182: ## ICD11 entry in `lexmatch-sssom-compare.py`
- [ ] @hrshdhgd @matentzn Review suggestions / etc
This was location 2 of 2 I updated to get ICD11 working in lexmatch. 

Nico mentioned that I should have >1k entries in lexmatch for this to be a success. I haven't ran `build-mondo-ingest` yet after this fix, but just from running the script, I got 7k files. I think Nico mostly means these files, both of which do have >1k entries each:
- `unmapped_icd11foundation_lex.tsv`
- `unmapped_icd11foundation_lex_exact.tsv`

I added documentation:
> ### 2.5. `lexmatch-sssom-compare.py`
> There is a section of branching logic with a comment "Map ontology filenames to prefixes". Add an entry there.

### Suggestions
I think probably branching logic inside of `lexmatch-sssom-compare.py` is not the best location for this kind of configuration. Thinking maybe moving to:
a. Some existing configuration file
  - a.a. `metadata/ONTOLOGY.yml`: My initial thought is add a `component_filename` field to `metadata/ONTOLOGY.yml`. And then we would have lexmatch simply check that, and then pull the `base_prefix_map`. Then we would pull `subject_prefixes` from there, although to do that, we'd have to go with [my suggestion](https://github.com/monarch-initiative/mondo-ingest/pull/434/files#r1531114745) of moving `subject_prefixes` from `mondo.config.sssom.yml`.
b. a new `lexmatch_config` (`.py`, `.json`, or `.yml`)

### FYIs
@hrshdhgd Was thinking about doing this in a separate PR but it is minor / we don't have to keep it. I also changed `combined_df = pd.concat(ont_df_list)` to `combined_df = pd.concat(ont_df_list) if ont_df_list else pd.DataFrame()`. This is because I got an error because I was running the script directly with just `icd11.foundation` as the only component, and since it didn't detect it, it got here and there was nothing to concat.

src/ontology/mondo-ingest.Makefile
joeflack4, line 162: ## No slurp outputs
Still trying to figure this out. When I run `migrate.py` manually, I get slurp outputs for `icd11foundation` no problem. However when I run `slurp-icd11foundation`, it gets `rm`'ed at the end, and I can't figure out what's causing that. But also I checked the `build-mondo-ingest` logs, and ICD11 isn't even getting run during that anyway, and I'm not sure why yet.

src/ontology/mondo-ingest.Makefile
```diff
@@ -476,6 +487,7 @@ slurp/%.tsv: $(COMPONENTSDIR)/%.owl $(TMPDIR)/mondo.sssom.tsv $(REPORTDIR)/%_map
 	--mondo-terms-path $(REPORTDIR)/mirror_signature-mondo.tsv \
 	--slurp-dir-path slurp/ \
 	--outpath $@
+.PRECIOUS: slurp/%.tsv
```
joeflack4, line 490: ## Bug fix: `slurp/icd11foundation.tsv`
- [ ] Review @matentzn @twhetzel 

@matentzn Saw that I'd observed `rm slurp/icd11foundation.tsv` after doing `slurp-icd11foundation` and recommended that I add a `.PRECIOUS` marker on the goal. That fixed the issue.

A few other things:
1. **Why just ICD11?**: I don't understand why this was a problem only for `icd11foundation`.
2. **Why did this work?**: After doing some minor reading, I'm not 100% sure I understand why `.PRECIOUS` was needed at all. Since our goal is `slurp/%.tsv`, it would seem to me that `slurp/icd11foundation.tsv` wouldn't be considered an intermediate. [ChatGPT thinks](https://chat.openai.com/share/d21169e2-3d52-4a3c-b6b6-9a1aafd58e0f) that it's considered an intermediate because (1) it has a `%` in it, and (2) because it's being called by a `.PHONY` goal. But I'm still not convinced. We have many such cases of (1) and (2) throughout our pipeline, but I do not see things getting `rm`'ed afterwards.
3. **`rm` issue**: I checked the build log from last week, and at the end it showed `rm reports/icd11foundation_exclusion_reasons.robot.template.tsv slurp/icd11foundation.tsv imports/ro_terms_combined.txt `. After adding that 1 `.PRECIOUS`, now the slurp file is no longer being removed, but the other two still are: `rm reports/icd11foundation_exclusion_reasons.robot.template.tsv imports/ro_terms_combined.txt`. I think we want these, so I'm going to apply the same fix for both of those as well. I pasted more of the log from `build-mondo-ingest` below just in case it's of any value.

<details><summary>Log related to (3)</summary>
<p>

I have the full logs saved as `.txt` files also if interested

From the build last week:
```
Release files are now in ../.. - now you should commit, push and make a release         on your git hosting site such as GitHub or GitLab
rm reports/icd11foundation_exclusion_reasons.robot.template.tsv slurp/icd11foundation.tsv imports/ro_terms_combined.txt
make[1]: Leaving directory '/work/src/ontology'
Mondo Ingest has been fully completed
```

From the build today:
```
Release files are now in ../.. - now you should commit, push and make a release         on your git hosting site such as GitHub or GitLab
rm reports/icd11foundation_exclusion_reasons.robot.template.tsv imports/ro_terms_combined.txt
make[1]: Leaving directory '/work/src/ontology'
Mondo Ingest has been fully completed
```

</p>
</details>

src/ontology/mondo-ingest.Makefile
joeflack4: Addressed by: https://github.com/monarch-initiative/mondo-ingest/pull/434#discussion_r1533102897

src/ontology/mondo-ingest.Makefile
joeflack4: Gonna do this in the future. No longer in scope of this PR unless I hear otherwise:
- https://github.com/monarch-initiative/icd11/issues/9

src/ontology/mondo-ingest.Makefile
```diff
@@ -246,6 +257,7 @@ $(REPORTDIR)/%_term_exclusions.txt $(REPORTDIR)/%_exclusion_reasons.robot.templa
 	--config-path metadata/$*.yml \
 	--outpath-txt $(REPORTDIR)/$*_term_exclusions.txt \
 	--outpath-robot-template-tsv $(REPORTDIR)/$*_exclusion_reasons.robot.template.tsv
+.PRECIOUS: $(REPORTDIR)/%_exclusion_reasons.robot.template.tsv
```
joeflack4, line 260: ## Bug fix: `rm reports/icd11foundation_exclusion_reasons.robot.template.tsv`
- [ ] Review and see questions @matentzn @twhetzel

AFAIK, as of 2024/03/20, this is the only remaining issue that we may want to resolve before merging. That's the only reason why this is still a draft (though #471 is not).

As stated in [the thread about the slurp bug fix](https://github.com/monarch-initiative/mondo-ingest/pull/434#discussion_r1533102897), I noticed at the end that `build-mondo-ingest` that this file was also being removed.

## Questions
Point (1) and (2) in the thread linked above also apply here.

But here is another question that only applies to this case:
If there are two targets in the goal `$(REPORTDIR)/%_term_exclusions.txt $(REPORTDIR)/%_exclusion_reasons.robot.template.tsv`, why is only 1 of them getting removed?

joeflack4:
CC @souzadevinicius This PR handles most of the setup for integrating ICD11 with Mondo. You may want to review this to get an idea of how such a thing is handled.

docs/sources/icd11foundation.md
matentzn, line 1: Are you sure you didnt edit this file manually? The docs are generated automatically from https://github.com/monarch-initiative/mondo-ingest/blob/icd11-ingest/src/ontology/metadata/icd11foundation.yml

src/ontology/config/icd11foundation-property-map.sssom.tsv
matentzn, line 1: Why did you choose to create a new property map rather than using an existing one like 
https://github.com/monarch-initiative/mondo-ingest/blob/icd11-ingest/src/ontology/config/property-map-1.sssom.tsv

Just saying this as ideally, there should be a single mapping file, sort of lile documentation for preprocessing pipeline (there cant be a single one due to some problems with ROBOT rename, but I think in your case why not add them all to an existing one)

src/ontology/metadata/icd11.yml
matentzn, line 1: what is this file? there is also icd11foundation.yml, which seems more appropriate.

src/ontology/metadata/mondo.sssom.config.yml
matentzn: Better avoid comments like this on specific PRs and make specific issues to discuss during our meetings.. Too much to unpack to answer.

matentzn: I dont know, I think this is related to the Mondo repo (not Mondo ingest repo), so you can ignore it for now.

src/ontology/mondo-ingest.Makefile
matentzn: Please make new issue and raise at meeting. Then resolve comment here. In issue, add some concrete examples from the ontology we can discuss!

src/ontology/mondo-ingest.Makefile
matentzn: I have no idea, but this is a good question - for you to answer by yourself. Not for this PR though :P

src/scripts/lexmatch-sssom-compare.py
matentzn: New issue please

docs/sources/icd11foundation.md
joeflack4: I did edit them manually, and they are supposed to be. I edited the `**Comments about this source:**` section.

There is actually an issue with this whole _dynamically generated -> manual edits_ process that I've brought up before, but I didn't see an issue for it, so I made one:
- #481

src/ontology/config/icd11foundation-property-map.sssom.tsv
joeflack4: I think my original comment on this was bundled in a big post and got resolved, but I also think we're on the same page. I think the main idea was that since there were 2 `property-map*.sssom.tsv` files, it seemed arbitrary which I picked, and it seemed like the better idea was to instead create a new file and namespace it, since these changes were just for 1 source.

**Solution a**
I skimmed and I thought you said we could merge these into a single file, so I created: #482. But I just closed it because you say there is some issue, so there must be multiple files. That is strange. In that case, I stick with my original determination: If there have to be multiple such files, then let 1+ of the files that apply to multiple sources be used on them, and any files that apply only to 1 source should be namespaced by that source, just like the rest of the files in the repository (I've started to do this more with the SPARQL files as well). I think it is a good change and makes the codebase easier to understand.

**Solution b**
**_Alternatively_**, if there _must_ be multiple such files because of some weird ROBOT issue, then how about this:
1 file gets all of the mappings `property-map-1.sssom.tsv`, and the other one is empty `property-map-2.sssom.tsv`. We could even rename them `property-map.sssom.tsv` and `property-map-null.sssom.tsv` or something, if it is allowable, and insert a comment into `property-map-null.sssom.tsv`

src/ontology/metadata/icd11.yml
joeflack4: This was a mistake! Deleted.

src/ontology/metadata/mondo.sssom.config.yml
joeflack4: I think for this one you are right. I spoke w/ Trish a little bit about it already. I'll make an issue.

src/scripts/lexmatch-sssom-compare.py
joeflack4: Done!:
- #484

joeflack4: @hrshdhgd Some minor comments I tagged you on here if you can take a look.

src/ontology/mondo-ingest.Makefile
joeflack4: Ok. I talked over it with Trish and she doesn't know either. ChatGPT also doesn't know. This will remain a mystery.

I'm going to continue adding tangent discussions to my PRs, because these are also supposed to get reviewed at meetings and this is the most time efficient and easiest, most contextual way for me to go through these discussions. If I add them to Workflowy or Google Docs, they just get buried and I can't remember the context months later.

src/scripts/lexmatch-sssom-compare.py
hrshdhgd: got it! Thanks..

src/ontology/mondo-ingest.Makefile
joeflack4: Trish and I looked today and verified that this has been addressed aadequately. Marking resolved.

docs/sources/icd11foundation.md
joeflack4: Trish and I went over this, and I moved the contents of 'comments about this source' to the YML: https://github.com/monarch-initiative/mondo-ingest/issues/481#issuecomment-2052615849

and Nico and I made a plan for how to get this programmatically into `.md` files.

src/ontology/config/icd11foundation-property-map.sssom.tsv
joeflack4: And also one advantage of splitting them up is that we can drop `--allow-missing-entities true`, and if the source changes and a mapping is no longer there, we'll know about it because it will error.

joeflack4: I assume there's already an issue in ROBOT about this, so we don't need to make a GitHub issue about needing a 2nd property map file.

joeflack4: Trish and I discussed and we are leaving this for now as is. Added additional comments above and to #482. Decided to close #482 for now as not planned.

src/ontology/mondo-ingest.Makefile
```diff
@@ -476,6 +488,7 @@ slurp/%.tsv: $(COMPONENTSDIR)/%.owl $(TMPDIR)/mondo.sssom.tsv $(REPORTDIR)/%_map
 	--mondo-terms-path $(REPORTDIR)/mirror_signature-mondo.tsv \
 	--slurp-dir-path slurp/ \
 	--outpath $@
+.PRECIOUS: slurp/%.tsv
```
joeflack4, line 491: ## More **`rm` issue**s
- [x] `rm reports/icd11foundation_exclusion_reasons.robot.template.tsv slurp/icd11foundation.tsv`
- [ ] `imports/ro_terms_combined.txt` https://github.com/monarch-initiative/mondo-ingest/issues/477

I checked the build log from last week, and at the end it showed `rm reports/icd11foundation_exclusion_reasons.robot.template.tsv slurp/icd11foundation.tsv imports/ro_terms_combined.txt`. After adding that 1 `.PRECIOUS`, now the slurp file is no longer being removed, but the other two still are: `rm reports/icd11foundation_exclusion_reasons.robot.template.tsv imports/ro_terms_combined.txt`. I think we want these, so I'm going to apply the same fix for both of those as well. I pasted more of the log from `build-mondo-ingest` below just in case it's of any value.

<details><summary>Log related to (3)</summary>
<p>

I have the full logs saved as `.txt` files also if interested

From the build last week:
```
Release files are now in ../.. - now you should commit, push and make a release         on your git hosting site such as GitHub or GitLab
rm reports/icd11foundation_exclusion_reasons.robot.template.tsv slurp/icd11foundation.tsv imports/ro_terms_combined.txt
make[1]: Leaving directory '/work/src/ontology'
Mondo Ingest has been fully completed
```

From the build today:
```
Release files are now in ../.. - now you should commit, push and make a release         on your git hosting site such as GitHub or GitLab
rm reports/icd11foundation_exclusion_reasons.robot.template.tsv imports/ro_terms_combined.txt
make[1]: Leaving directory '/work/src/ontology'
Mondo Ingest has been fully completed
```

</p>
</details> 

I've followed up on these two remaining issues here:
- Thread: [Bug fix: `rm reports/icd11foundation_exclusion_reasons.robot.template.tsv`](https://github.com/monarch-initiative/mondo-ingest/pull/434#discussion_r1533134445)
- #477

PR #440, opened 2024-02-13
Add pipeline for externally managed content

### LINKED ISSUE #439: Represent Externally Managed Content in the Mondo Ingest
Some providers such as NORD, GARD and MedGen directly provide content to Mondo which we pull in.

I would like to suggest, rather than adding a lot of spaghetti code to Mondo repo, to process _all_ external code here in the Mondo Ingest using this process:

1. We work with the content providers to provide us with a table with the information to pull in, say xrefs, labels, and / or subsets
2. We pull these tables and transform them into ROBOT templates here (OWL and TSV)
3. We reduce the code in the Mondo ingest rest (which is currently some hacky stuff which is in a draft PR here:  https://github.com/monarch-initiative/mondo/pull/6903) by a simple process (github action) that pulls the content from here and merges it into Mondo-edit (we dont need to answer the "how" here to move on - its a small change to what is already there).

The advantage is that we once and for all centralise all external content in the mondo ingest pipeline and have no side-loading shows for external content in Mondo repo. 

Note: Externally managed content are _not_ external resources in the sense of external terminologies like ORDO and OMIM. They are simply tables we download and merge into Mondo after we have agreed with the team and the external content provider that we should do so.

matentzn:
Fixes #439

Precursor to https://github.com/monarch-initiative/mondo/issues/6883 

Related to 
- https://github.com/monarch-initiative/mondo/issues/7149
- https://github.com/monarch-initiative/mondo/pull/6903

## Pre-merge checklist

- [x] Docs

- [x] QC
   - `sh run.sh make build-mondo-ingest` has been run on this branch (after `docker pull obolibrary/odkfull:dev), and no errors occurred **OR**

- [X] Account for any new packages

- [ ] Reviewed

## Overview

This PR:

- Introduces a system to turn a table into a ROBOT template
- Adds NORD as a proof of concept


src/ontology/config/external-content-robot-headers.json
```diff
@@ -0,0 +1,11 @@
+{
+    "mondo_id": "ID",
+    "report_ref": "A oboInOwl:hasDbXref",
+    "report_ref_source": ">A oboInOwl:source",	
+    "preferred_name": "A oboInOwl:hasExactSynonym",
+    "preferred_name_source": ">A oboInOwl:hasDbXref",	
+    "synonym_type": ">A oboInOwl:hasSynonymType",	
+    "subset": "AI oboInOwl:inSubset",
```
twhetzel, line 8: `AI`???

src/ontology/config/external-content-robot-headers.json
```diff
@@ -0,0 +1,11 @@
...
+    "mondo_id": "ID",
+    "report_ref": "A oboInOwl:hasDbXref",
+    "report_ref_source": ">A oboInOwl:source",	
+    "preferred_name": "A oboInOwl:hasExactSynonym",
+    "preferred_name_source": ">A oboInOwl:hasDbXref",	
+    "synonym_type": ">A oboInOwl:hasSynonymType",	
+    "subset": "AI oboInOwl:inSubset",	
+    "subset_source": ">A oboInOwl:source",	
+    "subset_source2": ">A oboInOwl:source"
+}
```
twhetzel, line 11: Can this have 3 `">A oboInOwl:source"`? I gather yes, but not comprehending this at the moment.

src/ontology/mondo-ingest.Makefile
```diff
@@ -541,6 +541,28 @@ $(REPORTDIR)/%.subclass.added.robot.tsv $(REPORTDIR)/%.subclass.confirmed.robot.
...
 	--onto-config-path metadata/$*.yml
 
+##################################
+## Externally managed content ####
+##################################
+
+EXTERNAL_CONTENT_DIR=external
+
+$(TMPDIR)/nord.tsv:
+	wget "https://rdbdev.wpengine.com/wp-content/uploads/mondo-export/rare_export.tsv" -O $@
```
twhetzel, line 551: How are versions of downloaded content managed between one run to the next, e.g. appended with the download date or moved to an archive type folder and the date added then?

Do the downloaded files always contain all content or are only new and changed content?

src/scripts/add-robot-template-header.py
```diff
@@ -0,0 +1,28 @@
...
+    # Create a new row based on the JSON config
+    new_row = {col: config.get(col, '') for col in df.columns}
+
+    # Append the new row to the DataFrame
+    df = pd.concat([pd.DataFrame([new_row]), df], ignore_index=True)
+
+    # Output the modified DataFrame
+    output = StringIO()
+    df.to_csv(output, sep='\t', index=False)
+    click.echo(output.getvalue())
```
twhetzel, line 25: I think this prints the entire df to the console, is that intended?

src/ontology/mondo-ingest.Makefile
twhetzel: We can take this offline, added to meeting notes.

twhetzel (approved):
looks good - a few questions/comments, but approved

src/ontology/config/external-content-robot-headers.json
matentzn: An annotation to an IRI: http://robot.obolibrary.org/template#generic-template-strings

If it's just A, the expectation is that the value is a literal.

src/ontology/config/external-content-robot-headers.json
matentzn: Yes, these would be three individual source annotations. If you look at an example class in Mondo, you can often find more than one source on each term.

src/ontology/mondo-ingest.Makefile
matentzn: Actually this is a good point, and I can easily add that. I will ping you once it is done. It is achieved by tagging the robot template owl version with the downloaded date.

src/scripts/add-robot-template-header.py
matentzn: Yes, this is how a lot of the unix tools operate. It save you the necessity of passing in an out file parameter and writing to it. Instead you simple redirect the output of this scripts to the terminal with the ">" symbol.

src/ontology/config/external-content-robot-headers.json
twhetzel: yeah, I've seen that. given the issue with "double comments" when obsoleting a term, i didn't know ROBOT could handle multiple sources.

src/scripts/add-robot-template-header.py
twhetzel: gotcha. are there any size limits of the `output` we should be aware of or check for before this line? for example, how big can the `output` be and this work?

matentzn: Good point.. I can't think of anything - the main point of streaming to std out (other than supporting use in unix pipelines) would be that the memory consumption is minimised. But that is a bit beyond my expertise.. it's a frequently used pattern in our scripts (oak does it too) so I hope it's alright..

matentzn: I will close this now, because we have been doing it so far, but we will keep it in mind when size limits (Out of memory errors etc) are emerging.

src/ontology/mondo-ingest.Makefile
matentzn: As you can see below, I now added explicit versioning information at least to the OWL file:

```
$(EXTERNAL_CONTENT_DIR)/%.robot.owl: $(EXTERNAL_CONTENT_DIR)/%.robot.tsv
	$(ROBOT) template --template $< \
	 	annotate \
				--ontology-iri $(URIBASE)/mondo/external/nord.robot.owl \
				--version-iri $(URIBASE)/mondo/external/$(TODAY)/nord.robot.owl \
				-o $@
```

This looks like:

```
        <owl:versionIRI rdf:resource="http://purl.obolibrary.org/obo/mondo/external/2024-02-22/nord.robot.owl"/>
```

The cool thing is that we can make this work like a PURL, and it tells us when it was created!

The "archiving" strategy emerges than naturally:

When we make a data refresh, we create a GitHub release. The GitHub release creates a `tag`, e.g. `v2024-02-22`. That `tag` can be used to access the datafile (archiving strategy). This is cool because it is exactly how it is done across all of OBO! 

If you have more questions about this, add to meeting minutes and we can debate! :)

matentzn:
@twhetzel I added a bit of versioning and finished the docs for this PR. I am now running the pipeline to see if nothing breaks! Can you check the docs to see if they are ok for now?

joeflack4:
@matentzn @twhetzel I woulda liked to have reviewed this PR. I did just now, but I was mistaken and thought it was just a direct commit, so I put [my review on the commit itself](https://github.com/monarch-initiative/mondo-ingest/commit/2f5981d3957d052fe9710c543d0a152b8420af41).

PR #443, opened 2024-02-17
Bugfix for fresh clone: `tmp/` does not exist

joeflack4:
## Overview
Resolves `cannot create tmp/debug.log: Directory nonexistent`. If run-command.sh runs on a fresh clone, it will fail because it tries to append to a debug log, but it can't do that because the directory for that file doesn't exist.

## Pre-merge checklist
- [x] Docs
- [x] QC
   - `sh run.sh make build-mondo-ingest` has been run on this branch (after `docker pull obolibrary/odkfull:dev), and no errors occurred **OR**
- [x] Account for any new packages
- [x] Reviewed

## Changes
    Bugfix tmp/ not exist
    If run-command.sh runs on a fresh clone, it will fail because it tries to append to a debug log, but it can't do that because the directory for that file doesn't exist.

--- 

CC @souzadevinicius

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,6 +96,7 @@ src/ontology/target/
 
 # src/ontology/tmp/
 src/ontology/tmp/*
+!src/ontology/tmp/.gitkeep
 !src/ontology/tmp/README.md
 
 # src/patterns/


src/scripts/run-command.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh
+mkdir -p tmp/
```
joeflack4, line 2: ## The bugfix
1-liner.

### How the issue manifested
I was running `build-mondo-ingest` on a fresh clone and got `cannot create tmp/debug.log: Directory nonexistent`.

<p>

```
make refresh-imports exclusions-all slurp-all mappings matches \
	recreate-unmapped-components sync documentation \
	prepare_release
../scripts/run-command.sh: 3: cannot create tmp/debug.log: Directory nonexistent
/usr/bin/time: tmp/debug.log: No such file or directory
make: *** [mondo-ingest.Makefile:335: build-mondo-ingest] Error 125
Command exited with non-zero status 2
### DEBUG STATS ###
Elapsed time: 0:00.00
Peak memory: 2176 kb
```

</p>
</details> 

### Questions
1. @matentzn @hrshdhgd **Fresh clone:** Do you guys always run builds on a fresh clone? Seemed safer to me.
2. @matentzn @twhetzel **Main branch:** I think this fix should go directly into `main`. Want me to do that instead or maybe do 2 separate PRs for the same change?
I assume we could run into difficulties trying to do merges to `main` and then `rebase` `develop`. In this case, probably not since we recently branched.
3. @matentzn **ODK**: Shall I open up a PR in ODK?
I see this file [in ODK](https://github.com/INCATools/ontology-development-kit/blob/e70cf5820485d32631e94340de36c67540ddc294/template/src/scripts/run-command.sh), so this should be a problem universally. That file is 4 months old. I'm surprised that this hasn't been noticed yet. I don't see a fix on the `develop` branch there either.

matentzn: Unfortunately this is not the right place to do this. please for now just put the `mkdir -p` into the concrete make goal that depends on it; if you want to make a PR on ODK, it should be done using this solution: https://www.theserverside.com/blog/Coffee-Talk-Java-News-Stories-and-Opinions/gitkeep-push-empty-folders-git-commit

matentzn (changes requested):

src/scripts/run-command.sh
matentzn: Feel free to simply apply this solution in mondo ingest though!

joeflack4: Yeah that makes sense. I'm not able to locate how this is run, though. I only see `run-command.sh` being erefrenced in `Makefile` as `SHELL = $(SCRIPTSDIR)/run-command.sh`, but I don't see anywhere where this is getting called.

I was thinking about a `.keep` / `.gitkeep` solution as well. So I'll just go with that both ODK and here as you suggest.

joeflack4: @matentzn I created a PR in ODK w/ your solution: https://github.com/INCATools/ontology-development-kit/pull/994

However I'm not sure I agree with this solution now. Please look at [these comments](https://github.com/INCATools/ontology-development-kit/pull/994#discussion_r1495076861), particularly question 1: _ I think this fix will only work for newly instantiated repos, but not already existing ones. What to do?
_.

Let's leave this PR open until we have sorted out the ODK PR.

matentzn: Just hotfix it in the mondo-ingest repo, manually! It does not have to work with "update_repo" once its commited, it works for mondo-ingest as well.

joeflack4: @matentzn @twhetzel Done! This should be ready for merge~

joeflack4: Actually, hopefully you guys are OK with this, this is a small PR and it is approved. I'm going to merge now and rebase so that when I run my `build-mondo-ingest` in a fresh clone of #412, it doesn't fail on this.

joeflack4: But I do have a question:
Shall I make a new PR and apply this change to `main` as well?
Otherwise those builds will also fail.

twhetzel:
The fix is ok.

PR #445, opened 2024-02-21
Fix: Double echo

joeflack4:
## Changes
    Resolved double echo
    - UX improvement / fix of logging from mondo-ingest.Makefile, where echo's would be printed twice because they did not start with @.

---

CC @souzadevinicius

### DIFF
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -271,10 +271,10 @@ $(REPORTDIR)/%_excluded_terms_in_mondo_xrefs.tsv $(REPORTDIR)/%_excluded_terms_i
 # Exclusions: all artefacts for single ontology
 .PHONY: exclusions-%
 exclusions-%: reports/%_exclusion_reasons.ttl reports/%_excluded_terms_in_mondo_xrefs.tsv  $(REPORTDIR)/%_term_exclusions.txt 
-	echo "$@ completed"
+	@echo "$@ completed"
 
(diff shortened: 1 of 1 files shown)


src/ontology/mondo-ingest.Makefile
```diff
@@ -400,7 +400,7 @@ ALL_COMPONENT_SIGNTAURE_REPORTS=$(foreach n,$(ALL_COMPONENT_IDS), reports/mirror
 
 .PHONY: signature_reports
 signature_reports: $(ALL_MIRROR_SIGNTAURE_REPORTS) $(ALL_COMPONENT_SIGNTAURE_REPORTS)
-	echo "Finished running signature reports.."
```
joeflack4, line 403: Also removed a second `.` here

src/ontology/mondo-ingest.Makefile
```diff
@@ -336,15 +336,15 @@ build-mondo-ingest:
 		mapped-deprecated-terms mapping-progress-report \
 		recreate-unmapped-components sync documentation \
 		prepare_release
-	echo "Mondo Ingest has been fully completed"
+	@echo "Mondo Ingest has been fully completed"
```
joeflack4, line 339: Example of one of the changes

joeflack4, line 339: ## ODK PR
Note also that the same issue also exists in the `Makefile`, e.g. `echo Finished running all tests successfully.`, which comes from ODK. So I'd like to do the same quick PR in ODK as well.

matentzn (approved):
Good move, thanks!

PR #451, opened 2024-02-22
Bugfix for fresh clone: `tmp/` does not exist

joeflack4:
## Overview
If run-command.sh runs on a fresh clone, it will fail because it tries to append to a debug log, but it can't do that because the directory for that file doesn't exist.

### Documentation
- Was the documentation added/updated under `docs/`?
   - [ ] Yes

### QC
- Was the full pipeline run before submitting this PR using `sh run.sh make build-mondo-ingest` on this branch (after `docker pull obolibrary/odkfull:dev`), and no errors occurred?
   - [ ] Yes
   
### New Packages
- Were any new *Python packages* added?
   - [x] No
</br>

- Were any other *non-Python packages* added?
   - [x] No

- Has the PR been sufficiently reviewed by at least one team member of the Mondo Technical team and all threads resolved?
   - [x] Yes

## Additional info
Related:
- #443

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,6 +96,7 @@ src/ontology/target/
 
 # src/ontology/tmp/
 src/ontology/tmp/*
+!src/ontology/tmp/.gitkeep
 !src/ontology/tmp/README.md
 
 # src/patterns/


joeflack4:
Merging as this is a mirror of #443 for `main` instead of `develop` and has been pre-approved.

PR #458, opened 2024-03-02
Bugfix: `unmapped/` location

joeflack4:
## Overview
Fixed bug in which location where output .owl's of unmapped terms were not matching up with the goal's target location. Now they should be going into the correct location.

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [x] No

- [x] No

- [ ] Yes

## Changes
Bugfix: unmapped/ outdir
Fixed bug in which location where output .owl's of unmapped terms were not matching up with the goal's target location.

### DIFF
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -236,7 +236,7 @@ unmapped/:
 
 unmapped/%-unmapped.owl: $(COMPONENTSDIR)/%.owl reports/%_unmapped_terms.tsv | unmapped/
 	cut -f 1 reports/$*_unmapped_terms.tsv | tail -n +2 > reports/$*_unmapped_terms.txt
-	$(ROBOT) filter -i components/$*.owl -T reports/$*_unmapped_terms.txt -o components/$@
+	$(ROBOT) filter -i components/$*.owl -T reports/$*_unmapped_terms.txt -o $@
 	rm reports/$*_unmapped_terms.txt
 
 .PHONY: recreate-unmapped-components


src/ontology/mondo-ingest.Makefile
```diff
@@ -236,7 +236,7 @@ unmapped/:
 
 unmapped/%-unmapped.owl: $(COMPONENTSDIR)/%.owl reports/%_unmapped_terms.tsv | unmapped/
 	cut -f 1 reports/$*_unmapped_terms.tsv | tail -n +2 > reports/$*_unmapped_terms.txt
-	$(ROBOT) filter -i components/$*.owl -T reports/$*_unmapped_terms.txt -o components/$@
```
joeflack4, line 239: ## The update
I'm surprised we didn't notice this sooner, but I have 2 dirs: `src/ontology/unmapped/` and `src/ontology/components/unmapped/`. The mismatch is:
- Expected: Goal says the files should go into `unmapped/`
- Actual: The files are actually going into `components/unmapped/`

I remember that there was some reason why we were initially putting things in one directory or the other, either as a simple solution to a bug, or to avoid committing the files. If I remember correctly though, we do want these files committed (to `unmapped/`), so I have made that change.

matentzn: We didnt notice it because it correctly added the files into the components/unmapped directory :P Very good catch.

I agree with your assesment, and I cant remember the historical reasons.

matentzn (approved):

PR #478, opened 2024-03-21
Bug fix: OMIMPS namespace && no OMIM parents in migration

joeflack4:
Resolves:
- #438
- https://github.com/monarch-initiative/omim/issues/103

## Overview
This PR resolves two things:
1. **OMIMPS namespace**: Was recently changed (https://github.com/monarch-initiative/omim/pull/107#discussion_r1534845966), which addressed https://github.com/monarch-initiative/omim/issues/103. I thought I raised a review comment about this, but either I didn't, or we forgot to follow up with the rest of the tasks in that issue, which leads to the necessity of this change here in `mondo-ingest` to match the updated namespace / URI prefix.
2. **Parents missing in OMIM slurp**: The reason this was happening is because all OMIM parents are in the OMIMPS namespace. However, until addressing (1) in `mondo-ingest` with this PR, those parents could no longer be detected.

### Documentation
Was the documentation added/updated under `docs/`?
- [x] Yes
- [ ] No, updates to the docs were not necessary after careful consideration

### QC
Was the full pipeline run before submitting this PR using `sh run.sh make build-mondo-ingest` on this branch (after 
`docker pull obolibrary/odkfull:dev`), and no errors occurred?
- [x] Yes #479
- [ ] No, there are no functional (code-related) changes to the pipeline in the PR, so no re-run is necessary
   
### New Packages
Were any new *Python packages* added?
- [ ] Yes, [requirements.txt.full](https://github.com/INCATools/ontology-development-kit/blob/master/requirements.txt.full) was updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

Were any other *non-Python packages* added?
- [ ] Yes, [Dockerfile](https://github.com/INCATools/ontology-development-kit/blob/master/Dockerfile) updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

### PR Review and Conversations Resolved
Has the PR been sufficiently reviewed by at least 1 team member of the Mondo Technical team and all threads resolved?
- [ ] Yes

## Related
- https://github.com/monarch-initiative/mondo-ingest/pull/479

PR #479, opened 2024-03-22
Bug fix: OMIMPS namespace && no OMIM parents in migration - data

joeflack4:
Data PR for:
- #478

No errors.

Elapsed time: 3:40:38


docs/metrics/icd11foundation.md
```diff
@@ -2,15 +2,15 @@
 
 **IRI:** http://purl.obolibrary.org/obo/mondo-ingest/components/icd11foundation.owl
 
-**Version IRI:** http://purl.obolibrary.org/obo/mondo-ingest/releases/2024-02-17/components/icd11foundation.owl
+**Version IRI:** http://purl.obolibrary.org/obo/mondo-ingest/releases/2024-03-22/components/icd11foundation.owl
```
joeflack4, line 5: ## FYI: Size of `icd11foundation-unmapped.owl`
I got:
```
remote: warning: File src/ontology/unmapped/icd11foundation-unmapped.owl is 70.64 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB
```

The limit is 100mb, and this is getting a little close. I take it this will often be the case for `*-unmapped.owl` files for new sources with lots of terms.

matentzn: I agree. My personal feeling is to commit it anyways, but if you or @twhetzel feel like this should be gitignored, its also fine. No issue either way

joeflack4:
The non-data mirror of this PR, #478, has been merged, so I'm gonna close this now.

I suppose in some cases we could merge these data PRs, but in some cases that will lead to conflicts, and when that happens, my thought is that neither side of the conflict is correct, and what needs to be done is to run a new build. So usually I will not merge these I think.

PR #480, opened 2024-03-22
ORDO sparql update

### LINKED ISSUE #469: `fix_complex_reification.ru` updates
## Overview
Does this file only apply to ORDO?

I've since removed this line from the `icd11foundation` goal, but it's still on several other goals.

I ask because it at the end, it has:
```sparql
  FILTER (isIRI(?entity) && (regex(str(?entity), "http://www.orpha.net/ORDO/")))
  BIND(IF(STRSTARTS(str(?precision_value), "NTBT"), skos:narrowMatch, IF(STRSTARTS(str(?precision_value), "BTNT"), skos:broadMatch, oboInOwl:hasDbXref)) as ?property)
```

## Suggestions
- [x] 1. If it indeed only applies to ORDO, then I suggest we remove this line from all the goals except for ORDO, as done in PR with similar situation: #459.
- [x] 2. If that's the case, I suggest we also prepend `ordo_` to the filename (like I've suggest in similar situation https://github.com/monarch-initiative/mondo-ingest/pull/459#discussion_r1510414581).

---

- [x] Advise @matentzn / @twhetzel 
- [x] Implement any decisions @joeflack4

joeflack4:
Resolves #469

## Overview
This is a quick code cleanliness update; removes operation from all component goals except for Orphanet. Also updates filename for clarity as to its purpose.

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [x] No

- [x] No

- [x] Yes


src/ontology/mondo-ingest.Makefile
```diff
@@ -70,7 +70,6 @@ $(COMPONENTSDIR)/omim.owl: $(TMPDIR)/omim_relevant_signature.txt | component-dow
 		query \
 			--update ../sparql/fix_omimps.ru \
 			--update ../sparql/fix-labels-with-brackets.ru \
-			--update ../sparql/fix_complex_reification.ru \
```
joeflack4, line 73: ## Deletions
From every goal other than for ORDO.

src/ontology/mondo-ingest.Makefile
```diff
@@ -83,7 +82,7 @@ $(COMPONENTSDIR)/ordo.owl: $(TMPDIR)/ordo_relevant_signature.txt config/properti
 		query \
 			--update ../sparql/fix_partof.ru \
 			--update ../sparql/fix_deprecated.ru \
-			--update ../sparql/fix_complex_reification.ru \
+			--update ../sparql/fix_complex_reification_ordo.ru \
```
joeflack4, line 85: ## Rename
Renamed the file and all of its references.

For filename namespacing options I considered:
a. `fix_complex_reification_ordo.ru`
b. `ordo_fix_complex_reification.ru`

I was originally gonna go with 'b'. I felt like namespacing primarily by source is a good idea and it's also usually what we do. But that wasn't the case for the `sparql/` folder. So I went with 'a'. Some examples of other cases where the source namespace was not at the beginning:
- `ordo_fix_complex_reification.ru`
- `fix_make_omim_exact.ru`

@twhetzel Perhaps I can make an issue for changing the namespacing, or create some grouping issue for misc codestyle / cleanup stuff like this. It's super low priority.

joeflack4, line 85: ## SPARQL renamings: Similar issues?
@twhetzel I noticed there was another file where the namespace of the component being operated on was not in the filename. Do you mind if I do another quick PR for this? It's low priority but also quick.

`fix_make_omim_exact.ru`: This file actually corrects something to do with OMIM, but it is only run on the _DOID_ component. It's not a mistake though. It's just updating some matches in that file. For namespacing change, I'd recommend `doid_fix_make_omim_refs_exact.ru`. But if I do the source namespace to the front, that makes it consistent w/ our repo but not with the `sparql/` folder. So I could also take the time to standardize those other names in the same PR.

I could also make a follow-up issue to look for more of these, or add to a grouping issue of similar code cleanliness as I proposed in my other comment, since it's a low priority.

matentzn (approved):

PR #486, opened 2024-04-04
Include ICD11 mappings

### LINKED ISSUE #140: Add ICD11 mappings back
Follow-up to : https://github.com/monarch-initiative/mondo-ingest/pull/135.

As per this PR we are `DELETE`ing the prefix `ICD11:` as a `dbxref`. We need to add it back when needed.

hrshdhgd:
Resolves  #140 and Addresses  #16 (bullet 3)

## Overview

This PR:
- Removes filter to ignore ICD11 refs.

### New Packages
- [x] No

- [x] No

- [x] Yes

### DIFF
--- a/src/sparql/rm_xref_by_prefix.ru
+++ b/src/sparql/rm_xref_by_prefix.ru
@@ -16,5 +16,5 @@ WHERE {
     oboInOwl:hasDbXref ?xref ;
     ?p1 ?o2 .
   }
-  FILTER( STRSTARTS(str(?xref), "UMLS_ICD9CM_2005_AUI:") || STRSTARTS(str(?xref), "ICD11:") || STRSTARTS(str(?xref), "SNOMEDCT_US_") || STRSTARTS(str(?xref), "IMDRF:") || STRSTARTS(str(?xref), "url:") )
+  FILTER( STRSTARTS(str(?xref), "UMLS_ICD9CM_2005_AUI:") || STRSTARTS(str(?xref), "SNOMEDCT_US_") || STRSTARTS(str(?xref), "IMDRF:") || STRSTARTS(str(?xref), "url:") )
 }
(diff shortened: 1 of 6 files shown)


matentzn (approved):

src/sparql/rm_xref_by_prefix.ru
```diff
@@ -16,5 +16,5 @@ WHERE {
     oboInOwl:hasDbXref ?xref ;
     ?p1 ?o2 .
   }
-  FILTER( STRSTARTS(str(?xref), "UMLS_ICD9CM_2005_AUI:") || STRSTARTS(str(?xref), "ICD11:") || STRSTARTS(str(?xref), "SNOMEDCT_US_") || STRSTARTS(str(?xref), "IMDRF:") || STRSTARTS(str(?xref), "url:") )
```
joeflack4, line 19: Just commenting on my understanding.
I see what's going on here to address #140. So this deletion query gets run on the DOID and NCIT components, removing these xrefs. You're removing ICD11 from this query, so thus we're adding the mappings back from these files, as they won't be removed any longer.

I thought that #140 had something to do with my new ICD11Foundation work but it does not.

src/sparql/rm_xref_by_prefix.ru
```diff
@@ -16,5 +16,5 @@ WHERE {
...
+  FILTER( STRSTARTS(str(?xref), "UMLS_ICD9CM_2005_AUI:") || STRSTARTS(str(?xref), "SNOMEDCT_US_") || STRSTARTS(str(?xref), "IMDRF:") || STRSTARTS(str(?xref), "url:") )
```
joeflack4, line 19: ## Merging into `develop` or `main`
@hrshdhgd I see that this is merging into main, but basically from now on, per #405 (which isn't completed yet) we're going to follow the normal git flow of `main` <-- `develop` <-- feature branches.

We could change the base on this PR but IDK how much it matters. I update `develop` from master regularly as needed anyway.

joeflack4 (approved):

src/sparql/rm_xref_by_prefix.ru
joeflack4: @hrshdhgd I approve as well as Nico. Prolly good to merge. I can update `develop` later.

PR #492, opened 2024-04-16
`develop` branch: Build results / data files

joeflack4:
Resolves #ISSUE(s).

## Overview

This PR:
- One
- Two
- Three

### Documentation
- [ ] Yes

### QC
- [x] Yes
   
### New Packages
- [x] No

- [x] No

- [ ] Yes


src/mappings/icd11foundation.sssom.tsv
```diff
@@ -0,0 +1,11 @@
+# curie_map:
+#   owl: http://www.w3.org/2002/07/owl#
+#   rdf: http://www.w3.org/1999/02/22-rdf-syntax-ns#
+#   rdfs: http://www.w3.org/2000/01/rdf-schema#
+#   semapv: https://w3id.org/semapv/vocab/
+#   skos: http://www.w3.org/2004/02/skos/core#
+#   sssom: https://w3id.org/sssom/
+# license: https://w3id.org/sssom/license/unspecified
+# mapping_set_id: https://w3id.org/sssom/mappings/8b5fc9cd-51d1-4baa-a4de-80c9dd83e0f2
+
```
joeflack4, line 10: ## Some empty mapping files
- [ ] @matentzn Double check

I believe you reviewed this already in a previous build of the ICD11 PR before we merged it, and I think this is OK, but we never discussed it, so I just wanted to double check.

OK files (>1k lines):
- src/ontology/lexmatch/unmapped_icd11foundation_lex.tsv 
- src/ontology/lexmatch/unmapped_icd11foundation_lex_exact.tsv

Empty files:
- src/mappings/icd11foundation.sssom.tsv
- src/ontology/lexmatch/mondo-only/unmapped_icd11foundation_mondo.tsv
- src/ontology/lexmatch/mondo-only/unmapped_icd11foundation_mondo_exact.tsv

joeflack4, line 10: ## General review
- [ ] @twhetzel Review

CC @matentzn 
So, we have this SOP #405 for development workflow. Trish and I discussed, and we added a new final step that wasn't there before, which happens after we merge `develop` into `main`:
- Run `build-mondo-ingest` again on main

It feels annoying to run the build back to back but it seems like the safe thing to do.

Another thing that's not clear based on the SOP, but what @twhetzel and I decided earlier that we will do in this case, and I suppose will be establishing precedent: When doing a build from `develop`, if it passes without error, it's permissible (and encouraged) that the person doing the build go ahead and merge into `main` without review.

If that's not good precedent, let's discuss. Maybe we should do a review pre-merge, but given the narrow time frame and that all of us have our attention on this, it should be OK to do this time.

twhetzel: @joeflack4 can we take this comment to one of our meetings so we don't clog up the PR with this process question? 

Also, remember only the code changes from `develop` get merged into `main` not any data files.

PR #499, opened 2024-04-22
Bug fix: `components/ordo.owl` missing mapping annotations

### LINKED ISSUE #485: `ordo-replace-annotation-based-mappings.ru` broken
## Overview
While working on https://github.com/monarch-initiative/icd11/pull/12, [Nico suggested](https://github.com/monarch-initiative/icd11/issues/10#issuecomment-2017919271) to _write a SPARQL query that reads ICD11 xrefs and asserts SKOS exact matches_.  There actually already was such a file that is supposed to do this, `ordo-replace-annotation-based-mappings.ru`. However, there are no `skos` matches annotations in `components/ordo.owl`, so this is broken.

### a. `obo:ECO_0000218` removed (_probably not the cause_)
The SPARQL file depends on this. Example: `<obo:ECO_0000218 xml:lang="en">E (Exact mapping: the two concepts are equivalent)</obo:ECO_0000218>`. It's not in `config/properties.txt`, so it gets removed. However, this happens after the SPARQL update happens, so I don't think this is the issue.

### b. Mapping string has changed
Looks like this is the cause.

What `ordo-replace-annotation-based-mappings.ru` expects:
> - E (Exact mapping: the two concepts are equivalent).\n- Specific code (The ORPHA code has its own code in the ICD10).

What the Orphanet OWL actually has as of 2024/03: 
> - E (Exact mapping: the two concepts are equivalent).\nSpecific code (ICD-10/ICD-11: ORPHAcode has its own code in the targeted terminology).

**Possible solutions**:
I suggest that we change this:
> `FILTER(STRSTARTS(STR(?mapping_precision_string), "- E (Exact mapping: the two concepts are equivalent).\n- Specific code (The ORPHA code has its own code in the ICD10)."))`

to this:
> `FILTER(STRSTARTS(STR(?mapping_precision_string), "- E (`

And we should do the sam ething with the other codes. I think they are: `- BTNT (`, `- NTBT (`, `- ND (`

joeflack4:
Resolves #485

Stalled because this is more involved than I thought it would be, and there are several options. And it is low priority.

## Overview

This PR:
- One
- Two
- Three

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [x] No

- [x] No

- [ ] Yes

### DIFF
--- a/src/ontology/config/ordo-mapping-codes_relationship-curies.csv
+++ b/src/ontology/config/ordo-mapping-codes_relationship-curies.csv
@@ -1,6 +1,0 @@
-﻿ordo_mapping_code,relationship_curie
-E,skos:exactMatch
-NTBT,skos:broadMatch
-BTNT,skos:narrowMatch
-ND,oboInOwl:hasDbXref
-W,sssom:notExactMatch
--- a/src/scripts/ordo_mapping_annotations/create_sparql__ordo_mapping_annotations_violation.py
+++ b/src/scripts/ordo_mapping_annotations/create_sparql__ordo_mapping_annotations_violation.py
@@ -1,3 +1,4 @@
+# TODO: this file might be obsolete now too
 """Update ORDO mapping annotations to formalize them semantically.
 Work described furthere here: https://github.com/monarch-initiative/mondo-ingest/issues/23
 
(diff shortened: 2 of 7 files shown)


joeflack4:
@twhetzel I'm closing this PR as it is more involved than I originally anticipated and it is a low priority.

Issue #482, opened 2024-03-29
Handling multiple `config/*property-map*.sssom.tsv`s


joeflack4:
## Overview
There are multiple `config/*property-map*.sssom.tsv` files. De-proliferate these.

## Sub-task list
- [ ] 1. Wait on #434 to be merged first, as that will introduce a new file
- [ ] 2. Implement solution
- [ ] 3. Update `mondo-ingest.Makefile`

### 2. Implement solution
a. Single file (if allowable, though there may be some weird ROBOT issue prohibiting it)
b. Merge all mappings into 1 file, and have a second, empty file (to get around ROBOT issue)
c. Keep `property-map-1` and `property-map-2` as-is, and then add other ones, e.g.`icd11foundation-property-map` file into one of these. One disadvantage of that is [this](https://github.com/monarch-initiative/mondo-ingest/issues/482#issuecomment-2052627274).

Additional details about solutions: https://github.com/monarch-initiative/mondo-ingest/pull/434#discussion_r1544921463

joeflack4:
Context: [this comment](https://github.com/monarch-initiative/mondo-ingest/pull/434#discussion_r1544323110)

@twhetzel I'm making this one high priority just because it is such a quick change to make.

joeflack4:
Actually maybe we can't do this

joeflack4:
And also one advantage of splitting them up is that we can drop `--allow-missing-entities true`, and if the source changes and a mapping is no longer there, we'll know about it because it will error.

joeflack4:
There's also the possible problem that if we merge them into 1 file and leave the other one blank, then when ROBOT fixes this problem, it might get annoyed by the blank file? Maybe not.

joeflack4:
We've decided this issue is a bit of a distraction. We can always re-open it if needed, especially if ROBOT fixes the issue where 2 property maps / more than one are arbitrarily required.

joeflack4:
[_Nico wrote_](https://github.com/monarch-initiative/mondo-ingest/pull/434#discussion_r1544323110):
> there cant be a single one due to some problems with ROBOT rename

@matentzn BTW, I couldn't find an issue for this in the [robot GitHub](https://github.com/ontodev/robot/issues).If there's not one there, should I make one?

matentzn:
If you like, the issue would go something like this:

> Add `--allow-duplicate-entities` as option in ROBOT rename.

[...]

PR #500, opened 2024-04-28
Combined `property-map.sssom.tsv`

joeflack4:
Related to:
- #482 
- https://github.com/ontodev/robot/issues/1195

Data PR:
- #505

## Overview
There have been 2 property map TSV files used by robot rename. However, we are now utilizing the `--allow-duplicates` flag, which allows us to consolidate them into 1 file.

### Documentation
- [ ] Yes

### QC
- [x] Yes
   
### New Packages
- [x] No

- [x] No

- [x] Yes

## Changes
    Consolidate property maps
    - There have been 2 property map TSV files used by robot rename. However, we are now utilizing the --allow-duplicates flag, which allows us to consolidate them into 1 file.
    - Delete: one of the property map files
    - Rename: the other of the files
    - Update: components goals to utilize the new/updated file
    - Update: Docs regarding this aspect of adding a new source

### DIFF
--- a/docs/developer/add-new-source.md
+++ b/docs/developer/add-new-source.md
@@ -55,12 +55,11 @@ When `mondo-ingest-odk.yaml` is edited and `update_repo` is run, a new `componen
 `src/ontology/Makefile`. This essentially downloads a copy of the source. But, for it to be used in `mondo-ingest`, it 
 needs to be modified. This is done by creating a `$(COMPONENTSDIR)/omim.owl` goal in `mondo-ingest.Makefile` which runs 
 a composite `robot` command. Here are some sections you'll likely want to add that command:  
-- `rename --mappings config/SOURCE-property-map.sssom.tsv`: You'll notice that there are two other lines on most goals 
-for `property-map.sssom.tsv` and `property-map-2.sssom.tsv`. These handle some common renamings. However, 
-if your source has any properties that need to be renamed which are specific to that source, then you should create a 
-new file for that. While doing this, it might be useful to check any other `SOURCE-property-map.sssom.tsv` files to see 
-if they share any of the same property renamings. If so, it might be a good idea to move them to 
-`property-map.sssom.tsv` instead.
+- `rename --mappings config/SOURCE-property-map.sssom.tsv`: You'll notice that all components goals have a such a line 
+using `property-map.sssom.tsv`. This is for some common property renamings. However, if your source has any properties 
+that need to be renamed which are specific to that source, then you should create a new file for that. While doing this,
+ it might be useful to check any other `SOURCE-property-map.sssom.tsv` files to see if they share any of the same  
+property renamings. If so, it might be a good idea to move them to `property-map.sssom.tsv` instead.
 - `query --update ../sparql/SPARQL.ru`: There may be more complex modifications that need to be made to the ontology. If
  so, this can be done by 1 or more SPARQL queries.
 - `remove -T config/properties.txt --select complement --select properties --trim true`: `properties.txt` contains a 
--- a/src/ontology/config/property-map.sssom.tsv
+++ b/src/ontology/config/property-map.sssom.tsv
@@ -1,4 +1,7 @@
 subject_id	object_id
+http://purl.org/dc/elements/1.1/source	oboInOwl:source
+http://www.w3.org/2004/02/skos/core#prefLabel	http://www.w3.org/2000/01/rdf-schema#label
+http://www.w3.org/2004/02/skos/core#altLabel	oboInOwl:hasExactSynonym
 http://www.ebi.ac.uk/efo/definition	http://purl.obolibrary.org/obo/IAO_0000115
 http://www.ebi.ac.uk/efo/reason_for_obsolescence	rdfs:comment
 http://www.ebi.ac.uk/efo/alternative_term	oboInOwl:hasExactSynonym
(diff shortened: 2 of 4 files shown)


src/ontology/config/property-map.sssom.tsv
```diff
@@ -1,4 +1,7 @@
 subject_id	object_id
+http://purl.org/dc/elements/1.1/source	oboInOwl:source
+http://www.w3.org/2004/02/skos/core#prefLabel	http://www.w3.org/2000/01/rdf-schema#label
+http://www.w3.org/2004/02/skos/core#altLabel	oboInOwl:hasExactSynonym
 http://www.ebi.ac.uk/efo/definition	http://purl.obolibrary.org/obo/IAO_0000115
 http://www.ebi.ac.uk/efo/reason_for_obsolescence	rdfs:comment
```
joeflack4, line 6: ## New file
Combines what was previously (now deleted):
- `property-map-1.sssom.tsv`
- `property-map-2.sssom.tsv`

docs/developer/add-new-source.md
```diff
@@ -55,12 +55,11 @@ When `mondo-ingest-odk.yaml` is edited and `update_repo` is run, a new `componen
...
 needs to be modified. This is done by creating a `$(COMPONENTSDIR)/omim.owl` goal in `mondo-ingest.Makefile` which runs 
 a composite `robot` command. Here are some sections you'll likely want to add that command:  
-- `rename --mappings config/SOURCE-property-map.sssom.tsv`: You'll notice that there are two other lines on most goals 
-for `property-map.sssom.tsv` and `property-map-2.sssom.tsv`. These handle some common renamings. However, 
-if your source has any properties that need to be renamed which are specific to that source, then you should create a 
-new file for that. While doing this, it might be useful to check any other `SOURCE-property-map.sssom.tsv` files to see 
-if they share any of the same property renamings. If so, it might be a good idea to move them to 
-`property-map.sssom.tsv` instead.
+- `rename --mappings config/SOURCE-property-map.sssom.tsv`: You'll notice that all components goals have a such a line 
+using `property-map.sssom.tsv`. This is for some common property renamings. However, if your source has any properties
```
joeflack4, line 59: ## Docs
The only reference of this file outside of the makefile was `add-new-source.md`, which I updated.

src/ontology/mondo-ingest.Makefile
```diff
@@ -63,8 +63,7 @@ $(TMPDIR)/ordo_relevant_signature.txt: component-download-ordo.owl | $(TMPDIR)
 #  - #60 needs to be fixed at source, but a workaround can probably be implemented in this goal
 $(COMPONENTSDIR)/omim.owl: $(TMPDIR)/omim_relevant_signature.txt | component-download-omim.owl
 	if [ $(COMP) = true ]; then $(ROBOT) remove -i $(TMPDIR)/component-download-omim.owl.owl --select imports \
-		rename --mappings config/property-map-1.sssom.tsv --allow-missing-entities true \
-		rename --mappings config/property-map-2.sssom.tsv --allow-missing-entities true \
+		rename --mappings config/property-map.sssom.tsv --allow-missing-entities true --allow-duplicates true \
```
joeflack4, line 66: ## Updated components goals
Updated all the components goals like this. Consolidated to just a single line, using the new filename and adding `--allow-duplicates true`.

joeflack4, line 66: ## About running the build
I ran the component and it passes. There should be no difference in any outputs as a result of this code. As such, I would prefer to merge this into develop without running the build. After all, we will always run the build on develop before merging to main, and in the very unlikely event of any issue caused by this, it would be caught at that time.

matentzn (commented):
I like this change very much but if you merge into dev before testing the pipeline, you may not be able to separate the resulting problems from other code merged into dev. This is a possibly impactful change (hopefully not)!

src/ontology/mondo-ingest.Makefile
twhetzel: I thought we decided to leave the use of the two property maps as-is to not add a testing burden to improve something that is not broken. 

If this is merged into `develop`, then yes the full build needs to be run.

joeflack4: Ah yes, we did discuss that. But circumstances changed a bit.

In #482, the suggested fix for combining these was this, which we decided not to do.
> Merge all mappings into 1 file, and have a second, empty file (to get around ROBOT issue)

While discussing that issue, we also decided that if there is a source that has property mappings not shared by any source, we can have a separate file for that; that's what I did for `icd11.foundation`, so we left that one as-is as well.

But then, Nico asked me to make this issue:
- https://github.com/ontodev/robot/issues/1195

Then during that, we discovered that the proposed flag already exists, and [Nico suggested](https://github.com/ontodev/robot/issues/1195#issuecomment-2066784275) I could potentially test it out to see if it worked.

joeflack4: I'll go ahead and run the build for this, then before merging into `develop`.

I don't mean to belabor or offend, but I just wasn't sure if you or Nico considered these additional negative aspects of running builds for all such small PRs, so I want to share:
i. This will discourage me from bothering with otherwise small and quick PRs.
  - In a way that has the benefit that I'll be more focused on bigger, more involved issues.
ii. This is an incentive for me to look for reasons, if possible, to combine small issues into a single PR. But this against the way we do things in Mondo, and a general good practice, to make single PRs for each separate concern, even for small e.g. single line changes.

joeflack4: Build success!
Let me know if you want me to upload the build result data files in a separate PR (I think we're doing that step as part of merging into main, and for other large PRs, but i can do it for every PR as well).

matentzn: If you could make a throw-away PR that would help!

joeflack4: Here we are!
- #505

docs/developer/add-new-source.md
twhetzel: Ok, and I see you renamed `property-map-2.sssom.tsv` to `property-map.sssom.tsv` and moved over the properties 👍

src/ontology/mondo-ingest.Makefile
twhetzel: The data files look good in the other PR.

twhetzel (approved):
looks good with these changes and the data PR

Issue #400, opened 2024-01-25
issue with Add excluded subclass of assertions for ICD10


nicolevasilevsky:
Related to: https://github.com/monarch-initiative/mondo/pull/7111

The output of this has a few mistakes, as you can see in the QC report: 
https://github.com/monarch-initiative/mondo/actions/runs/7432995102/job/20225429551?pr=7111

there are cases where a class is a subclass of itself and these should be removed:

1. MONDO:0005044	hypertensive disorder
2. MONDO:0005178	osteoarthritis
3. MONDO:0005385	vascular disorder

joeflack4:
Will look into this!

joeflack4:
Hey @nicolevasilevsky. Finally getting around to this one! Do you know if this is still an open issue? I'm unfamiliar with this part of our stack, so I'm struggling to think of where to look into this problem.

Maybe I've looked in the right places and the problem has simply resolved itself somehow? Can you confirm?

I started by looking in `mondo-edit.obo` and the latest release for `mondo.owl`. But I didn't find any instances of these classes subclassing themselves.

<details><summary>MONDO:0005044</summary>
<p>

```obo
[Term]
id: MONDO:0005044
name: hypertensive disorder
def: "Persistently high systemic arterial blood pressure. Based on multiple readings (blood pressure determination), hypertension is currently defined as when systolic pressure is consistently greater than 140 mm Hg or when diastolic pressure is consistently 90 mm Hg or more." [MESH:D006973]
synonym: "blood pressure, high" EXACT [NCIT:C3117]
synonym: "blood pressure, increased" EXACT [NCIT:C3117]
synonym: "high blood pressure" EXACT [DOID:10763]
synonym: "HTN" BROAD ABBREVIATION [DOID:10763]
synonym: "hyperpiesia" EXACT [DOID:10763]
synonym: "hypertension" RELATED [DOID:10763]
synonym: "hypertensive disease" EXACT [DOID:10763, MTH:NOCODE]
synonym: "increased blood pressure" EXACT [NCIT:C3117]
synonym: "pressure, high blood" EXACT [NCIT:C3117]
synonym: "vascular hypertensive disorder" EXACT [DOID:10763, NCIT:C3117]
xref: DOID:10763 {source="MONDO:equivalentTo", source="EFO:0000537"}
xref: EFO:0000537 {source="DOID:10763", source="MONDO:equivalentTo"}
[...]

PR #501, opened 2024-04-28
Subclass Sync: Self-parentage

joeflack4:
Partially addresses:
- https://github.com/monarch-initiative/mondo-ingest/issues/400

## Overview
This PR introduces reports and/or solutions for situations where the "subclass sync 'added'" pipeline is generating suggested relationships where the Mondo ID is the same for both subject and object.

### Documentation
- [ ] Yes

### QC
- [x] Yes
   
Build / data PR:
- #524

### New Packages
- [x] No

- [x] No

- [x] Yes

## Changs
Subclass Sync: Self-parentage
- An issue was reported https://github.com/monarch-initiative/mondo-ingest/issues/400 where the subclass-sync pipeline was generating robot templates with erroneous cases of relationship edges to be added into Mondo, where the subject and object were the same Mondo ID.
- Add: sync-subClassOf.added.self_parentage_issues.tsv and code to generate it.

### DIFF
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -511,33 +511,30 @@ slurp-all: $(foreach n,$(ALL_COMPONENT_IDS), slurp-$(n))
 sync: sync-subclassof
 
 .PHONY: sync-subclassof
-sync-subclassof: $(REPORTDIR)/sync-subClassOf.added.self_parentage_issues.tsv
-
-$(REPORTDIR)/sync-subClassOf.added.self_parentage_issues.tsv: tmp/mondo.sssom.tsv
-	rm -f $@
-	$(MAKE) $(REPORTDIR)/sync-subClassOf.direct-in-mondo-only.tsv
-	python3 $(SCRIPTSDIR)/sync_subclassof_collate_self_parentage.py \
-	--mondo-mappings-path $(TMPDIR)/mondo.sssom.tsv \
-	--in-outpath $@
+sync-subclassof: $(REPORTDIR)/sync-subClassOf.confirmed.tsv $(REPORTDIR)/sync-subClassOf.direct-in-mondo-only.tsv $(TMPDIR)/sync-subClassOf.added.self-parentage.tsv
 
 # todo: drop this? This is really just an alias here for quality of life, but not used by anything.
 .PHONY: sync-subclassof-%
-sync-subclassof-%: $(REPORTDIR)/%.subclass.direct-in-mondo-only.tsv
-	@echo "$@ completed"
+sync-subclassof-%: $(REPORTDIR)/%.subclass.confirmed.robot.tsv
+
+$(TMPDIR)/sync-subClassOf.added.self-parentage.tsv: $(foreach n,$(ALL_COMPONENT_IDS), $(TMPDIR)/$(n).subclass.self-parentage.tsv) tmp/mondo.sssom.tsv
+	python3 $(SCRIPTSDIR)/sync_subclassof_collate_self_parentage.py \
+	--mondo-mappings-path $(TMPDIR)/mondo.sssom.tsv \
 
 # Side effects: Deletes SOURCE.subclass.direct-in-mondo-only.tsv's from which the combination is made.
-$(REPORTDIR)/sync-subClassOf.direct-in-mondo-only.tsv: $(foreach n,$(ALL_COMPONENT_IDS), sync-subclassof-$(n)) tmp/mondo.db
+$(REPORTDIR)/sync-subClassOf.direct-in-mondo-only.tsv: $(foreach n,$(ALL_COMPONENT_IDS), $(REPORTDIR)/$(n).subclass.direct-in-mondo-only.tsv) tmp/mondo.db
 	python3 $(SCRIPTSDIR)/sync_subclassof_collate_direct_in_mondo_only.py --outpath $@
 
 $(REPORTDIR)/sync-subClassOf.confirmed.tsv: $(foreach n,$(ALL_COMPONENT_IDS), $(REPORTDIR)/$(n).subclass.confirmed.robot.tsv)
 	awk '(NR == 1) || (NR == 2) || (FNR > 2)' $(REPORTDIR)/*.subclass.confirmed.robot.tsv > $@
 
-$(REPORTDIR)/%.subclass.added.robot.tsv $(REPORTDIR)/%.subclass.confirmed.robot.tsv $(REPORTDIR)/%.subclass.direct-in-mondo-only.tsv $(REPORTDIR)/%.subclass.added-obsolete.robot.tsv: tmp/mondo-ingest.db tmp/mondo.db tmp/mondo.sssom.tsv
+$(REPORTDIR)/%.subclass.confirmed.robot.tsv $(REPORTDIR)/%.subclass.added.robot.tsv $(REPORTDIR)/%.subclass.added-obsolete.robot.tsv $(REPORTDIR)/%.subclass.direct-in-mondo-only.tsv $(TMPDIR)/%.subclass.self-parentage.tsv: tmp/mondo-ingest.db tmp/mondo.db tmp/mondo.sssom.tsv
 	python3 $(SCRIPTSDIR)/sync_subclassof.py \
 	--outpath-added $(REPORTDIR)/$*.subclass.added.robot.tsv \
 	--outpath-added-obsolete $(REPORTDIR)/$*.subclass.added-obsolete.robot.tsv \
 	--outpath-confirmed $(REPORTDIR)/$*.subclass.confirmed.robot.tsv \
 	--outpath-direct-in-mondo-only $(REPORTDIR)/$*.subclass.direct-in-mondo-only.tsv \
+	--outpath-self-parentage $(TMPDIR)/$*.subclass.self-parentage.tsv \
 	--mondo-db-path $(TMPDIR)/mondo.db \
 	--mondo-ingest-db-path $(TMPDIR)/mondo-ingest.db \
 	--mondo-mappings-path $(TMPDIR)/mondo.sssom.tsv \
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -511,7 +511,14 @@ slurp-all: $(foreach n,$(ALL_COMPONENT_IDS), slurp-$(n))
 sync: sync-subclassof
 
 .PHONY: sync-subclassof
-sync-subclassof: $(REPORTDIR)/sync-subClassOf.direct-in-mondo-only.tsv
+sync-subclassof: $(REPORTDIR)/sync-subClassOf.added.self_parentage_issues.tsv
+
+$(REPORTDIR)/sync-subClassOf.added.self_parentage_issues.tsv: tmp/mondo.sssom.tsv
+	rm -f $@
+	$(MAKE) $(REPORTDIR)/sync-subClassOf.direct-in-mondo-only.tsv
+	python3 $(SCRIPTSDIR)/sync_subclassof_collate_self_parentage.py \
+	--mondo-mappings-path $(TMPDIR)/mondo.sssom.tsv \
+	--in-outpath $@
 
 # todo: drop this? This is really just an alias here for quality of life, but not used by anything.
 .PHONY: sync-subclassof-%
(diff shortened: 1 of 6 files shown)


src/ontology/reports/sync-subClassOf.added.self_parentage_issues.tsv
```diff
@@ -0,0 +1,83 @@
+subject_mondo_id	subject_mondo_label	object_mondo_id	subject_source_id	object_source_id	object_mondo_label
```
joeflack4, line 1: ## New file: `reports/sync-subClassOf.added.self_parentage_issues.tsv`
This file collects all cases of offending suggested added edges where the same Mondo ID is both the subject and object (see: https://github.com/monarch-initiative/mondo-ingest/issues/400).

Here's a: [google sheet](https://docs.google.com/spreadsheets/d/1kjrzBGxGIirsMbHQWde3u8Z6YZRL0FyUSTM3eDWvEfI/edit#gid=324968868)

### Alternative approaches to this report
a. Keep producing this file as-is, w/ the current implementation.
b. Keep producing this file, but I could change the implementation a bit. Fore example, `sync-subClassOf.direct-in-mondo-only.tsv`is a similar case where a number of sources are included in the same file, but it gets created in a different way.
c. Keep producing, but 1 file for each source instead of / in addition to this single file collating all cases.
d. Keep producing, but add these file(s) to the `.gitignore`
e. No output files; just throw an error whenever this happens.

Optionally, we could go with one of (a) through (d), and also trigger an error to be thrown from the makefile if it detects any cases after it has run subclass sync for all sources.

joeflack4, line 1: ## Suggested action plan
- [ ] @matentzn @twhetzel @nicolevasilevsky Review and share thoughts

I think this [google sheet](https://docs.google.com/spreadsheets/d/1kjrzBGxGIirsMbHQWde3u8Z6YZRL0FyUSTM3eDWvEfI/edit#gid=324968868) needs to go through curation.

Copy/paste from [my comments](https://github.com/monarch-initiative/mondo-ingest/issues/400#issuecomment-2081684554) (I suggest taking a look as they have more detail) in https://github.com/monarch-initiative/mondo-ingest/issues/400:
> I think there multiple types of cases going on here:
> 1. In some cases, the suggestion from the "subclass sync - add" pipeline is correct, it might be that there are Mondo classes which need to get broken into 2 classes ;(
> 2. In other (hopefully most) cases, the suggestions are incorrect, and the pipeline needs to not make such suggestions.

> The reason that these suggestions are occurring is because they are instances where there is a legal `sub subClassOf obj` relationship in the source, but Mondo has done a proxy merge, so the same Mondo ID is mapped to both `sub` and `obj`.


> I think we need to go through the [google sheet](https://docs.google.com/spreadsheets/d/1kjrzBGxGIirsMbHQWde3u8Z6YZRL0FyUSTM3eDWvEfI/edit#gid=324968868) and parse out which of these cases that the pipeline is suggestion are correct and which are not. And I will probably need a file which lists all of Mondo's proxy merges so that I can have the pipeline read it and filter such suggestions out.

matentzn: Nice analysis. TBH, it does reveal some problems, no doubt,

I think we should go with (e) but write an info or warn level message into the log in case some debugging is needed.

1. Proxy merges are manually reviewed in Mondo and should therefore be trusted (even if its sometimes questionable - that is another issue). Proxy merges should be reviewed only once (when they are created), and after that, trusted and not reviewed again until a user complains.
1. The script should explicitly exclude subclass statement between classes that are proxy merged (simple filter as you suggest). Fee free to add a logger.warning() just so things like this can be more easily debugged.

I kinda like your table to just be silently generated so it can be reviewed if needed, but it does create churn. Trish should decide if it is needed. Alternatively you could make a one-off review of the 80 or so rows you found and request a curator to confirm the proxy merges once. Again, more churn, so again, Trish decision.

joeflack4: Sounds good to me! So I'm going to:
- [ ] 1. Filter these cases from the 'subclass sync - added' output TSV
- [ ] 2. `logger.warning()` when the cases arise
- [ ] 3. Await @twhetzel's determination on your last paragraph in the comment above, but for now will:
  - Move the output from `reports/` to the `.gitignore`'ed `tmp/` dir
  - Opt for no one-off review

joeflack4: [Nico opted not](https://github.com/monarch-initiative/mondo-ingest/pull/501#discussion_r1598101014) for this curation step.

src/scripts/sync_subclassof.py
```diff
@@ -410,6 +413,13 @@ def sync_subclassof(
     df3_obs.to_csv(outpath_added_obsolete, sep='\t', index=False)
     df3 = pd.concat([pd.DataFrame(subheader), df3])
     df3.to_csv(outpath_added, sep='\t', index=False)
+    # - issue: Mondo IDs subClassed to themselves. See: https://github.com/monarch-initiative/mondo-ingest/issues/400
+    df3_self_parentage = df3[df3['subject_mondo_id'] == df3['object_mondo_id']]
+    sp_err = f'Classes of self-parentage found for source {source_name}. This is a low urgency warning, as these are ' \
+             f'with almost 100% certainty caused by legal Mondo proxy merges.'
```
joeflack4, line 419: Warning message
- [ ] @matentzn @twhetzel review and suggest any changes. I wasn't sure how to phrase.

src/scripts/sync_subclassof_collate_self_parentage.py
```diff
@@ -0,0 +1,84 @@
+"""Sync SubClass: Report for Mondo classes having themselves as parents in the "added" template (legal proxy merges)."""
```
joeflack4, line 1: ## Refactor
When I made this report, I did it quick and dirty because I wasn't sure it'd be valued as a permanent output. Nico likes it, but in order to keep it, a refactor is necessary. I think if Nico realized this, maybe he wouldn't have suggested keeping it (as a 'silent' report (e.g. put in `tmp/` and don't throw an error). But I didn't have the foresight to think that Nico would suggest this, and I didn't realize until I'm halfway through this refactor that it actually takes a couple hours. But I'm more than halfway through now.

src/scripts/sync_subclassof.py
matentzn: The phrasing is not wrong, no strong opinions

src/scripts/sync_subclassof_collate_self_parentage.py
joeflack4: The refactor is now done.

src/ontology/mondo-ingest.Makefile
```diff
@@ -516,28 +516,38 @@ slurp-all: $(foreach n,$(ALL_COMPONENT_IDS), slurp-$(n))
...
 # todo: Until ODK has oaklib >= 0.5.20, need '$(MAKE) pip-oaklib' below https://github.com/INCATools/ontology-development-kit/issues/936
-$(REPORTDIR)/%.subclass.added.robot.tsv $(REPORTDIR)/%.subclass.confirmed.robot.tsv $(REPORTDIR)/%.subclass.direct-in-mondo-only.tsv $(REPORTDIR)/%.subclass.added-obsolete.robot.tsv: tmp/mondo-ingest.db tmp/mondo.db tmp/mondo.sssom.tsv
+$(REPORTDIR)/%.subclass.confirmed.robot.tsv $(REPORTDIR)/%.subclass.added.robot.tsv $(REPORTDIR)/%.subclass.added-obsolete.robot.tsv $(REPORTDIR)/%.subclass.direct-in-mondo-only.tsv $(TMPDIR)/%.subclass.self-parentage.tsv: tmp/mondo-ingest.db tmp/mondo.db tmp/mondo.sssom.tsv
 	$(MAKE) pip-oaklib
 	python3 $(SCRIPTSDIR)/sync_subclassof.py \
 	--outpath-added $(REPORTDIR)/$*.subclass.added.robot.tsv \
 	--outpath-added-obsolete $(REPORTDIR)/$*.subclass.added-obsolete.robot.tsv \
 	--outpath-confirmed $(REPORTDIR)/$*.subclass.confirmed.robot.tsv \
 	--outpath-direct-in-mondo-only $(REPORTDIR)/$*.subclass.direct-in-mondo-only.tsv \
+	--outpath-self-parentage $(TMPDIR)/$*.subclass.self-parentage.tsv \
```
joeflack4, line 550: ## Build
- [ ] @joeflack4 Run build

src/ontology/mondo-ingest.Makefile
```diff
@@ -516,28 +516,38 @@ slurp-all: $(foreach n,$(ALL_COMPONENT_IDS), slurp-$(n))
...
+	@rm -f $(TMPDIR)/sync-subClassOf.added.self_parentage_issues.tsv
+	$(MAKE) $(REPORTDIR)/sync-subClassOf.confirmed.tsv
+	$(MAKE) $(REPORTDIR)/sync-subClassOf.direct-in-mondo-only.tsv
+	$(MAKE) $(TMPDIR)/sync-subClassOf.added.self-parentage-issues.tsv
 
 # todo: drop this? This is really just an alias here for quality of life, but not used by anything.
 .PHONY: sync-subclassof-%
-sync-subclassof-%: $(REPORTDIR)/%.subclass.direct-in-mondo-only.tsv
-	@echo "$@ completed"
+sync-subclassof-%: $(REPORTDIR)/%.subclass.confirmed.robot.tsv
```
joeflack4, line 529: ## Other `sync-subclassof`-related `makefile` changes
While adding the goal for `$(TMPDIR)/sync-subClassOf.added.self-parentage.tsv`, the way I had things set up made less sense, and I noticed some things w/ how the dependency structure, etc, were set up that was just generally confusing. I updated several of these goals in several ways  to make the setup easier to understand and maintain

I also added some minor fixes / optimizations. For example, some `rm -f` statements that fix edge cases that can happen if you run / re-run parts of this pipeline locally, as opposed to a fresh build. I also changed one of the goal prereqs from a `.PHONY` alias goal to the actual goal represented by that alias, which in certain situations should prevent this goal from running unnecessarily. This also comes up only in local development / non-build context.

src/ontology/mondo-ingest.Makefile
joeflack4: Build success.

joeflack4:
All issues / threads resolved. Build complete.

src/ontology/mondo-ingest.Makefile
```diff
@@ -511,26 +511,36 @@ slurp-all: $(foreach n,$(ALL_COMPONENT_IDS), slurp-$(n))
 sync: sync-subclassof
 
 .PHONY: sync-subclassof
-sync-subclassof: $(REPORTDIR)/sync-subClassOf.direct-in-mondo-only.tsv
+sync-subclassof: $(foreach n,$(ALL_COMPONENT_IDS), $(REPORTDIR)/$(n).subclass.confirmed.robot.tsv)
+	@rm -f $(REPORTDIR)/sync-subClassOf.confirmed.tsv
+	@rm -f $(REPORTDIR)/sync-subClassOf.direct-in-mondo-only.tsv
+	@rm -f $(TMPDIR)/sync-subClassOf.added.self_parentage_issues.tsv
```
matentzn, line 517: This is a bit idiosyncratic (using rm), but I guess it safes you from using `-B` which could cause a cascade of updates.

src/ontology/mondo-ingest.Makefile
matentzn: Folding this `rm` pattern into the make pipelines itself should be done sparingly as it undermines the whole point of using a makefile: being able to avoid rebuilding files if none of the dependencies have changed. 

using `rm` pattern makes sense in development, not in production IMO, but just take it under advisement, not as a blocker for this PR.

matentzn (approved):
LGTM (meaning, as always, that I scrolled through the PR to look for bigger anti-patterns, not do a line by line combing.

Please be very careful with adding `rm` into your make goals  into production pipeline as this defeats the whole purpose of `make`.

nicolevasilevsky:
@joeflack4 I don't have expertise to review this PR but I'm okay with it if Nico and Trish approve.

src/ontology/mondo-ingest.Makefile
joeflack4: Yeah, I suppose I should have added a comment about this. I'm struggling to remember any reason why I did this other than what you mention.

<details><summary>Trying to remember</summary>
<p>

When I think about it, I can only think of me forgetting in the moment that when prereqs are newer than the target, it will rebuild the target, so this forces them to be rebuilt without running `-B`. Or, this might have been because I was originally building the parentage filed by appending to it each time I ran the subclass sync file on a given source, so I needed to remove it first, and I set up the pattern here to be the same for all 3 files. I've since changed it anyway; no longer doing the append thing.

</p>
</details> 

And also, maybe that original append design (which I'm no longer doing) reason is why I changed this goal so that these 3 things are inside the body rather than prereqs. I've just changed it to this:

```make
sync-subclassof: $(REPORTDIR)/sync-subClassOf.confirmed.tsv $(REPORTDIR)/sync-subClassOf.direct-in-mondo-only.tsv $(TMPDIR)/sync-subClassOf.added.self-parentage.tsv
```

src/ontology/mondo-ingest.Makefile
joeflack4: I thought about this and responded / made a change. See: [comment](https://github.com/monarch-initiative/mondo-ingest/pull/501#discussion_r1622929145)

PR #503, opened 2024-04-29
Remove duplicate prefix maps in `prefixes.csv`

joeflack4:
## Overview
Removed some duplicate entries (same prefix and URI prefix) in `prefixes.csv`.

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [x] No

- [x] No

- [ ] Yes

## Changes
- Bug fix: prefixes.csv: Removed some duplicate entries

### DIFF
--- a/src/ontology/config/prefixes.csv
+++ b/src/ontology/config/prefixes.csv
@@ -112,7 +112,6 @@ FOODON,http://purl.obolibrary.org/obo/FOODON_
 REX,http://purl.obolibrary.org/obo/REX_
 FOVT,http://purl.obolibrary.org/obo/FOVT_
 KISAO,http://purl.obolibrary.org/obo/KISAO_
-FBcv,http://purl.obolibrary.org/obo/FBcv_
 MAMO,http://purl.obolibrary.org/obo/MAMO_
 FBdv,http://purl.obolibrary.org/obo/FBdv_
 SBO,http://purl.obolibrary.org/obo/SBO_
@@ -243,14 +242,12 @@ HGNC,http://identifiers.org/hgnc/
 MEDDRA,http://identifiers.org/meddra/
 MEDGEN,http://identifiers.org/medgen/
 SCTID,http://identifiers.org/snomedct/
-UMLS,http://linkedlifedata.com/resource/umls/id/
 orcid,https://orcid.org/
 swrl,http://www.w3.org/2003/11/swrl#
 semapv,https://w3id.org/semapv/vocab/
 HGNC_SYMBOL,https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/
 HGNC,https://identifiers.org/hgnc/
 ncbi.gene,https://www.ncbi.nlm.nih.gov/gene/
-OMIMPS,https://omim.org/phenotypicSeries/PS
 STY,http://purl.bioontology.org/ontology/STY/
 sssom,https://w3id.org/sssom/
 biolink,https://w3id.org/biolink/vocab/


src/ontology/config/prefixes.csv
```diff
@@ -112,7 +112,6 @@ FOODON,http://purl.obolibrary.org/obo/FOODON_
 REX,http://purl.obolibrary.org/obo/REX_
 FOVT,http://purl.obolibrary.org/obo/FOVT_
 KISAO,http://purl.obolibrary.org/obo/KISAO_
-FBcv,http://purl.obolibrary.org/obo/FBcv_
```
joeflack4, line 115: ## The changes
There are 3 such cases of duplicates: Same prefix and URI prefix. From what I understand, removing these should have no effective change on outputs or runtime. Should just result in a codestyle fix.

src/ontology/config/prefixes.csv
```diff
@@ -243,14 +242,12 @@ HGNC,http://identifiers.org/hgnc/
 MEDDRA,http://identifiers.org/meddra/
 MEDGEN,http://identifiers.org/medgen/
 SCTID,http://identifiers.org/snomedct/
-UMLS,http://linkedlifedata.com/resource/umls/id/
```
matentzn, line 246: Not a duplicate. (different IRI, see my comment in issue)

src/ontology/config/prefixes.csv
matentzn: Ok, duplicate!

src/ontology/config/prefixes.csv
```diff
@@ -243,14 +242,12 @@ HGNC,http://identifiers.org/hgnc/
...
 orcid,https://orcid.org/
 swrl,http://www.w3.org/2003/11/swrl#
 semapv,https://w3id.org/semapv/vocab/
 HGNC_SYMBOL,https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/
 HGNC,https://identifiers.org/hgnc/
 ncbi.gene,https://www.ncbi.nlm.nih.gov/gene/
-OMIMPS,https://omim.org/phenotypicSeries/PS
```
matentzn, line 253: Not a duplicate (different IRI, see comment in issue)

matentzn (commented):
I understand your motivation (bring more order to the chaos) but I feel unqualified to understand the consequences of the proposed changes - your changes are _probably_ ok, but who knows what strange thing depends on this we didnt even know. Its ridiculous I know, but this could be a very hypothetical scenario where things could go wrong:

1. ORDO uses skos exact match to map to OMIM PURLS using the `https://omim.org/phenotypicSeries/PS` URI prefix
2. We rely on that value for lexmatch
3. We will miss out on ORDO -> OMIMPS mappings in the future

Not saying NOOOOO, just advising caution

src/ontology/config/prefixes.csv
joeflack4: Actually it is, and it isn't--but importantly--it is! See below (3 entries):

https://github.com/monarch-initiative/mondo-ingest/blob/adc201572f76b24fa80ff907ee726932249d9b87/src/ontology/config/prefixes.csv#L240-L246

src/ontology/config/prefixes.csv
joeflack4: This one is a dupe!:
https://github.com/monarch-initiative/mondo-ingest/blob/adc201572f76b24fa80ff907ee726932249d9b87/src/ontology/config/prefixes.csv#L234
https://github.com/monarch-initiative/mondo-ingest/blob/adc201572f76b24fa80ff907ee726932249d9b87/src/ontology/config/prefixes.csv#L253

joeflack4: I looked at [your comment](https://github.com/monarch-initiative/mondo-ingest/pull/503#pullrequestreview-2028050557), and I think it is invalidated now.

@matentzn Can you review / confirm that these are indeed pure duplicate rows?

Let me know, even that being the case, if you'd like me to run a build. And, if so, then I wonder if it might make sense to combine benign changes like these into a single PR so I don't have to run as many builds.

matentzn: Yep, the problem was that github preview uses lazy load; i just opened the file, hit browser search, didnt find the string and said "no dupe". If I would have scrolled, I would have seen it. Sorry!

matentzn (approved):

joeflack4:
@twhetzel Despite Nico's comment above, he approved this because at first he thought I was removing non-duplicate cases.

I marked 'no' on QC because there should be no effective changes here that should make any difference to any code, so I'm going to merge now!

PR #510, opened 2024-05-02
Create ORDO subset ROBOT template

### LINKED ISSUE #509: Create ORDO subset ROBOT template
This is a bit ad hoc, but we need it asap to fix the RD subset before the next release (in agreement with what we discussed with Orphanet). 

https://www.ebi.ac.uk/ols4/ontologies/ordo

I need using the raw  ORDO source, not mondo-ingest-ordo component (which has a different classification) and the currently up to date mondo.sssom.tsv.

A file with the following columns:

- mondo_id (from mondo.sssom.tsv - we dont care about cases where there is no Mondo id)
- ordo_id
- subset

where subset has one of the following values

- `http://purl.obolibrary.org/obo/mondo#ordo_group_of_disorders` (if its a Group of disorders, see screenshot)
- `http://purl.obolibrary.org/obo/mondo#ordo_clinical_subtype` (if its a Subtype)
- `http://purl.obolibrary.org/obo/mondo#ordo_disease` (if it is a disease/disorder).

Ideally robot template but anything goes. Thank you!

image

For now we can just generate this list in an adhoc manner, and I can make a more detailed ticket to add it to the release products.

Thank you!

### LINKED ISSUE #511: Create ORDO subset ROBOT template: Update goal to depend on component
Based on the work to update the subsets in Mondo based on information from ORDO, this question about whether to update the URL for the `component-download` came up. See https://github.com/monarch-initiative/mondo-ingest/pull/510/files#r1588599701

Response from Nico:
---
Look at
```
$(COMPONENTSDIR)/ordo.owl: $(TMPDIR)/ordo_relevant_signature.txt config/properties.txt | component-download-ordo.owl
	if [ $(COMP) = true ]; then $(ROBOT) remove -i $(TMPDIR)/component-download-ordo.owl.owl --select imports \
		merge ...
```
It is better to:

depend in component-download-ordo.owl
And use $(TMPDIR)/component-download-ordo.owl.owl
to ensure we are using the same version as the rest of the pipeline

### LINKED ISSUE #515: Create ORDO subset ROBOT template - updates
## Overview
In #510, we successfully created a robot template that updated ORDO rare disease subsets. We will need to re-run this in the future, but there are a few updates that need to be made.

## Sub-tasks
- [x] 1. [Utilize `curies`](https://github.com/monarch-initiative/mondo-ingest/pull/510#discussion_r1588599257)
- [x] [2](https://github.com/monarch-initiative/mondo-ingest/pull/510#pullrequestreview-2039368784). `reports/ordo-subsets.robot.template.tsv` would be moved to where NORD external content is located (`src/ontology/external` I think) as we should consider this template "externally managed content"
- [x] [3](https://github.com/monarch-initiative/mondo-ingest/pull/510#pullrequestreview-2039368784). We can make sure its processed as part of the general `build-mondo-ingest` pipeline

joeflack4:
Resolves #509
Resolves #511
Resolves #515
May address #514

## Overview
Creates a ROBOT template to update mondo terms with the correct ORDO subset annotation.

See results: [google sheet](https://docs.google.com/spreadsheets/d/1UB0L62Rd2ZhqG20-3C5eNAtYeO9D-N2pe01e6LvRdRA/edit#gid=976194052&fvid=313621884)

### Documentation
- [ ] Yes

### QC
Build / data / qc PR: #543
- [x] Yes
   
### New Packages
- [x] No

- [x] No

- [x] Yes


src/ontology/mondo-ingest.Makefile
```diff
@@ -598,6 +598,22 @@ update-jinja-sparql-queries:
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/create_sparql__ordo_replace_annotation_based_mappings.py
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/create_sparql__ordo_mapping_annotations_violation.py
 
+
+#################
+##### Ad hoc ####
```
joeflack4, line 603: ## Where I put these goals in the makefile
I made this section. Wasn't sure where else to put it.

src/ontology/mondo-ingest.Makefile
```diff
@@ -598,6 +598,22 @@ update-jinja-sparql-queries:
...
+#################
+tmp/ORDO_en_4.4.owl:
+	wget https://www.orphadata.com/data/ontologies/ordo/last_version/ORDO_en_4.4.owl -O $@
+
+tmp/ordo-subsets.tsv: tmp/ORDO_en_4.4.owl
+	$(ROBOT) query -i $< --query ../sparql/select-ordo-subsets.sparql $@
+
+reports/ordo-subsets.robot.template.tsv: tmp/ordo-subsets.tsv tmp/mondo.sssom.tsv
```
joeflack4, line 611: ## Makefile setup
Dependency tree
- `reports/ordo-subsets.robot.template.tsv`: through Python script
  - `tmp/ordo-subsets.tsv`: through SPARQL query
    - `tmp/ORDO_en_4.4.owl`: through `wget`
  - `mondo.sssom.tsv`

src/scripts/ordo_subsets.py
```diff
@@ -0,0 +1,83 @@
+"""Creates a ROBOT template to update mondo terms with the correct ORDO subset annotation."""
```
joeflack4, line 1: ## Temporary Python scripting in `mondo-ingest`
We weren't sure where to do this work, but `mondo-ingest` seemed like the best place. However, we have a mandate not to proliferate Python scripts in this repo. We took a look at `mondolib`, but it's not setup well currently to add additional scripts.

Mainly I've done it this way for now because I'm on vacation from tomorrow and will be back on Tuesday, and so I had to do something quick and dirty.

- [ ] @matentzn Advise preferences on how to change this before we merge it

src/ontology/mondo-ingest.Makefile
joeflack4, line 611: ## Adding this file to `mondo` repo
@matentzn @twhetzel Sorry, I don't have time for this tonight and will be on vacation for the next few days. I might be able to do this tomorrow, but if not, @twhetzel can you do it?

- [ ] Add file to that repo? @twhetzel (maybe @joeflack4 if time)

src/scripts/ordo_subsets.py
```diff
@@ -0,0 +1,83 @@
...
+
+import pandas as pd
+
+from utils import remove_angle_brackets
+
+
+SUBSET_MAP = {
+    'Orphanet:557493': 'http://purl.obolibrary.org/obo/mondo#ordo_disease',  # disorder
+    'Orphanet:557492': 'http://purl.obolibrary.org/obo/mondo#ordo_group_of_disorders',  # group of disorders
+    'Orphanet:557494': 'http://purl.obolibrary.org/obo/mondo#ordo_clinical_subtype',  # subtype of a disorder
```
joeflack4, line 12: ## Class to mondo prop mappings
@matentzn Please review. Trish and I describe our interpretation of what we needed to do in [this comment](https://github.com/monarch-initiative/mondo-ingest/issues/509#issuecomment-2091660366).

src/scripts/ordo_subsets.py
```diff
@@ -0,0 +1,83 @@
...
+}
+ROBOT_TEMPLATE_HEADER = {
+    'mondo_id': 'ID',
+    'ordo_id': 'A oboInOwl:hasDbXref',
+    'subset': 'A oboInOwl:inSubset',
```
joeflack4, line 17: ## Robot subheader
- [ ] @matentzn @twhetzel I think this is right, but please review

src/scripts/ordo_subsets.py
```diff
@@ -0,0 +1,83 @@
...
+        df[fld] = df[fld].map(lambda x: x.replace('http://www.orpha.net/ORDO/Orphanet_', 'Orphanet:'))
+
+    # Map to mondo subset IDs
+    df['subset'] = df['subset_ordo_class_id'].map(SUBSET_MAP)
+
+    # JOIN to get Mondo IDs
+    df = df.merge(mondo_df, how='inner', left_on='ordo_id', right_on='object_id')
+
+    # Column order, column filtering, and sorting
+    df = df[['mondo_id', 'ordo_id', 'subset', 'mondo_label', 'ordo_label']]\
```
joeflack4, line 61: ## Extra columns: `mondo_label`, `ordo_label`
I left these columns because I felt someone might benefit from them. They have no robot subheader, so they should be safe to leave in.

src/scripts/ordo_subsets.py
```diff
@@ -0,0 +1,83 @@
...
+    df = pd.read_csv(class_subsets_tsv_path, sep='\t').rename(columns={
+        '?cls': 'ordo_id',
+        '?lbl': 'ordo_label',
+        '?grp': 'subset_ordo_class_id',
+        '?grp_lbl': 'subset_ordo_class_label',
+    }).sort_values(['ordo_id', 'ordo_label'])
+    # - remove duplicate rows where labels are a specific language variation, e.g. @en
+    #   this alternative works as of 2024/05/02, but is potentially trepidacious:
+    # df2 = df.drop_duplicates(subset=['ordo_id'], keep='last')
+    df['label_has_lang_variation'] = df['ordo_label'].map(lambda label: '@' in label)
```
joeflack4, line 43: FYI: Deduped. We did a query of all subclasses of each subset, and we got some 6 duplicate entries with `@en` labels

src/scripts/ordo_subsets.py
```diff
@@ -0,0 +1,83 @@
...
+    df = df[~df['label_has_lang_variation']]
+    del df['label_has_lang_variation']  # this gets dropped later, but w/e
+
+    # Preprocess
+    for fld in ['ordo_id', 'subset_ordo_class_id']:
+        # - Strip annoying angle brackets from URIs
+        df[fld] = df[fld].map(remove_angle_brackets)
+        # - Compress class IDs
+        df[fld] = df[fld].map(lambda x: x.replace('http://www.orpha.net/ORDO/Orphanet_', 'Orphanet:'))
```
joeflack4, line 52: ## Temporary lack of `curies`
Due to lack of time. I can update this pre or post merge.

src/scripts/ordo_subsets.py
```diff
@@ -0,0 +1,83 @@
...
+
+    # Map to mondo subset IDs
+    df['subset'] = df['subset_ordo_class_id'].map(SUBSET_MAP)
+
+    # JOIN to get Mondo IDs
+    df = df.merge(mondo_df, how='inner', left_on='ordo_id', right_on='object_id')
```
joeflack4, line 58: FYI: there were 237 ORDO IDs not in `mondo.sssom.tsv`, so those were dropped in this process.

src/sparql/select-ordo-subsets.sparql
```diff
@@ -0,0 +1,13 @@
+prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+
+SELECT ?cls ?lbl ?grp ?grp_lbl
+WHERE {
+    VALUES ?grp {
+        <http://www.orpha.net/ORDO/Orphanet_557493>  # disorder
+        <http://www.orpha.net/ORDO/Orphanet_557492>  # group of disorders
+        <http://www.orpha.net/ORDO/Orphanet_557494>  # subtype of a disorder
+    }
+    ?cls rdfs:subClassOf+ ?grp .
```
joeflack4, line 10: FYI: The SPARQL query

src/ontology/mondo-ingest.Makefile
```diff
@@ -598,6 +598,22 @@ update-jinja-sparql-queries:
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/create_sparql__ordo_replace_annotation_based_mappings.py
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/create_sparql__ordo_mapping_annotations_violation.py
 
+
+#################
+##### Ad hoc ####
+#################
+tmp/ORDO_en_4.4.owl:
+	wget https://www.orphadata.com/data/ontologies/ordo/last_version/ORDO_en_4.4.owl -O $@
```
joeflack4, line 606: ## `component-download` due for an update?
- [ ] @matentzn respond

Should I also update `mondo-ingest-odk.yaml` to use this URL instead?

src/scripts/ordo_subsets.py
```diff
@@ -0,0 +1,83 @@
...
+
+    # Column order, column filtering, and sorting
+    df = df[['mondo_id', 'ordo_id', 'subset', 'mondo_label', 'ordo_label']]\
+        .sort_values(['mondo_id', 'ordo_id'])
```
joeflack4, line 62: ## Sorting
- [ ] @matentzn @twhetzel Advise

If you guys like, I can sort on `subset` primarily instead

src/ontology/mondo-ingest.Makefile
joeflack4, line 611: ## Google sheet of template
See results: [google sheet](https://docs.google.com/spreadsheets/d/1UB0L62Rd2ZhqG20-3C5eNAtYeO9D-N2pe01e6LvRdRA/edit#gid=976194052&fvid=313621884)

src/scripts/ordo_subsets.py
twhetzel: PS - I decided the Python code for this work should just be added into this repo.

src/ontology/mondo-ingest.Makefile
twhetzel: @joeflack4 what needs to be done here?

src/scripts/ordo_subsets.py
twhetzel: No need to sort here since it's all going into an Excel file and can be sorted there if needed.

src/scripts/ordo_subsets.py
twhetzel: What needs updating here?

src/scripts/ordo_subsets.py
twhetzel: To clarify, the SPARQL results returned two rows where the IRI was the same and the labels varied by one including the `en` tag?

src/scripts/ordo_subsets.py
```diff
@@ -0,0 +1,83 @@
...
+    class_subsets_tsv_path: str, mondo_mappings_path: str, outpath: str,
+):
+    """Creates a ROBOT template to update mondo terms with the correct ORDO subset annotation."""
+    # Read: mondo.sssom.tsv
+    mondo_df = pd.read_csv(mondo_mappings_path, sep='\t', comment='#').rename(columns={
+        'subject_id': 'mondo_id',
+        'subject_label': 'mondo_label',
+    })
+    # - keep only exact matches
+    mondo_df = mondo_df[mondo_df['predicate_id'] == 'skos:exactMatch']
```
twhetzel, line 31: I'm not sure if the list of Mondo IDs to tag with RD tags should be filtered here.

twhetzel (commented):
Added a few clarifying questions.

@matentzn do the Mondo terms from the sssom file need to be filtered by predicate=skos:exactMatch? See Line 31

src/scripts/ordo_subsets.py
joeflack4: Oh, ok cool!

src/ontology/mondo-ingest.Makefile
joeflack4: So, I didn't see these instructions in the issue, but if this is anything like other stuff I've worked on, I imagine that this robot template will need to somehow get from `mondo-ingest` into `mondo`.

Hmm... there is a difference based on this and what I've worked on previously like this, though. Previously, I've done something such as put the file in the releases of a repo like `medgen`, and then I wrote a goal in `mondo` to `wget` it, and put it in some directory, and then I ran that and committed the file, and that was the whole PR--just the goal and the committed file.

But, since this is coming from `mondo-ingest`, and there's some sort of way that we normally get files from `mondo-ingest`'s codebase directly into `mondo`, maybe there's a different way to do this?
One thing we could do is, rather than putting it in a `mondo-ingest` release, we could merge this into `main`, and then the 'raw' file could be `wget`'ed directly; I hope I'm explaining this well enough. But that's an other option.

src/scripts/ordo_subsets.py
joeflack4: If the outputs of this script are good and we never need to run this again, then I don't think we strictly need to do anything here.

But if we want this code to be standardized, then instead of this find/replace I'm doing, I would pass the path to one of the `mondo-ingest` config files which contains `curie_map`, and then I would use the `curies` package here, something like fist instantiate a `converter` and then `lambda x: converter.collapse(x)`.

src/scripts/ordo_subsets.py
joeflack4: You mean the rows of `mondo.sssom.tsv` should not be filtered here? Perhaps. FYI, the only other predicate in the file was `skos:broadMatch`. I think that if we don't filter everything but `skos:exactMatch`, then we would need to somehow include the predicate also in the robot template, but IDK what would need to be done in that situation. My guess is it should only be exact match, but will see if Nico confirms.

src/ontology/mondo-ingest.Makefile
matentzn: This looks amazing!

src/scripts/ordo_subsets.py
matentzn: It is correct, they should be filtered by exact match. We need to be a bit conservative assigning the ORDO subsets - only to those classes that are identical with the Mondo classes!

src/ontology/mondo-ingest.Makefile
matentzn: Look at 

```
$(COMPONENTSDIR)/ordo.owl: $(TMPDIR)/ordo_relevant_signature.txt config/properties.txt | component-download-ordo.owl
	if [ $(COMP) = true ]; then $(ROBOT) remove -i $(TMPDIR)/component-download-ordo.owl.owl --select imports \
		merge ...
```

It is better to:

1. depend in `component-download-ordo.owl`
2. And use `$(TMPDIR)/component-download-ordo.owl.owl`

to ensure we are using the same version as the rest of the pipeline

src/scripts/ordo_subsets.py
matentzn: Agreed!

src/ontology/mondo-ingest.Makefile
matentzn: I am doing that part - dont worry!

src/scripts/ordo_subsets.py
matentzn: This makes sense to me!

src/scripts/ordo_subsets.py
matentzn: Technically this has to be

```
ROBOT_TEMPLATE_HEADER = {
    'mondo_id': 'ID',
    'subset': 'A oboInOwl:inSubset',
    'ordo_id': '>A oboInOwl:hasDbXref',
```

The first two columns ensure the subset is created, the third column is provided as evidence.

src/scripts/ordo_subsets.py
matentzn: ok. They are not needed, but they do no harm either.

src/scripts/ordo_subsets.py
matentzn: Yikes! Can this be turned in an issue and assigned to me?

src/scripts/ordo_subsets.py
matentzn: Just corrected above comment. `oboInOwl:source` not `oboInOwl:hasDbXref`

src/ontology/mondo-ingest.Makefile
twhetzel: all we need to do for this is to create the ROBOT file

src/ontology/mondo-ingest.Makefile
twhetzel: I moved this comment to a [new ticket](https://github.com/monarch-initiative/mondo-ingest/issues/511) since it's work not directly related to this PR, despite being informed by this PR

src/scripts/ordo_subsets.py
twhetzel: Whew, thank goodness :)

src/scripts/ordo_subsets.py
twhetzel: Perfect! I'll update the code just to be consistent.
** There is no need to re-run the code and re-generate the ROBOT file though. Simple copy-paste in the headers was already done.

src/scripts/ordo_subsets.py
twhetzel: Gotcha, thanks for that requirement!

src/scripts/ordo_subsets.py
twhetzel: Sure, https://github.com/monarch-initiative/mondo-ingest/issues/512

src/scripts/ordo_subsets.py
joeflack4: If it's not going to be used again, then we technically don't even have to merge it. But I think it's good to commit these ad hoc workflows somewhere.

Maybe we just don't know if we'll need to use this code again, or maybe we might borrow this code to do something similar, in which case we ideally do want the code to be high / standard quality.

That being said, it should take 5-10 minutes to update this properly.

matentzn: This code _will definitely be used again_, as the ORDO subset pipeline is integral now to how we identify what Orphanet considers a rare disease. This pipeline will be run basically once every month!

I did not clearly express myself when I asked for this issue: I wanted a quick and dirty solution (one off) so I can implement this pipeline before the next release, and give you the time to develop a proper solution in the future. However, you are already so close to a proper solution that I would just suggest to push it over the finish line now.

matentzn:
> @matentzn do the Mondo terms from the sssom file need to be filtered by predicate=skos:exactMatch? See Line 31

Yes! We only care about exact matches in this pipeline.

matentzn (commented):
- [x] I would mildly strongly recommend to move to the correct ORDO components as the source, see my comments in #511.

I would also like it if 

- [ ] `reports/ordo-subsets.robot.template.tsv` would be moved to where NORD external content is located (src/ontology/external I think) as we should consider this template "externally managed content"
- [ ] We can make sure its processed as part of the general "build-mondo-ingest" pipeline

src/scripts/ordo_subsets.py
twhetzel: To state here the Slack comment, Pause on making this the proper solution for now.

src/scripts/ordo_subsets.py
matentzn, line 1: Please exclude 

MONDO:0000001 | http://purl.obolibrary.org/obo/mondo#ordo_disease | Orphanet:377788 | disease | disease
-- | -- | -- | -- | --

From the ORDO subset - which is the "ORDO" disorder class itself. Adding this makes the entire ontology become "rare"!

matentzn (changes requested):

src/scripts/ordo_subsets.py
```diff
@@ -0,0 +1,84 @@
...
+    'Orphanet:557492': 'http://purl.obolibrary.org/obo/mondo#ordo_group_of_disorders',  # group of disorders
+    'Orphanet:557494': 'http://purl.obolibrary.org/obo/mondo#ordo_clinical_subtype',  # subtype of a disorder
+}
+ROBOT_TEMPLATE_HEADER = {
+    'mondo_id': 'ID',
+    'subset': 'A oboInOwl:inSubset',
+    'ordo_id': '>A oboInOwl:source',
+}
+
+
```
matentzn, line 20: Investigate why these four classes without mappings ended up in the subset:

| Mondo ID      | label                                                                                                                   |
|---------------|-------------------------------------------------------------------------------------------------------------------------|
| MONDO:0011965 | familial temporal lobe epilepsy 2                                                                                       |
| MONDO:0009349 | holoprosencephaly 1                                                                                                     |
| MONDO:0100340 | Friedreich ataxia 1                                                                                                     |
| MONDO:0024525 | Fanconi renotubular syndrome 1                                                                                          |

Mandy using and outdate mondo.sssom.tsv?

joeflack4: That's probably it. I forgot and didn't run with `-B` or delete `mondo.sssom.tsv` first.

src/scripts/ordo_subsets.py
joeflack4: Hmm, OK, I thought we set up our SPARQL query to not include these classes themselves. Not sure how that got in there, I'll take a look.

twhetzel: The gSheet does not include `MONDO:0000001`.

src/scripts/ordo_subsets.py
twhetzel: None of these 4 are in the gSheet.

matentzn: I deleted them from the gsheet for the pipeline processing

src/ontology/mondo-ingest.Makefile
```diff
@@ -598,6 +598,19 @@ update-jinja-sparql-queries:
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/create_sparql__ordo_replace_annotation_based_mappings.py
 	python3 $(SCRIPTSDIR)/ordo_mapping_annotations/create_sparql__ordo_mapping_annotations_violation.py
 
+
+#################
+##### Ad hoc ####
+#################
+tmp/ordo-subsets.tsv: $(COMPONENTSDIR)/ordo.owl
```
joeflack4, line 605: ## Updated goal to depend on component
Addresses:
- https://github.com/monarch-initiative/mondo-ingest/issues/511#issuecomment-2094305764

src/scripts/ordo_subsets.py
joeflack4: @matentzn I assume you also deleted this one from the google sheet too.

src/ontology/mondo-ingest.Makefile
joeflack4, line 605: ## Delete `ordo-construct-subsets.ru`?
Would like Nico or Trish to confirm that it is OK to handle this issue / particularly as part of this PR. Would be a quick change, once a reviewer looks into the issue / confirms the plan.

Addresses:
- #514

src/scripts/ordo_subsets.py
matentzn: yes I deleted all 5 of them.

src/ontology/mondo-ingest.Makefile
matentzn: Better not as part of this PR - this is a different concern: the shape of the component, vs what we directly import into Mondo.

src/scripts/ordo_subsets.py
joeflack4: @matentzn Trish and I investigated. The reason it's being included is because `Orphanet:377788` "Disease" is a direct subclass of `Orphanet:557493` "Disorder", which is one of the subsets that we've selected. Makes sense to me. But yes, implicit in that is that it should select the entire tree as rare.

Can you make a new issue with instructions on what, if anything should be done?
a. Do nothing; continue selecting as-is.
b. I can exclude this class explicitly in the Python code, or via passing some kind of exclusion config file specifically for this pipeline.
c. something else

src/scripts/ordo_subsets.py
joeflack4: @matentzn Looked into this. I got a fresh copy of `mondo.sssom.tsv`, and these 4 classes do indeed still have mappings. So there seems to be no issue w/ the code in this PR; they seem like they should show up in our subset results.

Just FYI if you want to think through this some more, I found the original comment where these 4 classes were discussed; idk if it might help to revisit/reread: [comment](https://github.com/monarch-initiative/mondo/pull/7681#issuecomment-2098193144)

matentzn: Did you get a fresh copy of mondo.sssom.tsv, or use the mondo-ingest pipeline to generate a fresh copy? I hope the latter! If the mondo.sssom.tsv mapping file still contains "exact matches" to these four terms, can you check if they also exist in mondo-edit.obo? Because I determined they dont, but I may have made a mistake somewhere along the way.

joeflack4: @matentzn My bad; I made a boo-boo. I did run the goal in `mondo-ingest`, but not correctly.
@twhetzel When we went over this, I skipped through it too quickly. Even though the goal for `mondo.sssom.tsv` does do `cp $(TMPDIR)/mondo/src/ontology/mappings/mondo.sssom.tsv $@`,the prerequisite for this goal does indeed run `make mondo.owl mappings`. During our checking, it was not sufficient to simply `rm tmp/mondo.sssom.tsv; make tmp/mondo.sssom.tsv`. We needed to just do `make tmp/mondo.sssom.tsv -B` to rerun the pipeline.

In the past, I've sometimes run this incorrectly, and sometimes have run it correctly. Basically I often don't use `-B` for a lot of goals because it often runs a long pipeline that is not actually necessary at the time. 

I'm re-running it now but it takes a while. But I expect that indeed, these 4 classes will not show up after I'm done. I don't see any ORDO xrefs on them in the `mondo.sssom.tsv` in my `mondo` repo, nor in `mondo-edit.obo`.

---

@matentzn suggestion:
In order to make this better dummy-proof UX for me / in general, what do you think about this change to the goals around `tmp/mondo_repo_built`?

<details><summary>Current goal setup</summary>
<p>

```make
tmp/mondo_repo_built:
	if [ $(USE_MONDO_RELEASE) = true ]; then wget http://purl.obolibrary.org/obo/mondo.owl -O $@; else cd $(TMPDIR) &&\
		rm -rf ./mondo/ &&\
		git clone --depth 1 https://github.com/monarch-initiative/mondo &&\
		cd mondo/src/ontology &&\
		make mondo.owl mappings -B MIR=false IMP=false MIR=false &&\
		cd ../../../../ &&\
		touch $@; fi

$(TMPDIR)/mondo.sssom.tsv: $(TMPDIR)/mondo_repo_built
	cp $(TMPDIR)/mondo/src/ontology/mappings/mondo.sssom.tsv $@

$(TMPDIR)/mondo.owl: $(TMPDIR)/mondo_repo_built
	cp $(TMPDIR)/mondo/src/ontology/mondo.owl $@
```

</p>
</details> 

<details><summary>Proposed changes</summary>
<p>

```make
$(TMPDIR)/mondo.sssom.tsv:
	if [ $(USE_MONDO_RELEASE) = true ]; then wget http://purl.obolibrary.org/obo/mondo.owl -O $@; else cd $(TMPDIR) &&\
		rm -rf ./mondo/ &&\
		git clone --depth 1 https://github.com/monarch-initiative/mondo &&\
		cd mondo/src/ontology &&\
		make mappings -B MIR=false IMP=false MIR=false &&; fi
	cp $(TMPDIR)/mondo/src/ontology/mappings/mondo.sssom.tsv $@

$(TMPDIR)/mondo.owl:
	if [ $(USE_MONDO_RELEASE) = true ]; then wget http://purl.obolibrary.org/obo/mondo.owl -O $@; else cd $(TMPDIR) &&\
		rm -rf ./mondo/ &&\
		git clone --depth 1 https://github.com/monarch-initiative/mondo &&\
		cd mondo/src/ontology &&\
		make mondo.owl -B MIR=false IMP=false MIR=false &&; fi
	cp $(TMPDIR)/mondo/src/ontology/mondo.owl $@
```

</p>
</details>

src/scripts/ordo_subsets.py
joeflack4: @matentzn Actually here's my suggestion (d): Change `MONDO:0000001 skos:exactMatch Orphanet:377788` to `skos:broadMatch`.

In Slack DM, @matentzn just told me that:
> ...they (Orphanet) only want the classes in the disease/disorder branch to be counted as rare.

The root of this disease/disorder branch is, I believe, `Orphanet:377788` "Disease". But if what you say is true, these are all rare diseases.
I think that either (a) the label on this should be "Rare disease", or (b) maybe the label is fine, but since Orphanet is about rare diseases, perhaps it is implicit that it should be interpreted to mean "rare disease".

I take it that this would have huge implications though. Any thoughts?

matentzn: I didnt realise "disease" was a subclass of "disorder"... In this case, I think the "correct" way is for you to not do anything here (after all, disease is, indeed, in the "disorder" subset), but create an issue and assign to me to exclude the term from the rare disease subset generation pipeline (with the logic that, while the term is, indeed in the ordo_disorder subset, it should not be added to the rare disease subset, which you are not concerned with here).

src/scripts/ordo_subsets.py
matentzn: These suggestions would break the point of `make` caching though, which in turn would make running the pipeline much slower if you have to run it repeatedly. My preferred way would be to solve this via SOP:

If you want to ensure the latest version of the mondo repo during your development process, you run:

```
rm tmp/mondo_repo_built
```

Then continue.

joeflack4: Ah, you are right. I guess I just have to remember this. I just don't like having to remember a lot of little things. But I suppose I don't see a way around it either. 

In that case if I am just on my own just trying to get mondo.sssom.tsv, I think -B will have the same effect.

src/scripts/ordo_subsets.py
joeflack4: Done!
- https://github.com/monarch-initiative/mondo/issues/7704

@twhetzel @matentzn marking this resolved now

src/ontology/mondo-ingest.Makefile
joeflack4, line 1: ## Ready for merge I think
@twhetzel I created an issue for all the outstanding things we'd still like to do w/ this pipeline:
- #515

So I think we can merge now. If you want we can do it at next meeting.

src/ontology/mondo-ingest.Makefile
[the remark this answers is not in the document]
joeflack4: ## Not merging is holding up another PR
@matentzn I made #531 to address #527, but I needed to use this branch as the base for that PR. So this PR needs to be merged first before we can merge that one.

src/scripts/ordo_subsets.py
```diff
@@ -0,0 +1,84 @@
+"""Creates a ROBOT template to update mondo terms with the correct ORDO subset annotation."""
+from argparse import ArgumentParser
+
+import pandas as pd
+
+from utils import remove_angle_brackets
+
+
+SUBSET_MAP = {
+    'Orphanet:557493': 'http://purl.obolibrary.org/obo/mondo#ordo_disease',  # disorder
```
matentzn, line 10:
```suggestion
    'Orphanet:557493': 'http://purl.obolibrary.org/obo/mondo#ordo_disorder',  # disorder
```

Wait for discussion in `#mondo-ingest` slack channel to conclude.

src/scripts/ordo_subsets.py
```diff
@@ -0,0 +1,84 @@
...
+    'Orphanet:557492': 'http://purl.obolibrary.org/obo/mondo#ordo_group_of_disorders',  # group of disorders
+    'Orphanet:557494': 'http://purl.obolibrary.org/obo/mondo#ordo_clinical_subtype',  # subtype of a disorder
```
matentzn, line 12:
```suggestion
    'Orphanet:557494': 'http://purl.obolibrary.org/obo/mondo#ordo_subtype_of_a_disorder',  # subtype of a disorder
```

Wait for discussion in `#mondo-ingest` slack channel to conclude before accepting this.

matentzn (changes requested):
Also still an open issue on the use of curies.

src/scripts/ordo_subsets.py
joeflack4: I take it that is [this Slack convo](https://monarchinitiative.slack.com/archives/C03AW548J8M/p1716486163402139).

src/scripts/ordo_subsets.py
joeflack4: Resolving this conversation as this will be handled in:
- #515

src/ontology/mondo-ingest.Makefile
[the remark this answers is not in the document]
joeflack4: @matentzn Re: [this comment](https://github.com/monarch-initiative/mondo-ingest/pull/510#pullrequestreview-2074654675):
> Also still an open issue on the use of `curies`.

As mentioned in my comment above in this thread, that issue is to be addressed by:
- #515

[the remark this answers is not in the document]
joeflack4: New build after renaming mondo subset URIs:
- #543
-

src/ontology/mondo-ingest.Makefile
```diff
@@ -560,7 +554,20 @@ $(EXTERNAL_CONTENT_DIR)/nord.robot.tsv: $(TMPDIR)/nord.tsv config/external-conte
...
+	python3 $(SCRIPTSDIR)/ordo_subsets.py \
+	--mondo-mappings-path tmp/mondo.sssom.tsv \
+	--class-subsets-tsv-path tmp/ordo-subsets.tsv \
+	--onto-config-path metadata/ordo.yml \
+	--outpath $@
+
+.PHONY: external-content-ordo
+external-content-ordo: $(EXTERNAL_CONTENT_DIR)/ordo-subsets.robot.template.tsv
+
+update-externally-managed-content: external-content-nord external-content-ordo
```
joeflack4, line 570: ## Added to build
In order to do #543, I needed to address 2/3 of the sub-tasks in #515:
- Add this output to the `external-content` part of the pipeline
- Which, in turn, adds this output to `build-mondo-ingest`

src/scripts/ordo_subsets.py
```diff
@@ -0,0 +1,93 @@
...
+    df['label_has_lang_variation'] = df['ordo_label'].map(lambda label: '@' in label)
+    df = df[~df['label_has_lang_variation']]
+    del df['label_has_lang_variation']  # this gets dropped later, but w/e
+
+    # Preprocess
+    for fld in ['ordo_id', 'subset_ordo_class_id']:
+        # - Strip annoying angle brackets from URIs
+        df[fld] = df[fld].map(remove_angle_brackets)
+        # - Compress class IDs
+        df[fld] = df[fld].map(lambda x: conv.compress(x))
```
joeflack4, line 57: ## Utilized `curies`
Since I needed to do 2/3 of the tasks in #515, I went ahead and did the last one, adding `curies`.

I also did a few other very tiny things that don't result in changes in outputs; comments and moving the location of some code. Easier to review by looking at [the last commit](https://github.com/monarch-initiative/mondo-ingest/pull/510/commits/ca64342c390e538b8d12cf425672cf8c5ff22977).

src/ontology/mondo-ingest.Makefile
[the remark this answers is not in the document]
joeflack4: Build complete!

src/sparql/select-ordo-subsets.sparql
joeflack4, line 1: ## Need to merge #531 first
> [!WARNING]  
> #531 should be approved and merged into this PR before this PR is merged into `develop`

joeflack4: - [ ] @joeflack4 consider running build again after this is merged

matentzn: I will leave this to you and Trish

matentzn (approved):
LGTM!

src/scripts/ordo_subsets.py
joeflack4: Note: Nico approves.

src/ontology/mondo-ingest.Makefile
joeflack4: Note: Nico has reviewed and approves.

src/sparql/select-ordo-subsets.sparql
twhetzel: I reviewed and approved #531

PR #521, opened 2024-05-16
Create `mondo-nando.sssom.tsv`

matentzn:
also addresses https://github.com/monarch-initiative/mondolib/issues/11

---

This PR adds the manually curated SSSOM mappings between Nando and Mondo to the Mondo ingest.

This is needed for the Rare Disease Paper. To take the pressure of this PR, I will use a manual translation to proceed, but this still needs to be done at some point before the paper is submitted (which is soon, not sure how soon)

## Overview

This PR:
- Adds a sssom file that is manually updated to the mondo-ingest

### Documentation
- [ ] Yes

### QC
Build PR: 
- #552
- [x] Yes
   
### New Packages
- [x] No

- [x] No

- [x] Yes


src/ontology/mondo-ingest.Makefile
```diff
@@ -563,12 +563,18 @@ $(EXTERNAL_CONTENT_DIR)/%.robot.owl: $(EXTERNAL_CONTENT_DIR)/%.robot.tsv
 $(EXTERNAL_CONTENT_DIR)/nord.robot.tsv: $(TMPDIR)/nord.tsv config/external-content-robot-headers.json
 	mkdir -p $(EXTERNAL_CONTENT_DIR)
 	python ../scripts/add-robot-template-header.py $^ > $@
-.PRECIOUS: $(EXTERNAL_CONTENT_DIR)/nord.robot.owl
+.PRECIOUS: $(EXTERNAL_CONTENT_DIR)/nord.robot.tsv
 
-.PHONY: external-content-nord
+$(EXTERNAL_CONTENT_DIR)/nando-mappings.robot.tsv: $(MAPPINGSDIR)/nando-mondo.sssom.tsv
+	mkdir -p $(EXTERNAL_CONTENT_DIR)
+	# python $< $@ .... TODO @joeflack4 to fill in
```
matentzn, line 570: @joeflack4 this row is for you using your sssom -> mondo robot template converter

joeflack4: @matentzn Roger that, I'll whip something up.

joeflack4: @matentzn Ready for your review. I'll leave some more comments, but I'll drop the [robot template google sheet](https://docs.google.com/spreadsheets/d/1A4uJAOkFPnZ9VQj_tk8Hh1Auwdi9MuXtLflHGzW4LfM/edit#gid=581634025) here.

src/ontology/mondo-ingest.Makefile
```diff
@@ -563,12 +563,18 @@ $(EXTERNAL_CONTENT_DIR)/%.robot.owl: $(EXTERNAL_CONTENT_DIR)/%.robot.tsv
 $(EXTERNAL_CONTENT_DIR)/nord.robot.tsv: $(TMPDIR)/nord.tsv config/external-content-robot-headers.json
 	mkdir -p $(EXTERNAL_CONTENT_DIR)
 	python ../scripts/add-robot-template-header.py $^ > $@
-.PRECIOUS: $(EXTERNAL_CONTENT_DIR)/nord.robot.owl
+.PRECIOUS: $(EXTERNAL_CONTENT_DIR)/nord.robot.tsv
 
-.PHONY: external-content-nord
+$(EXTERNAL_CONTENT_DIR)/nando-mappings.robot.tsv: $(MAPPINGSDIR)/mondo-nando.sssom.tsv
+	mkdir -p $(EXTERNAL_CONTENT_DIR)
+	python ../scripts/sssom_to_robot_template.py --inpath $^ --outpath $@
```
joeflack4, line 570: Updated the make goal for creating robot template

src/scripts/sssom_to_robot_template.py
```diff
@@ -0,0 +1,54 @@
+"""Convert SSSOM to ROBOT template."""
```
joeflack4, line 1: Script for converting SSSOM to ROBOT template

addresses:
- https://github.com/monarch-initiative/mondolib/issues/11

src/scripts/sssom_to_robot_template.py
```diff
@@ -0,0 +1,54 @@
...
+    'object_id': 'A oboInOwl:hasDbXref',
+    'object_label': '',
+    'equivalence': '>A oboInOwl:source',
+    'ORCID': '>A oboInOwl:source SPLIT=|',
+}
+
+
+def sssom_to_robot_template(inpath: Union[str, Path], outpath: Union[str, Path]):
+    """Convert SSSOM to ROBOT template."""
+    msdf: MappingSetDataFrame = parse_sssom_table(inpath)
```
joeflack4, line 22: ## General thoughts: `parse_sssom_table()`
@matentzn We had a recent discussion that it would be nice to have `msdf.to_csv()`. It'd also be nice to have `read_csv()` or `read_tsv()`. I feel like this is a better name for what this function does, and also consistent w/ pandas, improving usability.

src/ontology/mondo-ingest.Makefile
```diff
@@ -563,12 +563,18 @@ $(EXTERNAL_CONTENT_DIR)/%.robot.owl: $(EXTERNAL_CONTENT_DIR)/%.robot.tsv
 $(EXTERNAL_CONTENT_DIR)/nord.robot.tsv: $(TMPDIR)/nord.tsv config/external-content-robot-headers.json
 	mkdir -p $(EXTERNAL_CONTENT_DIR)
 	python ../scripts/add-robot-template-header.py $^ > $@
-.PRECIOUS: $(EXTERNAL_CONTENT_DIR)/nord.robot.owl
+.PRECIOUS: $(EXTERNAL_CONTENT_DIR)/nord.robot.tsv
 
-.PHONY: external-content-nord
+$(EXTERNAL_CONTENT_DIR)/nando-mappings.robot.tsv: $(MAPPINGSDIR)/mondo-nando.sssom.tsv
```
joeflack4, line 568: @matentzn FYI: had to flip filename. You had `nando-mondo.sssom.tsv` here, but the file that is committed is `mondo-nando.sssom.tsv`

src/scripts/sssom_to_robot_template.py
```diff
@@ -0,0 +1,52 @@
...
+    df: pd.DataFrame = msdf.df
+
+    # Filter
+    df = df[df['predicate_id'] == 'skos:exactMatch']
```
joeflack4, line 26: FYI: filtering only exact matches

src/ontology/mondo-ingest.Makefile
matentzn: thx

src/ontology/mondo-ingest.Makefile
matentzn:
```suggestion
	python ../scripts/sssom_to_robot_template.py --inpath $< --outpath $@
```

for future proofing

src/scripts/sssom_to_robot_template.py
```diff
@@ -0,0 +1,52 @@
...
+from sssom import MappingSetDataFrame
+from sssom.parsers import parse_sssom_table
+
+ROBOT_ROW = {
+    'subject_id': 'ID',
+    'subject_label': '',
+    'object_id': 'A oboInOwl:hasDbXref',
+    'object_label': '',
+    'equivalence': '>A oboInOwl:source',
+    'ORCID': '>A oboInOwl:source SPLIT=|',
```
matentzn, line 16: to be more consistent with sssom:

```suggestion
    'author_id': '>A oboInOwl:source SPLIT=|',
```

matentzn: (not needed, if you think its not necessary to be consisent just close the comment)

src/scripts/sssom_to_robot_template.py
```diff
@@ -0,0 +1,52 @@
...
+    msdf: MappingSetDataFrame = parse_sssom_table(inpath)
+    df: pd.DataFrame = msdf.df
+
+    # Filter
+    df = df[df['predicate_id'] == 'skos:exactMatch']
+
+    # Conversion
+    df = df[['subject_id', 'subject_label', 'object_id', 'object_label']].sort_values(['subject_id', 'object_id'])
+    df['equivalence'] = 'MONDO:equivalentTo'
+    df['ORCID'] = '|'.join(msdf.metadata['creator_id'])
```
matentzn, line 31: fix this if you apply my suggestion above:

```suggestion
    df['author_id'] = '|'.join(msdf.metadata['creator_id'])
```

src/scripts/sssom_to_robot_template.py
```diff
@@ -0,0 +1,52 @@
...
+    df = pd.concat([pd.DataFrame([ROBOT_ROW]), df])
+
+    # Write
+    df.to_csv(outpath, sep='\t', index=False)
+
+
+def cli():
```
matentzn, line 38: This code should really be in mondolib!

matentzn:
Thanks @joeflack4 some minor suggestions but looks good!

src/scripts/sssom_to_robot_template.py
```diff
@@ -0,0 +1,52 @@
...
+
+import pandas as pd
+from sssom import MappingSetDataFrame
+from sssom.parsers import parse_sssom_table
+
+ROBOT_ROW = {
+    'subject_id': 'ID',
+    'subject_label': '',
+    'object_id': 'A oboInOwl:hasDbXref',
+    'object_label': '',
```
matentzn, line 14: Seems like columns starting with `>A` (axiom annotations) have to be directly following the main column for this to work `object_id`. Need to drop, or better move, `object_label` column somewhere later.

src/scripts/sssom_to_robot_template.py
matentzn: I just realised that the mappings are very often not exact.. grrr. instead of filtering, can do ith like this:

if skos:exactMatch --> Mondo:equivalentTo in `equivalence` column
else if skos:relatedMatch --> Mondo:whateverTheRelatedToCode
else if skos:broadMatch --> Mondo:whateverTheMondoIsNarrowerThanObjectCodeIs
else --> do not include anything in the `equivalence` column, but keep the record.

src/scripts/sssom_to_robot_template.py
```diff
@@ -0,0 +1,52 @@
...
+    df: pd.DataFrame = msdf.df
+
+    # Filter
+    df = df[df['predicate_id'] == 'skos:exactMatch']
+
+    # Conversion
+    df = df[['subject_id', 'subject_label', 'object_id', 'object_label']].sort_values(['subject_id', 'object_id'])
+    df['equivalence'] = 'MONDO:equivalentTo'
+    df['ORCID'] = '|'.join(msdf.metadata['creator_id'])
+    df = pd.concat([pd.DataFrame([ROBOT_ROW]), df])
```
matentzn, line 32: add another column:

df['mapping_provider'] = msdf.metadata['mapping_provider'] and have it be added the as `A oboInOwl:hasDbXref` just as the others.

src/ontology/mondo-ingest.Makefile
```diff
@@ -189,8 +189,9 @@ $(TMPDIR)/component-%.json: $(COMPONENTSDIR)/%.owl
 	$(ROBOT) convert -i $< -f json -o $@
 .PRECIOUS: $(TMPDIR)/component-%.json
 
-$(MAPPINGSDIR)/%.sssom.tsv: $(TMPDIR)/component-%.json metadata/mondo.sssom.config.yml
-	sssom parse $< -I obographs-json --prefix-map-mode merged -m metadata/mondo.sssom.config.yml -o $@
+$(MAPPINGSDIR)/%.sssom.tsv:
+	make $(TMPDIR)/component-$*.json metadata/mondo.sssom.config.yml
```
joeflack4, line 193:
```suggestion
	$(MAKE) $(TMPDIR)/component-$*.json metadata/mondo.sssom.config.yml
```
I don't know what the advantage of `$(MAKE)` is. Actually, I don't even see where it's defined. I do see that `MAKE_FAST` is defined in the `Makefile`, though.

joeflack4, line 193: @matentzn Why not dependencies?

The only reason I can remember why we move these goals from dependencies into the recipe block is cases of `.PHONY` goals, to avoid the goal from running 100% of the time. But these aren't phony. Why the move?

src/scripts/sssom_to_robot_template.py
joeflack4: @matentzn Sounds good!

I was going to suggest that this code might fit better into `sssom-py`, but what with the `MONDO` related annotations, I guess not.

joeflack4: @matentzn Are you sure about this one?
> else if `skos:broadMatch` --> `MONDO:mondoIsNarrowerThanSource`

---

I have my own notes about `skos:broadMatch` and `skos:narrowMatch`, where I've written:
> In a `skos:broadMatch` b, 'a' is broader than 'b'.

We're parsing `mondo-nando.sssom.tsv`, and the Mondo IDs are on the left (subject), and NanDO on the right (object). So for `skos:broadMatch` I should want `MONDO:mondoIsBroaderThanSource`, no?

I apologize, I know this is fundamental and I should know this.

As my source for this factoid, I have linked ChatGPT: https://chat.openai.com/share/3b6942a1-8152-4345-b1cd-763dd4403e2a

...which I realize could be totally wrong. The only reason I asked it is because the [skos docs](https://www.w3.org/TR/skos-reference/) are very hard to 

To make things more confusing, I've heard that `broadMatch` flips the order of `broader`, and `narrowMatch` flips the order of `narrower`, but I can't find any source that confirms this. What I do see is that these are sub-properties... and if that's the case, I'd imagine they probably would not be flipped.

It's easier to find documentation on `broader` and `narrower`, but the examples I find are conflicting.

(CC @twhetzel in case you happen to have a good solid understanding of this. Otherwise if Nico can give me a firm statement on this or link to a good source, you'd probably want to know too)

src/scripts/sssom_to_robot_template.py
joeflack4: it's good; i'll commit

src/scripts/sssom_to_robot_template.py
joeflack4: @matentzn That was the original goal, and I created https://github.com/monarch-initiative/mondolib/issues/11 in `mondolib` for that reason, but:
1. The current state of `mondolib` doesn't lean itself well to additions; needs a refactor.
2. For now, @twhetzel has OK'ed additional Python in `mondo-ingest`.

But I think we both would agree that eventually it should make its way there.

src/scripts/sssom_to_robot_template.py
joeflack4: Done!

src/scripts/sssom_to_robot_template.py
joeflack4: When you say add it to the `'object_id': 'A oboInOwl:hasDbXref'`, you mean add it as `'mapping_provider': '>A oboInOwl:source'`? This is what I've done. Can you confirm that's correct?

src/scripts/sssom_to_robot_template.py
joeflack4: Also you don't want `skos:narrowMatch` --> `MONDO:mondoIsBroaderThanSource`? For now, if that's the case, the cell will just be blank.

joeflack4: Alright, this is done. FYI, `exactMatch` is the only one in our current set but I imagine that may change in the future.

```py
PRED_MAP = {
    'skos:relatedMatch': 'MONDO:relatedTo',
    'skos:exactMatch': 'MONDO:equivalentTo',
    'skos:broadMatch': 'MONDO:mondoIsNarrowerThanSource',
}
```

```
df['equivalence'] = df['predicate_id'].map(lambda pred: PRED_MAP.get(pred, ''))
```

src/scripts/sssom_to_robot_template.py
twhetzel: I Ok'd more Python in `mondo-ingest` for the ORDO(?) work needed before the release, but ideally anything new should go in `mondolib`... let me see if I can refactor enough to get a stub for adding new Python code into `mondolib` for the near future.

_IF_ this work were moved to `mondolib` how soon does it need to be in there?

joeflack4: Ah, OK I think I heard this without the explicit ORDO qualifier. I'm totally fine w/ you/me/us refactoring this into / starting up work on `mondolib` whenever.

src/scripts/sssom_to_robot_template.py
matentzn: I hear you but this comment is not actionable in this PR; this would have to be implemented in sssom-py

src/scripts/sssom_to_robot_template.py
matentzn:
```
In a skos:broadMatch b, 'a' is broader than 'b'.
```

Check https://mapping-commons.github.io/sssom/mapping-predicates/#precision

"The object is conceptually broader than the subject.".

Your solution sounds correct to me!

src/ontology/mondo-ingest.Makefile
matentzn: I realised the dependencies would fire in cases where the mapping is manually curated (i.e. just uploaded, not extracted from a JSON file). This can be seen with the `nando` example in this PR. So `make` tries to generate the json dependency when building `$(MAPPINGSDIR)/mondo-nando.sssom.tsv` but cant, because its not there (and it is not needed).

matentzn: $(MAKE) uses the same make shell `make` creates a new one, a `make` shell within a `make` shell so to speak. This is a built-in variable

src/scripts/sssom_to_robot_template.py
matentzn: yes!

src/ontology/mondo-ingest.Makefile
matentzn: Either one is fine here, you can push your commit

src/scripts/sssom_to_robot_template.py
```diff
@@ -0,0 +1,57 @@
...
+import pandas as pd
+from sssom import MappingSetDataFrame
+from sssom.parsers import parse_sssom_table
+
+ROBOT_ROW = {
+    'subject_id': 'ID',
+    'subject_label': '',
+    'object_id': 'A oboInOwl:hasDbXref',
+    'equivalence': '>A oboInOwl:source',
+    'author_id': '>A oboInOwl:source SPLIT=|',
```
matentzn, line 15:
```suggestion
    'author_id': '>AI oboInOwl:source SPLIT=|',
```

Orcids are IRIs:

xref: ICD10CM:N00-N99 {source="MONDO:obsoleteEquivalent", source="https://orcid.org/0000-0001-5208-3432", source="https://orcid.org/0000-0002-4142-7153"}

src/scripts/sssom_to_robot_template.py
joeflack4: NP; I guess your preference is not to bring it up in the PR and just make an issue. I just like to bring it up in advance of making the issue, but I guess going straight to the issue is fine; I'll try to remember this preference, as it is a habit of mine to bring up tangential thoughts in PRs.

src/ontology/mondo-ingest.Makefile
joeflack4: Ahh, I see, thanks for explaining!
I'll do the commit here just for consistency, then.

src/scripts/sssom_to_robot_template.py
joeflack4: Understood! Thanks. Committing the suggestion

src/scripts/sssom_to_robot_template.py
joeflack4: I changed this issue from being about allowing `MappingSetDataFrame` to accept paths, to instead being about these `pandas`-like functions/methods:
- https://github.com/mapping-commons/sssom-py/issues/517

src/scripts/sssom_to_robot_template.py
```diff
@@ -0,0 +1,57 @@
...
+        .sort_values(['subject_id', 'object_id'])
+    df['author_id'] = '|'.join(msdf.metadata['creator_id'])
+    df['mapping_provider'] = msdf.metadata['mapping_provider']
+    df = pd.concat([pd.DataFrame([ROBOT_ROW]), df])
+
+    # Write
+    df.to_csv(outpath, sep='\t', index=False)
+
+
+def cli():
```
joeflack4, line 43: - [ ] @matentzn undraft and build or keep drafted

@matentzn The only open thread I see is the [possible `mondolib` refactor](https://github.com/monarch-initiative/mondo-ingest/pull/521#discussion_r1604500855). Other than that, I don't have any todo's left in this PR. If you don't either, and if we opt to do the `mondolib` thing later, then we can undraft this PR and run the build, I think.

src/scripts/sssom_to_robot_template.py
matentzn: me too tbh

src/scripts/sssom_to_robot_template.py
matentzn: Dont worry I will take it from here. There are still some issues to iron out which do not relate to you (related to the source mapping fidelity)

src/scripts/sssom_to_robot_template.py
joeflack4: @matentzn FYI Trish requested that I this work move this over to `mondolib`. I probably won't complete that until 1-3 weeks from now, though, so we don't need to hold up this PR until that's done; can remove this code in a subsequent PR.

matentzn: There is no great urgency here, as I can (a) work of the branch for now and (b) there are multiple issues to iron out with the raw data!

joeflack4: Got it! Spoke with Trish and we prioritized https://github.com/monarch-initiative/mondolib/issues/12 for later.

src/ontology/mondo-ingest.Makefile
```diff
@@ -553,7 +557,9 @@ $(TMPDIR)/nord.tsv:
 	wget "https://rdbdev.wpengine.com/wp-content/uploads/mondo-export/rare_export.tsv" -O $@
 
 $(EXTERNAL_CONTENT_DIR)/%.robot.owl: $(EXTERNAL_CONTENT_DIR)/%.robot.tsv
-	$(ROBOT) template --template $< \
+	$(ROBOT) template \
+		--template $< \
+		--prefix "orcid: https://orcid.org/" \
```
matentzn, line 562: Necessary so that the `orcid` is correctly transformed into a URI in OWL format.

src/ontology/mondo-ingest.Makefile
```diff
@@ -563,12 +569,22 @@ $(EXTERNAL_CONTENT_DIR)/%.robot.owl: $(EXTERNAL_CONTENT_DIR)/%.robot.tsv
 $(EXTERNAL_CONTENT_DIR)/nord.robot.tsv: $(TMPDIR)/nord.tsv config/external-content-robot-headers.json
 	mkdir -p $(EXTERNAL_CONTENT_DIR)
 	python ../scripts/add-robot-template-header.py $^ > $@
-.PRECIOUS: $(EXTERNAL_CONTENT_DIR)/nord.robot.owl
+.PRECIOUS: $(EXTERNAL_CONTENT_DIR)/nord.robot.tsv
+
+$(MAPPINGSDIR)/mondo-nando.sssom.tsv: $(MAPPINGSDIR)/nando-mondo.sssom.tsv
+	sssom invert $(MAPPINGSDIR)/nando-mondo.sssom.tsv --no-merge-inverted -o $@
+	sssom annotate $@ --mapping_provider "MONDO:NANDO" -o $@
```
matentzn, line 576: This is necessary so we get the correct "source" annotation for mondo edit

matentzn (commented):
Approved! Cant approve officially as this is my PR.

src/ontology/mondo-ingest.Makefile
joeflack4: @matentzn I assume this looks good? Can you check these boxes above and marked resolved if so?

src/ontology/external/nando-mappings.robot.tsv
joeflack4, line 1: ## Build / QC step
- [ ] @matentzn or @joeflack4 Run the build

I would approve, but this is the only thing that seems to be missing now. If you want, I can run it for you.

src/scripts/sssom_to_robot_template.py
twhetzel: And just to clarify, https://github.com/monarch-initiative/mondolib/issues/12 was prioritized for later since Joe was pulled into work on subsets (e.g. all the ORDO PRs and this PR).

src/ontology/external/nando-mappings.robot.tsv
joeflack4: I ran it. Won't be able to review until Tuesday I think, but here's the PR:
- #552

src/ontology/mondo-ingest.Makefile
twhetzel: @joeflack4 does `nando-mondo.sssom.tsv` need to be removed from the commit?

joeflack4: @twhetzel Good question. Originally, I believe there was just 1 file. I looked now and there is indeed a `mondo-nando` and a `nando-mondo` variation. This doesn't seem to be an error. The subject/objects correctly match with the naming order. And there is a separate goal for each of these in the `makefile`. So I guess I'm not worried about it, but it would be helpful if @matentzn could clarify why we're doing this.

matentzn: The nando-mondo mappings were supplied by the Nando, the pipeline invertes it in row 595 to be mondo-nando which is what we need in Mondo. Then you generate the ROBOT template from it in externally provided content! All looks great.

matentzn (commented):
Approved

src/ontology/mondo-ingest.Makefile
matentzn: The output looks good!

PR #530, opened 2024-05-21
Build 2024-05-20

twhetzel:
## Overview

New build.

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [x] No

- [x] No

- [ ] Yes


src/ontology/slurp/omim.tsv
joeflack4, line 1: ## How to review builds
@matentzn Do you have any advice? I feel like I should be reviewing my own builds or those of others, but other than checking some of the final outputs (if they even appear in the diff), such as `slurp` or `lexmatch`, and doing a super basic face check, I'm not sure what else to look for.

joeflack4, line 1: ## Build QC?
@twhetzel It seems like some of the face checking we do for builds is pretty basic. I don't think we have a GH issue for creating a QC for builds. I think such a check would run on every PR too, which is probably fine.

I suppose I ought to create an issue for this? It would be nice to have. What would be the urgency/priority of such?

joeflack4 (approved):
LGTM at a very basic face-check; just scrolling through the files, as well as checking a couple outputs, while keeping in mind the small changes that the build is based on (#513). As I mention [here](https://github.com/monarch-initiative/mondo-ingest/pull/530/files#r1608880771), I'm not really good yet at reviewing builds.

src/ontology/slurp/omim.tsv
joeflack4, line 1: ## Process thought / suggestion
I'm not necessarily suggesting this, but just a thought. The process we followed w/ this change --> build process was:
1. Open a PR w/ changes (#513)
2. Merge into `main`
3. Run the build off of `main`

We could also do:
1. Open a PR w/ changes (e.g. #513)
2. Run a build off of the feature branch
3. Merge build branch into feature branch
4. Merge feature branch into `main`

I guess this would allow us to notice and correct if anything went wrong, before merging into `main`. However the only reason I don't think this is too important is because the files that #513 changes are not code changes.

sabrinatoro (commented):
I do not see a new/updated slurp file for DO. 
The DO file included x-refs that were already in Mondo main. Therefore I would expect to see a new/updated file for DO

twhetzel:
I ran `sh run.sh make build-mondo-ingest`, was another make command supposed to be run for this?

matentzn:
It looks like you ran `sh run.sh make build-mondo-ingest` but you should run: `sh run.sh make build-mondo-ingest -B` if you run the whole pipeline. That should probably be mentioned in the docs as well.

joeflack4:
Or, can do what I do and clone a copy of the repo and run the build there. I think this should have the same effect as `-B`, but probably the reason I started doing this is so I could continue development while the build runs in parallel.

twhetzel:
Thanks, I'll re-run with `-B` and created a ticket to update the docs.

Issue #529, opened 2024-05-20
Goal &/or SOP for updating `mondo.sssom.tsv`


joeflack4:
## Overview
Unfortunately, the goal / nature of updating `mondo.sssom.tsv` is different than other artefacts. It does not suffice to simply delete the file and run the goal.

## Sub-tasks
- [ ] a. Goal
- [ ] b. SOP
  - [ ] b.i. Decide on the procedure
  - [ ] b.ii. Write the SOP

#### a. `make tmp/mondo.sssom.tsv -B`
This is [what I proposed](https://github.com/monarch-initiative/mondo-ingest/pull/510#discussion_r1597500072). I don't know if there is some reason why it wouldn't work or otherwise if it has any downsides.

#### b. `rm tmp/mondo_repo_built tmp/mondo.sssom.tsv; make tmp/mondo.sssom.tsv`
Nico [proposed this](https://github.com/monarch-initiative/mondo-ingest/pull/510#discussion_r1597374539).

## Additional info
Context: [discussion where the idea for this SOP came up](https://github.com/monarch-initiative/mondo-ingest/pull/510#discussion_r1597374539)

If need inspiration for writing SOP, look at this as an example: https://mondo.readthedocs.io/en/stable/editors-guide/mondo-docs-website-editing/

Related:
- #538

joeflack4:
@matentzn It seems to me that 'a' should work and is simpler. Do you think it is the best option or is 'b' better?

matentzn:
I think for this very isolated request "refreshing mondo.sssom.tsv" they are identical, and your solution is ok. The problem comes when you want to run larger pipelines that depend on mondo.sssom.tsv - you cant just use `-B` in those unless you want to re-run absolutely everything! I am just telling you how I would do it, you do whatever you prefer. Both have the same effect in the end - mondo.sssom.tsv will be refreshed. but `-B` invalidates _all_ the `make` cache, which means that everything else is rerun as well!

twhetzel:
Given the past issues we've run into with `-B` and it's effects on running pipelines, I would prefer option (b) since it sounds like it will work for both the issue of "refreshing mondo.sssom.tsv" and be able to be used within the context of larger pipelines.

joeflack4:
Thanks for your thoughts, guys.

Oh, I just thought of this, if you guys don't mind me adding, can I do this instead?:

Create a goal:
```make
.PHONY: refresh-mondo-sssom
refresh-mondo-sssom:
[...]

PR #531, opened 2024-05-21
Add the orphanet subset to the Slurp file

### LINKED ISSUE #527: Add the orphanet subset to the Slurp file
when reviewing the slurp/ordo.tsv, it would be helpful to know whether the orphanet term is in the "grouping disease", "disorder", or "subtype of disease" subset in orphanet. 
Please add a column in the slurp/ordo.tsv for the orphanet subset. Thank you!

joeflack4:
Resolves #527

## Overview
Adds rare disease subset information to `slurp/ordo.tsv` in a new column.

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [x] No

- [x] No

- [x] Yes

---

@souzadevinicius FYI


src/ontology/mondo-ingest.Makefile
```diff
@@ -598,6 +610,19 @@ update-jinja-sparql-queries:
...
+
+#################
+##### Ad hoc ####
+#################
+tmp/ordo-subsets.tsv: $(COMPONENTSDIR)/ordo.owl
+	$(ROBOT) query -i $< --query ../sparql/select-ordo-subsets.sparql $@
+
+reports/ordo-subsets.robot.template.tsv: tmp/ordo-subsets.tsv tmp/mondo.sssom.tsv
+	python3 $(SCRIPTSDIR)/ordo_subsets.py \
+	--mondo-mappings-path tmp/mondo.sssom.tsv \
```
twhetzel, line 622: Is this getting the latest `mondo.sssom.tsv` file?

joeflack4: ## This shouldn't have been part of the diff
@twhetzel Sorry, this is a wonky PR situation. That change is not part of this PR. It's part of #510, the branch for which this PR was used as a base.

I originally set this PR to be merged into `develop`, because that is where we do want to merge it. However, I forgot that this would make it so that all of the other changes in #510 would also show up here. I've now changed it so that this will merge into #510, so now the diff that shows is correct. The workflow we'll want to do, I think, is: (1) merge #510 into `develop` (we're just waiting on Nico's final explicit approval for that), and then I can (2) change the base for this to merge into `develop`.

twhetzel: no worries. i just wanted to make sure that the new SOP/goal in https://github.com/monarch-initiative/mondo-ingest/issues/529 is accounted for here in `reports/ordo-subsets.robot.template.tsv` so that it also has the latest `mondo.sssom.tsv` file

joeflack4: ## Answer to the question
In any case, as per your original question, there are several make goals which depend on `tmp/mondo.sssom.tsv`. You bring up a good point, related to #529. The way this dependency is set up, simply by listing `tmp/mondo.sssom.tsv` in the prerequisites of the goal, is the typical way to do it in `make`. However, in the event that the `mondo` repo has changed since `tmp/mondo.sssom.tsv` was generated, `make` will have no way of knowing that, and will think that `tmp/mondo.sssom.tsv` is up-to-date, and use the one that's presently there on the system.

Perhaps we're fine with that being the behavior. During a fresh build, this will not be a problem. It'll only be a problem in development situations. If we want to fix this for those situations as well, then, as far as I see it, we have 2 options.

Both options involve at least this change:
Move dependency into the goal's body. In this example, that would change it to:
```make
reports/ordo-subsets.robot.template.tsv: tmp/ordo-subsets.tsv 
	$(MAKE) tmp/mondo.sssom.tsv
	python3 $(SCRIPTSDIR)/ordo_subsets.py \
	...
```

### a. Always run a full refresh of `tmp/mondo.sssom.tsv`
Always run a refresh each time a goal is run which has it as a dependency.

This involves nothing else in addition to the change I just described above.

This could be be a real damper on local development. Not always terribly so, though. When I'm developing a goal that uses Python, I usually do all of my iteration in the debugger, and then run the make goal once at the end just to make sure it works.

### b. Intelligently refresh `tmp/mondo.sssom.tsv` when needed
This would be cool to do at some point. Basically, in addition to the change I described above, we'd have an if statement at the top of the goal for `tmp/mondo.sssom.tsv` or `tmp/mondo_repo_built`. It would compare the latest git commit hashes on `master` for  the cloned `mondo` dir in `tmp/` and the actual repo on GitHub. If they're the same, do nothing. Else, it will re-run `tmp/mondo_repo_built` --> `tmp/mondo.sssom.tsv`.

---

I think this merits its own GH issue.

src/ontology/slurp/ordo.tsv
```diff
@@ -1,13 +1,13 @@
...
-MONDO:0968989	non-saccular limited dorsal myeloschisis	Orphanet:645343	MONDO:equivalentTo	Non-saccular limited dorsal myeloschisis	A rare form of limited dorsal myeloschisis (LDM), characterized by a non saccular cutaneous stigmata (midline skin abnormality classically dimple, pit or sometimes angioma), the stalk is attached to this cutaneous stigmata. Fibroneural stalk varies in thickness and complexity.	MONDO:0958345|MONDO:8000030|MONDO:8000034
-MONDO:0968990	genetic central precocious puberty in male	Orphanet:650097	MONDO:equivalentTo	Genetic central precocious puberty in male		MONDO:8000031|MONDO:0958356
-MONDO:0968991	non-genetic central precocious puberty in male	Orphanet:650102	MONDO:equivalentTo	Non-genetic central precocious puberty in male		MONDO:0958356|MONDO:8000031
+mondo_id	mondo_label	xref	xref_source	original_label	definition	parents	subset
+ID	LABEL	A oboInOwl:hasDbXref	>A oboInOwl:source SPLIT=|		A IAO:0000115	SC %	
+MONDO:0958337	disorder with optic nerve compression	Orphanet:519337	MONDO:equivalentTo	Disorder with optic nerve compression		MONDO:8000033	group of disorders
```
joeflack4, line 3: ## Output
@sabrinatoro Here you are! Check out this [google sheet](https://docs.google.com/spreadsheets/d/13styiaK6DhF03VF3IEGg-78sJH1pICOfPWzG5iowUpY/edit#gid=1585402401).

src/scripts/ordo_subsets.py
```diff
@@ -18,19 +18,9 @@
...
-    # Read: mondo.sssom.tsv
-    mondo_df = pd.read_csv(mondo_mappings_path, sep='\t', comment='#').rename(columns={
-        'subject_id': 'mondo_id',
-        'subject_label': 'mondo_label',
-    })
-    # - keep only exact matches
-    mondo_df = mondo_df[mondo_df['predicate_id'] == 'skos:exactMatch']
-
-    # Read: ORDO class subsets query results
+def get_formatted_subsets_df(class_subsets_tsv_path: str):
```
joeflack4, line 21: ## `get_formatted_subsets_df()`
In `ordo_subsets.py`, which is part of #510, there existed some code I needed to re-use; namely, the reading and formatting of `tmp/ordo-subsets.tsv`. So I created this function for that, which is now used in 2 places. That's what the whole diff in `ordo_subsets.py` is about, just moving some code into this new function.

src/ontology/slurp/ordo.tsv
joeflack4: @sabrinatoro You don't need this merged into `main` any time soon, do you? Or, phrased another way--do you foresee needing this information to show up in another build very soon?

If not, then I suppose that the google sheet will be fine for your immediate purposes. I only bring this up because I was planning to merge this into `develop`, which is our typical workflow,  but I don't know when the next time is that `develop` would get merged into `main`. If you foresee that you'll need to see this information again in a build very soon, then I think we can do a merge of `develop` into `main` after this PR is merged.

joeflack4: @sabrinatoro Does the `subset` column look good to you?
It will show one of:
- disorder
- group of disorders
- subtype of a disorder

If you want, alternatively I could have it show one of these things instead:
```python
    'Orphanet:557493': 'http://purl.obolibrary.org/obo/mondo#ordo_disease',  # disorder
    'Orphanet:557492': 'http://purl.obolibrary.org/obo/mondo#ordo_group_of_disorders',  # group of disorders
    'Orphanet:557494': 'http://purl.obolibrary.org/obo/mondo#ordo_clinical_subtype',  # subtype of a disorder
```

src/ontology/mondo-ingest.Makefile
twhetzel: I agree this could use it's own GH issue and maybe a ~todo/TODO for the line in this PR to evaluate the outcome of the future ticket

src/ontology/slurp/ordo.tsv
twhetzel: I also took a look at the spreadsheet and things look good

Joe, if in the future you're interested to know more about Orphanet for things like this, one place to check is the Orphanet web site (https://www.orpha.net/en/disease) and search for a disease, e.g. saccular spinal dysraphism with a stalk to the dome. When you click on the resulting [link](https://www.orpha.net/en/disease/detail/645319?name=saccular%20spinal%20dysraphism%20with%20a%20stalk%20to%20the%20dome&mode=name) for the disease name and scroll down the Classification level is listed on the page.

src/ontology/mondo-ingest.Makefile
joeflack4: @twhetzel Sounds good! I created:
- #538 

However, I didn't add a `todo` because this is more pervasive; there are 4 goals that use `tmp/mondo.sssom.tsv` as a dependency. If you want I can insert a `todo` on each of them.

twhetzel: RE: `todo` - whichever works better for your tracking

sabrinatoro:
Thank you @joeflack4 ! That looks great! The additional column in the spreadsheet is exactly what we needed, and it will make review easier. Thank you! 

I do not need this merged into main soon. So there is no urgency from my perspective.

src/ontology/slurp/ordo.tsv
joeflack4: [Sabrina approves](https://github.com/monarch-initiative/mondo-ingest/pull/531#issuecomment-2128148710)! Marking this one resolved.

twhetzel (approved):
Looks good to me, just one comment for clarification

src/scripts/migrate.py
```diff
@@ -173,6 +174,25 @@ def slurp_docs():
...
+    """Adds rare_disease_subset column to the ORDO migration TSV."""
+    # Read inputs
+    df: pd.DataFrame = pd.read_csv(slurp_path, sep='\t')
+    df_subsets: pd.DataFrame = get_formatted_subsets_df(subsets_path)
+    df_subsets = df_subsets[['ordo_id', 'subset_ordo_class_label']]\
+        .rename(columns={'ordo_id': 'xref', 'subset_ordo_class_label': 'subset'})
+
+    # Edge case: re-running
+    if 'subset' in df.columns:
+        del df['subset']
```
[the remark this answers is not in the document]
joeflack4: Good question! Basically we're editing `slurp/ordo.tsv` to add a `subset` column. The edge case is, if `slurp/ordo.tsv` exists already and it already has that column, we want to replace whatever it has there. There are a couple ways to do this, but I simply deleted the column from the slurp file, then joined with the separate file which lists all ORDO classes and their subsets, which reside in a column called 'subset'. So if I didn't delete it first, I'd have 2 copies of subset columns after the join, subset_x and subset_y I think, and then I'd have to rename one of them and drop the other. So the more elegant way seemed to be simply deleting the column that was going to be overwritten anyway.

PR #532, opened 2024-05-21
Delete goals: `slurp-no-updates`

joeflack4:
## Overview
Deletes some unused goals which add no additional functionality.

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [x] No

- [x] No

- [ ] Yes

### DIFF
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -527,18 +527,9 @@ slurp-%: slurp/%.tsv
 slurp-ordo: slurp/ordo.tsv
 	$(MAKE) slurp-modifications-ordo
 
-.PHONY: slurp-no-updates-%
-slurp-no-updates-%: slurp/%.tsv
-	@echo "$@ completed".
-
 .PHONY: slurp-docs
 slurp-docs:
 	python3 $(SCRIPTSDIR)/migrate.py --docs
-
-.PHONY: slurp-all-no-updates
-slurp-all-no-updates: $(foreach n,$(ALL_COMPONENT_IDS), slurp-no-updates-$(n))
-	$(MAKE) slurp-modifications
-	@echo "$@ ($^) completed".
 
 .PHONY: slurp-all
 slurp-all: $(foreach n,$(ALL_COMPONENT_IDS), slurp-$(n))


src/ontology/mondo-ingest.Makefile
```diff
@@ -486,18 +486,10 @@ slurp/%.tsv: $(COMPONENTSDIR)/%.owl $(TMPDIR)/mondo.sssom.tsv $(REPORTDIR)/%_map
...
-slurp-no-updates-%: slurp/%.tsv
-	@echo "$@ completed".
-
 .PHONY: slurp-docs
 slurp-docs:
 	python3 $(SCRIPTSDIR)/migrate.py --docs
 
-.PHONY: slurp-all-no-updates
-slurp-all-no-updates: $(foreach n,$(ALL_COMPONENT_IDS), slurp-no-updates-$(n))
-	@echo "$@ ($^) completed".
```
joeflack4, line 499: ## Deletion of unused goals
Deletes:
- `slurp-all-no-updates`
- `slurp-no-updates-%`

Thes currently do the exact same thing as `slurp-all` and `slurp-%`. They're not called by any other goals, so they're useless.

I am trying to remember what these goals were originally for.

It may be that they were a variation of running the slurp that did not involve doing some `pip install`s beforehand. The `pip install` would have been because ODK did not have high enough versions of some libraries at the time, likely OAK or maybe `sssom-py`.

I also have some recollection of `-B` being part of / necessary for some goal, maybe for part of the `slurp` pipeline, and this `-no-updates` variation would have been the same thing, but without `-B`.

Either way, whatever purpose they had appears to no longer be valid. They're doing nothing here, so I'm deleting them.

joeflack4, line 499: ## Skip build QC for this one?
This seems like a good case of skipping the need to run a build, because nowhere in `build-mondo-ingest` are these goals called. Deletion of them should have no effect on the build.

matentzn (approved):
If your comments are correct lets merge

joeflack4:
@matentzn I stand by the comments.
Looks like I need to resolve conflicts before merging.

Just need @twhetzel 's eye. @Trish, in summary, this PR removes some goals that are not part of the build and which we really didn't end up using and don't need

twhetzel:
Since this has been open for awhile and seemingly not a significant change it can be included in the January release.

src/ontology/mondo-ingest.Makefile
twhetzel: In this very specific and limited case, it's ok to not run a test build for this change.

twhetzel (approved):
ok to remove these goals

PR #534, opened 2024-05-21
Bugfix: slurp docs `parents` column

joeflack4:
Resolves #ISSUE(s).

## Overview

This PR:
- One
- Two
- Three

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [ ] No

- [ ] No

- [ ] Yes


src/scripts/migrate.py
```diff
@@ -154,6 +154,8 @@ def slurp_docs():
         ontology_name = os.path.basename(path).replace(FILENAME_GLOB_PATTERN[1:], '')
         ontology_page_relpath = f'./migrate_{ontology_name.lower()}.md'
         df = pd.read_csv(path, sep='\t').fillna('')
+        # Fix issue where the | in the `parents` column is causing table rendering issues
+        df['parents'] = df['parents'].apply(lambda x: x.replace('|', ','))
```
joeflack4, line 158: ## The bug fix
@twhetzel I ran this successfully w/out error. Let me know when you've reviewed, and I can run the full build for this per protocol.

src/scripts/migrate.py
joeflack4, line 1: ## Outputs
I didn't commit the `.md` files that get output into `docs/reports/`, as the difference generated would be more than just the fix I've implemented. The inputs I have are different than the last time when those were run, so there would be other differences in the tables.

@matentzn @twhetzel I've opted for (a), but let me know your thoughts on showing outputs for this bug fix:
a. Don't commit `.md` outputs along w/ this PR
b. If either of you do want me to run and commit the outputs, as this is just getting merged into `develop`, I can do that. 
c. If we run the full build, then those outputs will come with it, and be up-to-date.
d. Alternatively, if you want I can attach static files of the outputs to this comment or show a screenshot.

src/scripts/migrate.py
joeflack4: @twhetzel Forgot this was still open.

I have a comment saying:
>         # Fix issue where the | in the `parents` column is causing table rendering issues

To clarify, there was a jaggedness to the table that was being caused by this problem:
- Markdown tables use `|` as a cell delimiter
- In ROBOT templates, we usually use `|` as a delimiter

This collision was causing the Markdown tables to render multiple cells when there were multiple parents. You can see it if you look in some of the migrate `.md` files. This PR fixes that.

twhetzel:
Since this PR has been open for awhile and is a small change, it's ok to add into the January release.

twhetzel (approved):
Approving as you say this fixes the issue.

joeflack4:
@matentzn Just you now, unless you want to defer to Trish & I.

twhetzel:
@joeflack4 can this be merged into `develop` now?

joeflack4:
@twhetzel I leave it up to you. Refer to my comment right above yours; that's the only reason I didn't merge.

twhetzel:
For just fixing the formatting, it's fine without Nico's review. Was there anything else to check here?

PR #537, opened 2024-05-22
Build 2024-05-22 (ran with error)

twhetzel:
This ended with an error status:
```
Traceback (most recent call last):
  File "/work/src/ontology/../scripts/sync_subclassof.py", line 502, in <module>
    cli()
  File "/work/src/ontology/../scripts/sync_subclassof.py", line 481, in cli
    sync_subclassof(**d)
  File "/work/src/ontology/../scripts/sync_subclassof.py", line 313, in sync_subclassof
    raise ValueError(
ValueError: Detected error in consistency of sets of terms gathered from Mondo.

 1. Mondo SCR ancestors: 257177
 2. Mondo direct SCR relationships: 33484
 3. Mondo indirect SCR relationships: 223743
 Intersection (Top 5): [('MONDO:0006013', 'rdfs:subClassOf', 'MONDO:0700053'), ('MONDO:0033954', 'rdfs:subClassOf', 'MONDO:0100004'), ('MONDO:0968981', 'rdfs:subClassOf', 'MONDO:0019588'), ('MONDO:0025485', 'rdfs:subClassOf', 'MONDO:0700053'), ('MONDO:0958299', 'rdfs:subClassOf', 'MONDO:0958160')]
 "1" should be same as "2" + "3", but instead it has n less rels: 50
make[1]: *** [mondo-ingest.Makefile:536: reports/doid.subclass.direct-in-mondo-only.tsv] Error 1
make[1]: Leaving directory '/work/src/ontology'
make: *** [mondo-ingest.Makefile:334: build-mondo-ingest] Error 2
Command exited with non-zero status 2
```

PR #542, opened 2024-05-24
Build 2024-May-23 (re-run build)

twhetzel:
the build was run as: sh run.sh make build-mondo-ingest -B

this was re-run due to the failure in https://github.com/monarch-initiative/mondo-ingest/pull/537

### Documentation
- [ ] Yes

### QC
- [x] Yes
   
### New Packages
- [x] No

- [x] No

- [x] Yes


src/ontology/lexmatch/README.md
joeflack4, line 1: ## What's this build for?
Is this a routine build or is this build linked to another PR or issue?

joeflack4: Also, I wrote a thread in a recent PR, IDR if I got a response, but I'm generally not sure of what to do when reviewing builds. I mostly just face check slurp and lexmatch, and just skim through it. I spend like 1-2 minutes on that. If anyone knows of a better practice, let me know.

matentzn: Maybe add a documentation page "How to review a data build" to the docs:

1. The data build is not reviewed for specific changes in content, but general patterns of changes. It is recommended to spend 10 minutes reviewing each databuild.
2. There are two important reasons to review databuilds: (1) looking out for large unexplainable changes and (2) increasing your familiarity with the data generated by the pipeline. The later is as important as the former: as data stewards in the Mondo Ingest pipeline you should understand all data (every single file!) that is generated insight out, and the best way to do that is to review each file many times until it sticks. It is not wrong to use a data release to ask questions like: "what is the purpose of this file?".
3. Ensure that no files are added or removed. There are few good reasons for files being added or removed and if they happen they should be explained.
4. ORDO, DOID and OMIM matches and migration files should have "reasonable" changes, i.e. be in line with one could expect as a consequence of a few weeks worth of curation (example: 1000 added lines is not a good sign, but 70 removed lines is within reason).
5. Metrics and ontology related files should change within reason (numbers like axiom counts changing in the realms of 250 plus minus are nearly always ok, changes between 250 and 1000 are worth a second look, and changes beyond 1000 merit an investigation).
6. No file should be totally empty in a data sync.


----------

From your perspective, then, you could have noticed the massive changes in NCIT, and I would have told you that there had not been an NCIT release for over a year, now there was one, and the big change is very much expected.

matentzn (approved):
Looks great, @joeflack4 maybe make issue with the one comment I made under yours!

src/ontology/lexmatch/README.md
joeflack4: Thanks, this guidance is beyond what I expected! Been really wondering about this since we've been more active with builds. I think it's a good idea to have a page for this. I created the issue:
- #544

There are a few docs issues piling up, so we can probably efficiently do them in a batch; like I think there's a doc-athon coming up IIRC.

joeflack4: There are deleted files in this PR, but it looks OK.

I noticed that there are 2 deleted files in this diff. They're both OMIMPS. I investigated, thinking that this must be because of the `www` issue. I have 2 PRs that address this, one in `omim`, and one in `mondo-ingest`, and both are basically ready to merge. I just checked the diff in the `mondo-ingest` PR, and was relieved to see that the files are not deleted there: https://github.com/monarch-initiative/mondo-ingest/pull/520/files

joeflack4 (approved):

src/ontology/lexmatch/split-mapping-set/mondo_broadmatch_omimps.tsv
joeflack4, line 1: Deleted file

joeflack4 (changes requested):

src/ontology/lexmatch/split-mapping-set/mondo_broadmatch_omimps.tsv
matentzn: Good catch! Could have been OMIMPS, but it this case it likely is not. My guess is it is related to https://github.com/monarch-initiative/mondo/pull/7563! Do you understand how the two might be related?

joeflack4: I see; I didn't know about that. In that case, I'm not sure which is the cause of OMIMPS not appearing here. If it is caused by #&563, does this (also) mean that the removal of these OMIMPS file is not a problem?

matentzn: Again this is a better of judgement. Here is how my brain thinks of it:

1. If you are right (and it is related to the prefix), then _all_ omimps related files (in particular the "exact" ones) would be removed.
2. If I am right, only the non-exact ones (close, broad, etc) would be removed.

So I would check if the "exact" one is still there!

twhetzel: Screenshot 2024-05-26 at 10 48 22 AM

joeflack4: Ah OK, so looks like no problem then!

@twhetzel I think you said that this build is no longer relevant and was replaced by:
- #548 

So I'm going to close this. Feel free to re-open if I am incorrect about that!

PR #543, opened 2024-05-24
Create ORDO subset ROBOT template - build

joeflack4:
Build for:
- #510 

@matentzn I pinged you [on slack](https://monarchinitiative.slack.com/archives/C03AW548J8M/p1716579904321739) per your request.
CC @twhetzel


src/ontology/external/nord.robot.owl
matentzn, line 1: Is it wrong to expect a change like this one here to the ORDO subset this PR is about?

matentzn (changes requested):

src/ontology/external/nord.robot.owl
joeflack4: Probably not, I'm guessing. Remember that this is a `build-mondo-ingest` QC PR; it's not something that will be merged.

I didn't get a chance to review the NORD PR though, so I'm not sure how the pipeline works yet.  But, taking a look now... yes, it is an expected change. The only change is:
```diff
-        <owl:versionIRI rdf:resource="http://purl.obolibrary.org/obo/mondo/external/2024-05-01/nord.robot.owl"/>
+        <owl:versionIRI rdf:resource="http://purl.obolibrary.org/obo/mondo/external/2024-05-24/nord.robot.owl"/>
```
Which is a normal result of the `make` goal: `--version-iri $(URIBASE)/mondo/external/$(TODAY)/nord.robot.owl`

joeflack4: Thinking about this, if we wanted less spurious diffs, we could replace `$(TODAY)` in various goals with some new variable `ARTEFACT_RELEASE_DATE`, but I don't know if there's a reliable way to gauge the release date of any given artefact unless it is listed publicly. This is related to the whole issue of `wget`s being the source of many pipelines, and without a way to check the hash of the file before we download it, we end up triggering the whole pipeline.

matentzn: Sorry my comment was made on the NORD file - my question was pertaining to the subject of the the PR: the Ordo subset file. The full pipeline run should generate that file, and I am noting that the file does not seem to be there! Are wer sure a full pipeline run will in fact generate the Ordo subset file the same way as it does the Nord one?

matentzn: (And I don't mind the date in the diff, it gives me confidence the files were indeed updated)

joeflack4: Ahhh. That's funny. The reason it's not running is because I forgot that we were going to add it to `build-mondo-ingest` as part of:
- #515

In order to fix this right now, I needed to complete 2/3 of the sub-tasks in that issue, so I went ahead and just did all 3 anyway.

Strangely, I encountered an issue when running the build on a fresh clone:
- #545 

So for now I'm re-running the build from my development environment using `-B`.

joeflack4: Build complete! `external/ordo-subsets.robot.template.tsv` is now included.

src/ontology/external/ordo-subsets.robot.template.tsv
matentzn, line 1: Similar to nord.robot.tsv, I would expect src/ontology/external/ordo-subsets.robot.template.owl.

matentzn (changes requested):
Small remaining issue! Else looks good.

src/ontology/external/ordo-subsets.robot.template.tsv
joeflack4: - [ ] @matentzn clarify
Are you sure about this? I don't see why.

There is no `ordo.robot.tsv` that is created via the same pipeline as `nord.robot.tsv`. It's a different kind of output w/ a different way of making it.

Therefore, I had originally coded up a different way of adding this to the build. I had added a `source-specific-modifications` goal, w/ `source-specific-modifications-ordo`, which kicked off this pipeline. I had thought of "externally managed content" as being "the source is ~100% externally managed", so I didn't consider ORDO a fit for this, but I've since changed my mind and I think it that the "~100%" qualifier doesn't matter. In any case, the main reason I added it to `update-externally-managed-content` is because of your request in [this comment](https://github.com/monarch-initiative/mondo-ingest/pull/510#pullrequestreview-2039368784).

I hope I don't have to change the whole way that I'm doing this ORDO subsets to be like NORD. I think probably you are not insisting on it; you just don't remember the internals of how I'm creating these outputs. For a refresher, you can look at #510. But if you really do want me to do this I will have to update #510 and I don't know if it will be a quick change!

src/ontology/external/ordo-subsets.robot.tsv
joeflack4, line 1: ## File rename
I renamed this to remove the `.template` part of the filename to be in line w/ the changes made in:
- #546

src/ontology/external/ordo-subsets.robot.owl
joeflack4, line 1: ## Created `external/ordo-subsets.robot.owl`
@matentzn added this goal in the following merged PR, so I updated this PR to include it in the build. I didn't re-run the full build; I just ran the goal.
- #546

src/ontology/external/ordo-subsets.robot.template.tsv
joeflack4: - [ ] @matentzn Approve this build PR if it looks good now, and we can close it
- [ ] @matentzn Re-review #510 and, if looks good, we can merge it

Yep, I definitely misunderstood. Thanks for the PR; that's probably the most efficient way.

I still have some lack of clarity though about the things in my comment above. As I mentioned, I was wondering if the `ordo-subsets` stuff even belonged in `update-externally-managed-content` because, unlike NORD, our ORDO integration is mostly internally managed, I'd say.

Further, I don't know if there is any purpose for `external/ordo-subsets.robot.owl`. As far as I was aware, only the robot template was needed.

Despite my lack of full understanding, I merged #546 because you have more knowledge than me about the bigger picture around these changes, and I trust that these changes make sense. In the unlikely event that you were wrong about something here (I trust you are not), I can always fix in a new commit.

matentzn: > Further, I don't know if there is any purpose for external/ordo-subsets.robot.owl. As far as I was aware, only the robot template was needed.

If you ask me about this in a meeting I will show you! For externally managed content, I use the OWL files rather than the tables. This ensures that any malformed data tables (i.e. those that do not correctly convert to OWL) are caught at mondo-ingest, and I don't have to go on expensive quests for fixing them.

src/ontology/external/ordo-subsets.robot.tsv
matentzn: thx

matentzn (approved):
LGTM!

src/ontology/external/ordo-subsets.robot.template.tsv
joeflack4: Hmm, I'll add this to my epic rolling list. Converting to OWL as a validation step at the very least makes some sense; I don't see a `validate` command in `robot template`. And I could kind of see how you might use Protégé to interrogate. But I would not expect this template to be malformed since we created it, and there's not a lot to inspect here with these subsets. Perhaps this has to do with being able to spot if the assigned subsets don't make any sense. In any case I'll stop speculating and save it for a feuture demo!

PR #548, opened 2024-05-28
Update ICD11 relevant SPARQL query to exclude classes in the Extension Code branch

### LINKED ISSUE #547: Update ICD11 relevant SPARQL query to exclude classes in the Extension Code branch
Update `src/sparql/icd11foundation-relevant-signature.sparql` to exclude any class that is a child of the Extension Code branch.

joeflack4:
Resolves #547

## Overview
Updates ICD11 SPARQL to exclude extension code branch.

### Documentation
- [ ] Yes

### QC
Build: 
- #549
- [x] Yes
   
### New Packages
- [x] No

- [x] No

- [x] Yes


src/sparql/icd11foundation-relevant-signature.sparql
```diff
@@ -15,4 +14,7 @@ WHERE {
     }
   }
   FILTER(isIRI(?term))
+  FILTER NOT EXISTS {
+    ?term rdfs:subClassOf* <http://id.who.int/icd/entity/979408586> .
```
joeflack4, line 18: ## The change in question
All other changes to the SPARQL are cosmetic.

joeflack4, line 18: ## Running build
Currently I'm running the build, for the workflow:
`main` <-- this PR <-- build PR

When the build is done I'll open a PR for it and link that here.

src/sparql/icd11foundation-relevant-signature.sparql
```diff
@@ -1,12 +1,11 @@
-prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
 
-### All diseases
```
twhetzel, line 3: nit: This could be something like all subclasses of "ICD Category" except classes that are also subclasses of "Extension Codes"

src/sparql/icd11foundation-relevant-signature.sparql
twhetzel: I don't follow the PR path here. The eventual endpoint here should be a new build run in `main` that has the ICD11 Foundation content mapped from content created with this new SPARQL query as well as any other new files pulled from the `mondo` repo, e.g. to account for the latest curation Sabrina did for DO

twhetzel (approved):
**PR Review Comment**
query looks good. i don't follow the PR path, but as long as it gets us to what i described in the comment should be all good!

src/sparql/icd11foundation-relevant-signature.sparql
joeflack4: Yep, that will be the result! I created this branch off of `main`, so this workflow will have the exact same results, and I could be wrong but I think it is more in-line with the SOP. The workflow has the benefit of not causing any issues on `main` in the event of something being wrong or unexpected with the changes.

src/sparql/icd11foundation-relevant-signature.sparql
joeflack4: Once the build is done I can make this change.

src/sparql/icd11foundation-relevant-signature.sparql
joeflack4: Build done!
- #549

However:
_Trish wrote:_
> Let's also make sure there is a `unmapped_icd11foundation_lex_exact.tsv` file in the results from the pipeline since I do not see that file in the build I ran last week.

I'm not seeing that file in the outputs ;(. Do you want to take a preliminary look into why this may be the case? I won't be able to look at this until tomorrow afternoon, but I can look into it then.

twhetzel: The `unmapped_icd11foundation_lex_exact.tsv` is in this build.

joeflack4: Oh, so it is. It was late and I was heading to bed, and I ls'ed the `unmapped/` directory instead of the `lexmatch/` lol.

PR #551, opened 2024-05-31
Intelligently refresh `mondo.sssom.tsv` & `mondo.owl`

### LINKED ISSUE #538: Intelligently refresh `mondo.sssom.tsv` & `mondo.owl`
## Overview
In non-build situations, goals that use `tmp/mondo.sssom.tsv` will often an outdated version. This is because it is dependent on `tmp/mondo_repo_built`, which is not triggered in the normal `make` way. Normally a `make` goal runs if the target of the goal hasn't been created, or if any of its prerequisites are newer than the target. In this case, one of the prerequisites is the `mondo` repo itself.

During a fresh build, this will not be a problem. It'll only be a problem in development situations.

## Possible solutions
Originally we were considering other solutions, but we chose to go with (b).

Both options involve at least this change:
Move dependency into the goal's body. In this example, that would change it to:

```make
reports/ordo-subsets.robot.template.tsv: tmp/ordo-subsets.tsv 
	$(MAKE) tmp/mondo.sssom.tsv
	python3 $(SCRIPTSDIR)/ordo_subsets.py \
	...
```

### a. ~Always run a full refresh of `tmp/mondo.sssom.tsv`~
<p>
https://github.com/monarch-initiative/mondo-ingest/issues/529#issuecomment-2123315334

Always run a refresh each time a goal is run which has it as a dependency.

This involves nothing else in addition to the change I just described above.

This could be be a real damper on local development. Not always terribly so, though. When I'm developing a goal that uses Python, I usually do all of my iteration in the debugger, and then run the make goal once at the end just to make sure it works.

</p>
</details>

### b. Intelligently refresh `tmp/mondo.sssom.tsv` when needed
This would be cool to do at some point. Basically, in addition to the change I described above, we'd have an if statement at the top of the goal for `tmp/mondo.sssom.tsv` or `tmp/mondo_repo_built`. It would compare the latest git commit hashes on `master` for  the cloned `mondo` dir in `tmp/` and the actual repo on GitHub. If they're the same, do nothing. Else, it will re-run `tmp/mondo_repo_built` --> `tmp/mondo.sssom.tsv`.

## Additional info
Original context: https://github.com/monarch-initiative/mondo-ingest/pull/531#discussion_r1609048542

joeflack4:
Resolves #538

## Overview
Updates to make it so that `mondo.sssom.tsv` (and also `mondo.owl`) get refreshed when the files get updated in the mondo repo. Prior to these changes, you'd have to manually refresh using -B or otherwise be at risk of using stale inputs for these files.

### Documentation
- [ ] Yes

### QC
- [x] Yes
   
Build PR:
- #586

### New Packages
- [x] No

- [x] No

- [x] Yes

### DIFF
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -219,7 +219,8 @@ unmapped-terms-docs: $(foreach n,$(ALL_COMPONENT_IDS), reports/$(n)_unmapped_ter
 .PHONY: unmapped-terms-tables
 unmapped-terms-tables: $(foreach n,$(ALL_COMPONENT_IDS), reports/$(n)_mapping_status.tsv)
 
-$(REPORTDIR)/%_mapping_status.tsv $(REPORTDIR)/%_unmapped_terms.tsv: $(REPORTDIR)/%_term_exclusions.txt $(TMPDIR)/mondo.sssom.tsv metadata/%.yml $(COMPONENTSDIR)/%.db
+$(REPORTDIR)/%_mapping_status.tsv $(REPORTDIR)/%_unmapped_terms.tsv: $(REPORTDIR)/%_term_exclusions.txt metadata/%.yml $(COMPONENTSDIR)/%.db
+	$(MAKE) up-to-date-mondo.sssom.tsv
 	python3 $(SCRIPTSDIR)/unmapped_tables.py \
 	--exclusions-path $(REPORTDIR)/$*_term_exclusions.txt \
 	--mondo-mappings-path $(TMPDIR)/mondo.sssom.tsv \
@@ -257,7 +258,9 @@ $(REPORTDIR)/%_term_exclusions.txt $(REPORTDIR)/%_exclusion_reasons.robot.templa
 $(REPORTDIR)/%_exclusion_reasons.ttl: component-download-%.owl $(REPORTDIR)/%_exclusion_reasons.robot.template.tsv
 	$(ROBOT) template --input $(TMPDIR)/component-download-$*.owl.owl --add-prefixes config/context.json --template $(REPORTDIR)/$*_exclusion_reasons.robot.template.tsv --output $(REPORTDIR)/$*_exclusion_reasons.ttl
 
-$(REPORTDIR)/%_excluded_terms_in_mondo_xrefs.tsv $(REPORTDIR)/%_excluded_terms_in_mondo_xrefs_summary.tsv: $(TMPDIR)/mondo.sssom.tsv tmp/mondo.owl metadata/%.yml $(REPORTDIR)/component_signature-%.tsv $(REPORTDIR)/mirror_signature-%.tsv
+# todo: Should this also be a prereq $(TMPDIR)/component-download-$*.owl.owl? Perhaps worried about refreshing when not need to? But then we'd just use COMP=false if so?
+$(REPORTDIR)/%_excluded_terms_in_mondo_xrefs.tsv $(REPORTDIR)/%_excluded_terms_in_mondo_xrefs_summary.tsv: metadata/%.yml $(REPORTDIR)/component_signature-%.tsv $(REPORTDIR)/mirror_signature-%.tsv
+	$(MAKE) up-to-date-mondo.sssom.tsv
 	python3 $(RELEASEDIR)/src/analysis/problematic_exclusions.py \
 	--mondo-mappings-path $(TMPDIR)/mondo.sssom.tsv \
 	--onto-path $(TMPDIR)/component-download-$*.owl.owl \
@@ -356,23 +359,43 @@ deploy-mondo-ingest:
 
 USE_MONDO_RELEASE=false
 
+# Builds tmp/mondo/ and rebuilds mondo.owl and mondo.sssom.tsv, and stores hash of latest commit of mondo repo main branch in tmp/mondo_repo_built
 tmp/mondo_repo_built:
 	if [ $(USE_MONDO_RELEASE) = true ]; then wget http://purl.obolibrary.org/obo/mondo.owl -O $@; else cd $(TMPDIR) &&\
 		rm -rf ./mondo/ &&\
 		git clone --depth 1 https://github.com/monarch-initiative/mondo &&\
 		cd mondo/src/ontology &&\
 		make mondo.owl mappings -B MIR=false IMP=false MIR=false &&\
-		cd ../../../../ &&\
-		touch $@; fi
+		latest_hash=$$(git rev-parse origin/master) &&\
+		echo "$$latest_hash" > $@ &&\
+		cp $@ mappings/mondo.sssom.tsv mondo.owl ../../../; fi
 
-$(TMPDIR)/mondo.sssom.tsv: $(TMPDIR)/mondo_repo_built
-	cp $(TMPDIR)/mondo/src/ontology/mappings/mondo.sssom.tsv $@
+# Triggers a refresh of tmp/mondo/ and a rebuild of mondo.owl and mondo.sssom.tsv, only if mondo repo main branch has new commits
+.PHONY: refresh-mondo-clone
+refresh-mondo-clone:
+	@if [ ! -d tmp/mondo ]; then \
+		$(MAKE) tmp/mondo_repo_built -B; \
+	else \
+		current_hash=$$(cat tmp/mondo_repo_built); \
+		cd tmp/mondo; \
+		git fetch origin; \
+		latest_hash=$$(git rev-parse origin/master); \
+		if [ ! "$$current_hash" = "$$latest_hash" ]; then \
+			cd .. && mv mondo mondo-bak && mv mondo_repo_built mondo_repo_built-bak; \
+			cd .. && $(MAKE) tmp/mondo_repo_built -B; \
+			rm -rf tmp/mondo-bak tmp/mondo_repo_built-bak; \
+		fi; \
+	fi
 
-$(TMPDIR)/mondo.owl: $(TMPDIR)/mondo_repo_built
-	cp $(TMPDIR)/mondo/src/ontology/mondo.owl $@
+.PHONY: up-to-date-mondo.sssom.tsv
+up-to-date-mondo.sssom.tsv: refresh-mondo-clone
 
-reports/mirror_signature-mondo.tsv: tmp/mondo.owl
-	$(ROBOT) query -i $< --query ../sparql/classes.sparql $@
+.PHONY: up-to-date-mondo.owl
+up-to-date-mondo.owl: refresh-mondo-clone
+
+reports/mirror_signature-mondo.tsv:
+	$(MAKE) up-to-date-mondo.owl
+	$(ROBOT) query -i tmp/mondo.owl --query ../sparql/classes.sparql $@
 	(head -n 1 $@ && tail -n +2 $@ | sort) > $@-temp
 	mv $@-temp $@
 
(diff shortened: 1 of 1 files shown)


src/ontology/mondo-ingest.Makefile
```diff
@@ -351,14 +351,46 @@ deploy-mondo-ingest:
...
 		make mondo.owl mappings -B MIR=false IMP=false MIR=false &&\
+		latest_hash=$$(git rev-parse origin/master) &&\
 		cd ../../../../ &&\
-		touch $@; fi
+		echo "$$latest_hash" > $@; fi
+
+# TODO
+# explain what does here. explain that -B is just in case tmp/mondo_repo_built exists but not the rpeo
+.PHONY: refresh-mondo-clone
+refresh-mondo-clone:
```
joeflack4, line 370: ## Preliminary results
@matentzn @twhetzel So far I got this working. It was a little trickier than I'd hoped; the intersecting of ShellScript and makefile always feels a bit frustrated.

So far these (4) conditions seem to be working:

1. If mondo repo clone does not exist in `tmp/`, makes it
2. If it exists and is up to date, do nothing
3. If it exists, and I've manually changed the hash in `tmp/mondo_repo_built`, it re-creates the mondo clone
4 If it exists, and latest commit hash on `mondo` `origin/master`  is different than what is in `tmp/mondo_repo_built`, it re-creates the mondo clone

My next task is to put `$(MAKE) refresh-mondo-sssom` or `$(MAKE) refresh-mondo-owl` in the bodies of the respective goals that use them, and remove `monso.sssom.tsv` and `mondo.owl` from the prereqs of those goals.

- [ ] @joeflack4 Maybe I should rename the `refresh-` in `refresh-mondo-sssom` etc to `updated-` or `up-to-date`. Because it won't always refresh if the thing is up to date; it's having an up-to-date file that is the idea.

twhetzel: The goal name is fine with with me. I take `refresh-` to mean do something to get the latest version available.

matentzn: Fine. Per irrelevant conventions I tended to use:

1. Refresh for imports
2. Recreate for components
3. Update for anything else

This is not very consistent, so you go with what you like

src/ontology/mondo-ingest.Makefile
```diff
@@ -351,14 +351,46 @@ deploy-mondo-ingest:
...
+			cd .. && $(MAKE) tmp/mondo_repo_built -B; \
+			rm -rf tmp/mondo-bak tmp/mondo_repo_built-bak; \
+		fi; \
+	fi
+
+.PHONY: refresh-mondo-owl
+refresh-mondo-owl: refresh-mondo-clone
+
+.PHONY: refresh-mondo-sssom
+refresh-mondo-sssom: refresh-mondo-clone
```
matentzn, line 389: Note if you seperate the two, you might increase runtime, as the build may (may) run some of the same goals required for the mondo.owl and mondo.sssom.tsv multiple times.

joeflack4: I'm wondering how this would increase how often `refresh-mondo-clone` would be called, but it shouldn't matter much, because this PR does two things:
1. Makes sure these files get updated whenever needed.
2. Makes sure they do not get recreated is they are up-to-date.

Regarding (2),as discussed in [the issue](https://github.com/monarch-initiative/mondo-ingest/issues/538), this goal will not do anything if the latest commit hash on `main` hasn't changed. So, if `refresh-mondo-clone` does get run more often than it should, the only downside here is dependency on a network connection, for the `git fetch` command. There is latency in running that command, but it is nominal.

matentzn: mondo.owl and mondo.sssom.tsv share some of the same targets in the Mondo (not mondo ingest!) make pipeline and they may be run twice. For example, both depend on the goal "filtered.obo". It is possible that this will add 5 minutes of runtime, but it is also possible no.

joeflack4: Hmm, I guess it's not worth worrying about too much right now. I'll just say that I don't think that will be the case. I'm basically not changing how any of the goals in `mondo-ingest.Makefile` relate to `mondo.owl` and `mondo.sssom.tsv`, nor how those relate to `tmp/mondo/` / `tmp/mondo_repo_built`. Further, I kept `refresh-mondo-clone` (which is the new `tmp/mondo_repo_built` goal) the same as it was before, in that it runs `make mondo.owl mappings -B MIR=false IMP=false MIR=false &&\`.

Therefore, whether `refresh-mondo-sssom` or `refresh-mondo-owl` runs doesn't matter; both of these files will be built together whenever one of them is called.

And, of course, subsequent calls of these goals will do nothing so long as no new commits have happened in `mondo`.

joeflack4, line 389: ## Unlikely scenario: new commit mid-build
There is an unlikely but possible scenario where, if a new commit goes into `mondo` during the runtime of the build, it's possible that different versions of `mondo.owl` and `mondo.sssom.tsv` could be built at different times in the build, and this could introduce discrepancies.

I can think of a solution to this, where I make the command more intelligent--not only would it check for new commits, but timestamp of the commit would also have to be older than when the build started. And if a new commit was detected, it could throw an error or a warning.

But I think this is probably a low priority. I guess I could make an issue about handling this contingency, perhaps.

src/ontology/mondo-ingest.Makefile
```diff
@@ -398,8 +421,9 @@ signature_reports: $(ALL_MIRROR_SIGNTAURE_REPORTS) $(ALL_COMPONENT_SIGNTAURE_REP
 # - Doeesn't include: broad mappings
 # - Comes from make tmp/mondo.owl
 
-tmp/mondo.sssom.ttl: tmp/mondo.sssom.tsv
-	sssom convert $< -O rdf -o $@
+tmp/mondo.sssom.ttl:
+	$(MAKE) up-to-date-mondo.sssom.tsv
```
joeflack4, line 425: ## Usage of new goals
The prereqs `tmp/mondo.owl` and `tmp/mondo.sssom.tsv` have been moved from the goal prereqs position into the body of the goal and called using `$(MAKE)`. This ensures that the goal isn't run all of the time, even if nothing has changed (as that is what happens if you have a `.PHONY` goal as a prereq).

Also fixed any references that would have been broken due to positional syntax, e.g. `$<`.

src/ontology/mondo-ingest.Makefile
joeflack4, line 1: ## Considerations: Goals always run when`mondo` updates
This PR does not ensure that, should an update to the `mondo` repo happens, all goals which use anything that comes from the `mondo` repo (i.e. `mondo.sssom.tsv` and `mondo.owl`) will always run.

It only ensures that **if** the goals would run (i.e. because (a) the target had never been created, or (b) one of its other prereqs was newer than the target), then `mondo.sssom.tsv` and `mondo.owl` will get updated first.

- [ ] @matentzn @twhetzel **Make an issue?**: Is this a problem significant enough that it merits a low priority issue? I can't even think how to elegantly solve it.

### Special cases
There are some cases of goals that we might consider are so "strongly coupled" to either `mondo.sssom.tsv` or `mondo.owl`, that we should also make a similar phony goal wrapper. For example, `tmp/mondo.db`. Its only prereq is `tmp/mondo.owl`. So perhaps, as we have made a goal `up-to-date-mondo.owl`, we could also make a goal `up-to-date-mondo.db` that will ensure that any time a goal needs to use `mondo.db`, it is similarly ensured to be updated. It may be that this is the only special case worth considering.

- [ ] @matentzn @twhetzel **Do only this special case: `mondo.db`**? I could add to this PR; not much work.

src/ontology/mondo-ingest.Makefile
```diff
@@ -351,23 +354,43 @@ deploy-mondo-ingest:
 
 USE_MONDO_RELEASE=false
 
+# TODO: explain what does here
```
joeflack4, line 357: ## Comments as documentation
- [ ] @joeflack4 todo

I undrafted the PR because this is quick and otherwise it is ready. I'll just add some comments about what `tmp/mondo_repo_built` and `refresh-mondo-clone`, just because they look really confusing and are otherwise idiosyncratic.

joeflack4: done

src/ontology/mondo-ingest.Makefile
```diff
@@ -355,23 +358,43 @@ deploy-mondo-ingest:
...
-$(TMPDIR)/mondo.owl: $(TMPDIR)/mondo_repo_built
-	cp $(TMPDIR)/mondo/src/ontology/mondo.owl $@
-
-reports/mirror_signature-mondo.tsv: tmp/mondo.owl
-	$(ROBOT) query -i $< --query ../sparql/classes.sparql $@
+		latest_hash=$$(git rev-parse origin/master) &&\
+		echo "$$latest_hash" > $@ &&\
+		cp $@ mappings/mondo.sssom.tsv mondo.owl ../../../; fi
+
+# Triggers a refresh of tmp/mondo/ and a rebuild of mondo.owl and mondo.sssom.tsv, only if mondo repo main branch has new commits
```
joeflack4, line 372: ## Doc comments
Added comments to `refresh-mondo-clone` and `tmp/mondo_repo_built`, explaining what they do.

src/ontology/mondo-ingest.Makefile
joeflack4, line 1: ## QC build
- [ ] 1. @twhetzel (or @matentzn) review and approve
- [ ] 2. @joeflack4 then, will run a build

src/ontology/mondo-ingest.Makefile
```diff
@@ -355,23 +358,43 @@ deploy-mondo-ingest:
...
+			cd .. && $(MAKE) tmp/mondo_repo_built -B; \
+			rm -rf tmp/mondo-bak tmp/mondo_repo_built-bak; \
+		fi; \
+	fi
+
+.PHONY: up-to-date-mondo.sssom.tsv
+up-to-date-mondo.sssom.tsv: refresh-mondo-clone
+
+.PHONY: up-to-date-mondo.owl
+up-to-date-mondo.owl: refresh-mondo-clone
```
joeflack4, line 393: ## Add goal `up-to-date-mondo-edit.owl`?
This is a new file that depends on `tmp/mondo/`. See:
- https://github.com/monarch-initiative/mondo-ingest/pull/568#pullrequestreview-2119425130

twhetzel (commented):
I'm fine with these changes and would like to see the feature build data.

src/ontology/mondo-ingest.Makefile
twhetzel: This PR should only address the items discussed, `mondo.sssom.tsv` & `mondo.owl`. If there are cases where `tmp/mondo.db` is not being updated as needed add a ticket for that and create a light weight design doc in the Google Drive Mondo-->Technical Issues to explain the issue, how you plan to solve it, and any questions we should address before working on the code.

src/ontology/mondo-ingest.Makefile
```diff
@@ -256,7 +257,9 @@ $(REPORTDIR)/%_term_exclusions.txt $(REPORTDIR)/%_exclusion_reasons.robot.templa
 $(REPORTDIR)/%_exclusion_reasons.ttl: component-download-%.owl $(REPORTDIR)/%_exclusion_reasons.robot.template.tsv
 	$(ROBOT) template --input $(TMPDIR)/component-download-$*.owl.owl --add-prefixes config/context.json --template $(REPORTDIR)/$*_exclusion_reasons.robot.template.tsv --output $(REPORTDIR)/$*_exclusion_reasons.ttl
 
-$(REPORTDIR)/%_excluded_terms_in_mondo_xrefs.tsv $(REPORTDIR)/%_excluded_terms_in_mondo_xrefs_summary.tsv: $(TMPDIR)/mondo.sssom.tsv tmp/mondo.owl metadata/%.yml $(REPORTDIR)/component_signature-%.tsv $(REPORTDIR)/mirror_signature-%.tsv
+# todo: Should this also be a prereq $(TMPDIR)/component-download-$*.owl.owl? Perhaps worried about refreshing when not need to? But then we'd just use COMP=false if so?
```
twhetzel, line 260: @joeflack4 is this still an active comment or has it been addressed?

joeflack4: Active comment; it's unrelated to this PR, just something I noticed. Sometimes I commit these comments directly to `develop`, as is better if I have the time. Lowercase `todo`, indicating it is not critical.

src/ontology/mondo-ingest.Makefile
joeflack4: - [ ] @joeflack4 create an issue

src/ontology/mondo-ingest.Makefile
twhetzel: ok, but this is a question not an action item.

joeflack4: Yes, basically the `todo` is to answer that question. I suppose almost anything with a `?` should be considered some temporary thing to be answered; not stay int he code. I find that adding `todo` makes this clearer; I can tell quickly which things are just plain explanatory comments, and which things need action. Part of that is because of IDE syntax highlighting.

PR #553, opened 2024-06-05
Fix ordo subset pipeline

matentzn:
The ORDO source for the subset query was set to be the ordo.owl component in Mondo Ingest. That was wrong, it should have been the orginal, unaltered ORDO.


twhetzel (approved):
lgtm

src/ontology/mondo-ingest.Makefile
```diff
@@ -571,8 +571,9 @@ $(EXTERNAL_CONTENT_DIR)/nord.robot.tsv: $(TMPDIR)/nord.tsv config/external-conte
 .PHONY: external-content-nord
 external-content-nord: $(EXTERNAL_CONTENT_DIR)/nord.robot.tsv $(EXTERNAL_CONTENT_DIR)/nord.robot.owl
 
-tmp/ordo-subsets.tsv: $(COMPONENTSDIR)/ordo.owl
-	$(ROBOT) query -i $< --query ../sparql/select-ordo-subsets.sparql $@
+tmp/ordo-subsets.tsv:
+	$(MAKE) component-download-ordo.owl
```
joeflack4, line 575: Ah, I see. This is because the component currently strips away the subset information, I think.

joeflack4 (approved):

src/ontology/mondo-ingest.Makefile
matentzn: What is more important: the component adds additional isa links that are not there in ORDO, which confuses the subset pipeline!

PR #554, opened 2024-06-05
`main` <- `develop`

joeflack4:
## Overview
Merging `develop` into `main`.

This is the conceptually `develop` into `main` where `main-new` was a branch from the latest version of `main`. It was done this way this time in order to run the build once.

### QC
- [x] Yes
   
Build PR:
- #556

- [x] Yes

## Additional info
Context: [Slack](https://monarchinitiative.slack.com/archives/C03AW548J8M/p1717531121525599)

Includes:
1. Merge commit from `develop` into `main-new`, which is a copy of `main`.
2. #555
3. #557
4. #556 - to be merged once that build finishes and is reviewed


src/ontology/mondo-ingest.Makefile
joeflack4, line 1: ## Development workflow
### The one this PR is using
A special pull request. This branch, `main-new`, is the product of checking out a copy of `main`, then merging `develop` into it. There will be a separate build PR which, when successful, will be merged into this, and then this will be merged into main.

Thus, after merging into main, the build will not need to be re-run. This development workflow allows us to run the build 1 time when merging `develop` into `main`, as opposed to this other workflow: (i) run build on develop, merge to main, (ii) run build on main.

### The SOP

twhetzel: While (1) does take more time, it's not active developer time and is more clear. In theory, I believe (3) should work, but is not as straightforward with git operations as (1). For now, I would have opted for (1) to keep things simple and understandable. However, I am also trying to recall some other workflows I've used that are similar to (3) and how they worked out.

joeflack4, line 1: ## Build review
- [ ] @matentzn / @twhetzel / @joeflack4 review build

- #556

matentzn: Agree with Trish!

src/ontology/mondo-ingest.Makefile
```diff
@@ -563,12 +573,40 @@ $(EXTERNAL_CONTENT_DIR)/%.robot.owl: $(EXTERNAL_CONTENT_DIR)/%.robot.tsv
 $(EXTERNAL_CONTENT_DIR)/nord.robot.tsv: $(TMPDIR)/nord.tsv config/external-content-robot-headers.json
 	mkdir -p $(EXTERNAL_CONTENT_DIR)
 	python ../scripts/add-robot-template-header.py $^ > $@
-.PRECIOUS: $(EXTERNAL_CONTENT_DIR)/nord.robot.owl
+.PRECIOUS: $(EXTERNAL_CONTENT_DIR)/nord.robot.tsv
 
 .PHONY: external-content-nord
+external-content-nord: $(EXTERNAL_CONTENT_DIR)/nord.robot.tsv $(EXTERNAL_CONTENT_DIR)/nord.robot.owl
+
+tmp/ordo-subsets.tsv:
```
matentzn, line 581: @joeflack4 this is good, alleviates my worry in the data build PR.  (i.e. its correct).

src/ontology/mondo-ingest.Makefile
joeflack4: It has been decided! The SOP shall stand.

twhetzel: I'm open to re-visiting in ~3 months and we can see how this current SOP is working and if things are any more settled since there's always some new unexpected thing popping up :)

twhetzel (approved):
As expected based on the review of https://github.com/monarch-initiative/mondo-ingest/pull/556, this Build also has the same issue. Just adding here to document for consistency. There is an [issue](https://github.com/monarch-initiative/mondo-ingest/issues/560) to address this.

PR #555, opened 2024-06-05
- Bug fix: slurp_ordo_mods() - was missing an arg in get_formatted_su…

joeflack4:
## Overview
- Bug fix: slurp_ordo_mods() - was missing an arg in get_formatted_subsets_df() following a refactor.

### QC
This is a fix mid build.

### DIFF
--- a/src/scripts/migrate.py
+++ b/src/scripts/migrate.py
@@ -12,7 +12,8 @@ Resources
 import os
 from argparse import ArgumentParser
 from glob import glob
-from typing import Dict, List, Set
+from pathlib import Path
+from typing import Dict, List, Set, Union
 
 import pandas as pd
 from jinja2 import Template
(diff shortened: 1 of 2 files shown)


src/scripts/migrate.py
```diff
@@ -174,13 +176,17 @@ def slurp_docs():
...
-    slurp_path: str = os.path.join(SLURP_DIR, 'ordo.tsv'), subsets_path: str = os.path.join(TEMP_DIR, 'ordo-subsets.tsv')
+    slurp_path: Union[str, Path] = os.path.join(SLURP_DIR, 'ordo.tsv'),
+    subsets_path: Union[str, Path] = os.path.join(TEMP_DIR, 'ordo-subsets.tsv'),
+    onto_config_path: Union[str, Path] = os.path.join(METADATA_DIR, 'ordo.yml'),
 ):
     """Adds rare_disease_subset column to the ORDO migration TSV."""
     # Read inputs
     df: pd.DataFrame = pd.read_csv(slurp_path, sep='\t')
-    df_subsets: pd.DataFrame = get_formatted_subsets_df(subsets_path)
+    df_subsets: pd.DataFrame = get_formatted_subsets_df(subsets_path, onto_config_path)
```
joeflack4, line 189: The fix

src/scripts/migrate.py
joeflack4, line 1: ## Development workflow for this PR
I encountered a bug while running a build for the `develop` --> `main`. The easiest thing to do was fix it as a commit, but this would have caused a mix of different kinds of changes in #556: (i) this bug fix, (ii) build results. Normally we like to keep these things separate, so I created this separate PR. Then I merged it into `main-new` (#554), which is the base for `main-new-build` (#556). After doing that merge, I then rebased `main-new-build`.

PR #539, opened 2024-05-22
Lexmatch `make` updates

joeflack4:
## Overview
    Misc lexmatch updates
    - Silenced some make recipe steps
    - Added a missing .PHONY
    - Added a missing prereq
    - Added a clarifying comment

### Documentation
Was the documentation added/updated under `docs/`?
- [ ] Yes
- [x] No, updates to the docs were not necessary after careful consideration

### QC
Was the full pipeline run before submitting this PR using `sh run.sh make build-mondo-ingest` on this branch (after 
`docker pull obolibrary/odkfull:dev`), and no errors occurred?
- [ ] Yes
- [x] No, there are no functional (code-related) changes to the pipeline in the PR, so no re-run is necessary
   
### New Packages
Were any new *Python packages* added?
- [ ] Yes, [requirements.txt.full](https://github.com/INCATools/ontology-development-kit/blob/master/requirements.txt.full) was updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

Were any other *non-Python packages* added?
- [ ] Yes, [Dockerfile](https://github.com/INCATools/ontology-development-kit/blob/master/Dockerfile) updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

### PR Review and Conversations Resolved
Has the PR been sufficiently reviewed by at least 1 team member of the Mondo Technical team and all threads resolved?
- [ ] Yes

PR #557, opened 2024-06-05
Bug fix: `pip-bioregistry`

joeflack4:
## Overview
We removed a `bioregistry` line (https://github.com/monarch-initiative/mondo-ingest/pull/539#discussion_r1628469187) from the makefile because we thought that this was in ODK now. I'm running a build, and it is not present, so I'm adding it back.


src/ontology/mondo-ingest.Makefile
```diff
@@ -427,6 +427,7 @@ tmp/merged.owl: tmp/mondo.owl mondo-ingest.owl tmp/mondo.sssom.ttl
 $(MAPPINGSDIR)/mondo-sources-all-lexical.sssom.tsv: $(SCRIPTSDIR)/match-mondo-sources-all-lexical.py tmp/merged.db $(MAPPINGSDIR)/rejected-mappings.tsv
 	rm -f $(MAPPINGSDIR)/mondo-sources-all-lexical.sssom.tsv
 	rm -f $(MAPPINGSDIR)/mondo-sources-all-lexical-2.sssom.tsv
+	$(MAKE) pip-bioregistry
```
joeflack4, line 430: The change

src/ontology/mondo-ingest.Makefile
joeflack4, line 1: ## Follow-ups
Next week I'll take a look at ODK, and if necessary, open an issue to add `bioregistry`.

PR #559, opened 2024-06-06
HGNC robot template

joeflack4:
Addresses sub-tasks in:
- https://github.com/monarch-initiative/mondo-ingest/issues/558

Related:
- https://github.com/monarch-initiative/omim/pull/113
- https://github.com/monarch-initiative/mondo/pull/7806

## Overview
Update `mondo_genes.csv` to be a proper ROBOT template, and ties into pipeline for externally managed content.

### Documentation
Was the documentation added/updated under `docs/`?
- [ ] Yes
- [x] No, updates to the docs were not necessary after careful consideration

### QC
Was the full pipeline run before submitting this PR using `sh run.sh make build-mondo-ingest` on this branch (after 
`docker pull obolibrary/odkfull:dev`), and no errors occurred?
- [x] Yes
- [ ] No, there are no functional (code-related) changes to the pipeline in the PR, so no re-run is necessary
   
Build PR:
- #564

### New Packages
Were any new *Python packages* added?
- [ ] Yes, [requirements.txt.full](https://github.com/INCATools/ontology-development-kit/blob/master/requirements.txt.full) was updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

Were any other *non-Python packages* added?
- [ ] Yes, [Dockerfile](https://github.com/INCATools/ontology-development-kit/blob/master/Dockerfile) updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

### PR Review and Conversations Resolved
Has the PR been sufficiently reviewed by at least 1 team member of the Mondo Technical team and all threads resolved?
- [x] Yes

--- 

CC: @souzadevinicius Thought this would be a good one for you to review

PR #564, opened 2024-06-13
HGNC ROBOT template - build

joeflack4:
QC / build PR for: 
- #559


docs/metrics/doid.md
joeflack4, line 1: ## Build done
I don't know if there's much if anything to review there, because I've committed the outputs in `external/` in this #559 already. Just face check nothing broke I guess.

matentzn (approved):
I am assuming the run did correctly generate `src/ontology/external/mondo_genes.robot.owl` which just does not show up because the branch is on top of the feature branch! 

LGTM!

twhetzel:
If the result files are in #https://github.com/monarch-initiative/mondo-ingest/pull/559 then overall I think this Build is ok, but I do have questions on the ROBOT file created in that PR https://github.com/monarch-initiative/mondo-ingest/pull/559#issuecomment-2166586440. Those questions do not effect this feature branch Build review PR though. 

Curious though, why do the "mondo-genes..." files not show up here?

twhetzel (changes requested):
A few comments and there is a design question in https://github.com/monarch-initiative/mondo-ingest/pull/559

docs/metrics/doid.md
joeflack4: Nico:
> I am assuming the run did correctly generate `src/ontology/external/mondo_genes.robot.owl` which just does not show up because the branch is on top of the feature branch!

Yep. If it didn't generate it, there'd be a diff showing that the file was lost.

joeflack4: Trish:
> But how did other files change?

@twhetzel I'm pretty sure this is because the inputs to this pipeline are many (our sources, and the `mondo` repo itself), and change often. In the event that even 1 of these change, it can result in many files changing. In this case, only 54 files changed, which is sorta in line with my expectations; since we just ran a build very recently, I wouldn't expect that many files to change, and 54 is a relatively low diff.

Trish:
> I'm curious how the build is being run and branching strategy.

I've been doing it the same way every time, pretty much:
* Feature branch PRs: I branch them off of `develop`
* Build PRs for feature branches: I branch them off of the feature branch
Also:
* When `main` gets updated: I merge it into `develop`
  * Then I rebase all of my active feature branches from `develop`: I wasn't doing this all of the time before, but now I'm trying to remember to always do that too.

joeflack4:
Closing this PR, as it has built successfully and been approved.

It does say that Trish requested changes, but that is pertaining to:
> A few comments and there is a design question in https://github.com/monarch-initiative/mondo-ingest/pull/559

I looked into that, and it looks like the questions were all addressed, so I'm closing this now.

Issue #544, opened 2024-05-25
Docs: How to review a data build


joeflack4:
## Overview
Add / update a page to include these docs:

1. The data build is not reviewed for specific changes in content, but general patterns of changes. It is recommended to spend 10 minutes reviewing each databuild.
2. There are two important reasons to review data builds: (1) looking out for large unexplainable changes and (2) increasing your familiarity with the data generated by the pipeline. The later is as important as the former: as data stewards in the Mondo Ingest pipeline you should understand all data (every single file!) that is generated insight out, and the best way to do that is to review each file many times until it sticks. It is not wrong to use a data release to ask questions like: "what is the purpose of this file?".
3. Ensure that no files are added or removed. There are few good reasons for files being added or removed and if they happen they should be explained.
4. ORDO, DOID and OMIM matches and migration files should have "reasonable" changes, i.e. be in line with one could expect as a consequence of a few weeks worth of curation (example: 1000 added lines is not a good sign, but 70 removed lines is within reason).
5. Metrics and ontology related files should change within reason (numbers like axiom counts changing in the realms of 250 plus minus are nearly always ok, changes between 250 and 1000 are worth a second look, and changes beyond 1000 merit an investigation).
6. No file should be totally empty in a data sync.

## Additional info
Context: [Original discussion](https://github.com/monarch-initiative/mondo-ingest/pull/542#discussion_r1614748642)

joeflack4:
Also maybe worth adding, but I remember also:
* General face check (I suppose many of the bullets above qualify for that)
* Lexmatch outputs: Check to make sure there are plenty of rows
* Slurp outputs: Check to make sure there are plenty of rows

joeflack4:
We could also make a QC test / script, maybe even make a GH action for it ; but I don’t know if we’re at the point where that’s worth doing.

joeflack4:
@matentzn About this criteria:
> Ensure that no files are added or removed. There are few good reasons for files being added or removed and if they happen they should be explained.

I sometimes see files removed, but they are like lexical mapping `closematch` files, or `broadmatch` or maybe `narrowmatch`. I take it we're not quite so worried about some files being removed?

If there were `exactmatch` files being removed, certain other files, I would be more worried.

matentzn:
> I sometimes see files removed or added, but they are like lexical mapping closematch files, or broadmatch or maybe narrowmatch. I take it we're not quite so worried about some files being removed?


[...]

PR #566, opened 2024-06-13
Update all Mondo Ingest data files (12 June 2024)

matentzn:
## Overview

This PR:
- Updates all data files according to our data ingest pipeline

```
sh run.sh make build-mondo-ingest -B
```

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [X] No

- [X] No

- [ ] Yes


src/ontology/lexmatch/README.md
joeflack4, line 1: ## Pre-merge checklist
- [ ] @matentzn @twhetzel Give thoughts / agreement

Minor nit, but just thinking; I don't think we need this checklist in the OP for data build PRs, and can opt to manually remove it for such PRs.

joeflack4, line 1: ## General review - pass
Just have some minor questions in this comment, but otherwise LGTM!

joeflack4, line 1: ## Minor questions
This is the first time I've utilized this SOP, and I found it very helpful!:
- #544 

I have some follow-ups / questions regarding this criteria, though:
> 4. No file should be totally empty in a data sync.

@matentzn When you say "data sync", do you mean the outputs of pipelines such as `sync-synonyms`, `sync-subclassof`, etc?

If so, then this would be the case for most outputs, but potentially not _all_ outputs? I'm thinking of the `*-added` outputs for `sync-subclassof`. If we integrate these new additions, then when we next run the build, these files should be empty, no?

If by "data sync", you mean more than just these `sync` pipelines, what other pipelines would they be? Does this include lexmatch? Because I am seeing some empty files in `mappings/`: icd10cm, icd10who, and icd10foundation.

- [ ] Based on @matentzn's answers, I may want to update #544 with further clarity.

twhetzel: Sure, but it's easy enough to just check what's applicable and document what's being done. There are a lot more high priority things to spend time on :)

matentzn: I like the structured OP for all PRs, and it takes me less than 20 seconds to fill it in - so even if it is redundant, it is not worth it IMO to create an exception for one PR type over another

matentzn:
@twhetzel can we merge soon?

twhetzel (approved):
Generally looks ok to me. 

Are the additions in `src/ontology/unmapped/omim-unmapped.owl` a concern or expected?

matentzn:
Thanks, looks ok! Reflects the new OMIM terms (additions) and the ones that were mapped (and are there for not anymore "unmapped").

PR #570, opened 2024-06-17
Update regexes in lexmatch rules

matentzn:
With the latest version of OAK (later than ODK 1.5.1) the "Broad match terms with the term 'other' in them." causes a problem (`re.error: global flags not at the start of the expression at position 2`). @cmungall advised that this is not correct yaml syntax (using the "r".

## Overview

This PR:
- Updates the lexmatch rules removing the `r` in front of all match expressions (as per @cmungall advice)
- Updates the `match` clause of the `'s` rule, which was flagged as illegal yaml by my linter

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [X] No

- [X] No

- [ ] Yes

### Additional checklist

- [ ] Needs to be reviewed by @cmungall 
- [ ] Needs to be carefully tested with OAK 0.6.9 (!)

### DIFF
--- a/src/ontology/config/mondo-match-rules.yaml
+++ b/src/ontology/config/mondo-match-rules.yaml
@@ -70,7 +70,7 @@ rules:
 
   - synonymizer:
       description: Remove box brackets bound info from the label.
-      match: r'\[[^)]*\]'
+      match: '\[[^)]*\]'
       match_scope: "*"
       replacement: ""
 
@@ -88,26 +88,26 @@ rules:
   
   - synonymizer:
       description: Broad match terms with the term 'other' in them.
-      match: r'(?i)^Other '
+      match: '(?i)^Other '
       match_scope: "*"
       replacement: ""
       qualifier: "broad"
 
   - synonymizer:
       description: Replace disease with disorder in the label.
-      match: r'disease'
+      match: 'disease'
       match_scope: "*"
       replacement: "disorder"
 
   - synonymizer:
       description: Replace numbers by "type number" in the label.
-      match: r'(?<!type)\s(\d)'
+      match: '(?<!type)\s(\d)'
       match_scope: "*"
       replacement: " type \\1"
 
   - synonymizer:
       description: Replace "'s" by "s" in the label.
-      match: r'\'s'
+      match: "[']s"
       match_scope: "*"
       replacement: "s"
 


src/ontology/config/mondo-match-rules.yaml
```diff
@@ -88,26 +88,26 @@ rules:
...
       description: Replace numbers by "type number" in the label.
-      match: r'(?<!type)\s(\d)'
+      match: '(?<!type)\s(\d)'
       match_scope: "*"
       replacement: " type \\1"
 
   - synonymizer:
       description: Replace "'s" by "s" in the label.
-      match: r'\'s'
+      match: "[']s"
```
joeflack4, line 110: I'm unfamiliar with these rules and these updates. Just going to trust you guys on this one!

joeflack4 (approved):

src/ontology/config/mondo-match-rules.yaml
cmungall: See the comments at the top of the PR from Nico which explains it all. `r` is python syntax, not yaml! https://stackoverflow.com/questions/33729045/what-does-an-r-represent-before-a-string-in-python

This used to work because the code would pass the yaml string through a python `eval`. I hope it's clear why that is wrong!

cmungall (approved):

src/ontology/config/mondo-match-rules.yaml
cmungall: This is each to check:

z.yaml
```yaml
foo: '\[[^)]*\]'
```

```python
>>> import yaml
>>> obj = yaml.safe_load(open("z.yaml"))
>>> obj
{'foo': '\\[[^)]*\\]'}
>>> import re
>>> re.compile(obj["foo"])
re.compile('\\[[^)]*\\]')
>>> pattern=re.compile(obj["foo"])
>>> pattern.match('[x]')
<re.Match object; span=(0, 3), match='[x]'>
>>> pattern.match('[)]')
>>> pattern.match('a')
>>> pattern.match('[]')
<re.Match object; span=(0, 2), match='[]'>
>>> pattern.match('[xx')
>>>
```

so works as presumably expected

hrshdhgd (approved):

src/ontology/config/mondo-match-rules.yaml
joeflack4: Oh, thanks for taking the time! That is indeed what I was more or less what suspecting regarding the regex / `r'` changes.

twhetzel (approved):
LGTM - I changed this to merge into `develop`.

PR #573, opened 2024-06-18
Build, `main` 2024-06-17

joeflack4:
Building off of `main` just after merging `develop`.


docs/reports/migrate.md
joeflack4, line 1: ## General approval
This passes the SOP / my face check.

joeflack4, line 1: ## Pondering - many changes even after just running
I am curious, though. Haven't put thought into this yet. But we just ran this on `develop` (#571). But there were really not many differences between `main` and `develop` when we merged it other than output files.

The one difference I can remember is that there was a single line in a single exclusion config (Trish and I discussed).

@matentzn @twhetzel I plan to follow up on these on Slack.
So what I'm wondering is... if that was the only difference between the branches; such a tiny difference, then why do we have this many files changed after running these builds basically back-to-back (difference of about 4 hours)?

**Possible causes:**
- Sorting: Lack of sorting of outputs causes inconsistencies.
- Maybe the tiny change was sufficient to cause a large diff?
- Maybe there were more differences between main and develop than I realize?

**Other observations in these diffs**
- `docs/metrics/`: I noticed something curious in #571 that I was going to bring up, but I'll mention it here because it's related. in #571 I noticed that all the `metadata/*metrics.json` files changed, but there were no updates to `docs/metrics/`. Seems like a bug to me. However with this PR, there are changes to both sets of files. Strange.
- `docs/reports/mirate*.md`: There is a change to the file for DO and OMIM, but the only corresponding slurp TSV with a change is OMIM. I'd expect these to be 1:1.
- _OMIM/DO_: Sets of files where OMIM and DO are the only sources in diffs: `docs/reports/`, `mappings/` (other than nando), `lexmatch/split-mapping-set/`
- _OMIM/ORDO_: Sets of files where OMIM and DO are the only sources in diffs: `reports/`
- _OMIM_: The only source that had a diff in `slurp/`.

matentzn: Two immediate notes:

1. OMIM basically always changes, so tight turnarounds need to be checked agains the omim release (monarch repo) schedule
2. DO changed 3 weeks ago, and @twhetzel fix of the xrefs could account for the drastic change in axiom counts.

src/mappings/doid.sssom.tsv
matentzn, line 1: 🎉

matentzn (approved):
LGTM

docs/reports/migrate.md
twhetzel: Yes, the DO changes in this release to remedy the previously lost rows from the `doid.sssom.tsv` file are expected. However, generally we would not expect such as large addition.

src/mappings/doid.sssom.tsv
twhetzel: Good to see these restored!

twhetzel (approved):
LGTM

PR #578, opened 2024-06-19
Build - 2024-06-19

joeflack4:
## Overview
Re-running after `mondo` PRs [just merged](https://monarchinitiative.slack.com/archives/G01JACCEDAB/p1718808192145239) (from review of alignment files).

## Sub-tasks
- [x] 1. Fix the issue in `mondo-ingest` (#581)
- [x] 2. Fix the issue in `sssom-py` (make an issue for now) https://github.com/mapping-commons/sssom-py/issues/546
- [ ] 3. @joeflack4 Investigate why this is happening (no `confidence` values)

## Related
- #583 
- #581
- Alternate build: https://github.com/monarch-initiative/mondo-ingest/pull/582

PR #581, opened 2024-06-21
Add confidence fill conditional in lexmatch compare

matentzn:
Currently the filter_redundant_rows in the sssom packages fails when

1. There is a confidence columns and
2. no single value is set

This needs to be fixed in sssom package.

In the meantime, this is a workaround.

Resolves #ISSUE(s).

## Overview

This PR:
- One
- Two
- Three

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [ ] No

- [ ] No

- [ ] Yes

### DIFF
--- a/src/scripts/lexmatch-sssom-compare.py
+++ b/src/scripts/lexmatch-sssom-compare.py
@@ -296,6 +296,14 @@ def get_unmapped_df(
     ]
 
     new_df = pd.concat([unmapped_lex_df, unmapped_mondo_df], axis=0)
+    if 'confidence' not in new_df.columns:
+        new_df['confidence'] = 0.5
+    else:
+        new_df['confidence'] = pd.to_numeric(new_df['confidence'], errors='coerce')  # Convert to numeric, invalid parsing will be NaN
+        new_df['confidence'] = new_df['confidence'].apply(lambda x: x if 0 <= x <= 1 else 0.5)  # Replace out of range values with default
+        new_df['confidence'].fillna(0.5, inplace=True)  # Replace NaN with default value
+
+    new_df.to_csv("check_me_out.sssom.tsv", sep="\t")
     filtered_new_df = filter_redundant_rows(new_df)
     return filtered_new_df
 


src/scripts/lexmatch-sssom-compare.py
```diff
@@ -296,6 +296,14 @@ def get_unmapped_df(
     ]
 
     new_df = pd.concat([unmapped_lex_df, unmapped_mondo_df], axis=0)
+    if 'confidence' not in new_df.columns:
+        new_df['confidence'] = 0.5
```
joeflack4, line 300: ## Default to confidence 0.5
So, confidence ranges 0 to 1, 0.5 being "equally likely to be a match or not". This makes sense as a default, then.

### Questions
1. What are the consequences, if any, to setting this as a default value?
2. When you said earlier you were running a build with `-B`, it was this branch? To see if this bug fix worked?
3. It seems that (i) this fix will be wholly sufficient to address the new [AttributeError](https://github.com/monarch-initiative/mondo-ingest/pull/578#discussion_r1646653777) in `mondo-ingest`. And I know of (ii) the fix needed in `sssom-py`. Also, (iii), you wanted me to investigate why there are no confidence values being assigned. Other than these 3 things, are there any other tasks for us to complete as a result of dealing with this error?

src/scripts/lexmatch-sssom-compare.py
joeflack4, line 1: ## Questions
1. What are the consequences, if any, to setting 0.5 as a default value?
2. When you said earlier you were running a build with `-B`, it was this branch? To see if this bug fix worked?
3. This is only in _draft_ because the build still needs to be run successfully, yes?
4. It seems that (i) this fix will be wholly sufficient to address the new [AttributeError](https://github.com/monarch-initiative/mondo-ingest/pull/578#discussion_r1646653777) in `mondo-ingest`. And I know of (ii) the fix needed in `sssom-py`. Also, (iii), you wanted me to investigate why there are no confidence values being assigned. Other than these 3 things, are there any other tasks for us to complete as a result of dealing with this error?

joeflack4 (approved):

src/scripts/lexmatch-sssom-compare.py
matentzn: Built failed: #582 with different error :/

> When you said earlier you were running a build with -B, it was this branch? To see if this bug fix worked?

Yes

> What are the consequences, if any, to setting 0.5 as a default value?

None that I can think off

> This is only in draft because the build still needs to be run successfully, yes?

Yes. Handing it over to @twhetzel now as its getting to late for me to continue

> . Other than these 3 things, are there any other tasks for us to complete as a result of dealing with this error?

Figuring out the new error (OP of data release #582)

joeflack4, line 1: Note, this is a code change and should normally go into `develop`. See more: 
- https://github.com/monarch-initiative/mondo-ingest/pull/582#discussion_r1649328825

twhetzel: this is a hotfix for `main`

src/scripts/lexmatch-sssom-compare.py
hrshdhgd, line 300: Not needed. The[ PR in `sssom-py`](https://github.com/mapping-commons/sssom-py/pull/548) should eliminate the need for this.

hrshdhgd (changes requested):
Not needed. The[ PR in `sssom-py`](https://github.com/mapping-commons/sssom-py/pull/548) should eliminate the need for this.

joeflack4:
> Not needed. The PR in sssom-py https://github.com/mapping-commons/sssom-py/pull/548 should eliminate the need for this.

That is good to know!
I'm actually surprised though that this PR wasn't automatically closed, as #582 includes it and has already been merged. I double checked to make sure that this commit was included there and it is.

In any case gonna close this now.

Issue #399, opened 2024-01-24
Remove GARD


joeflack4:
## Overview
This is causing errors when Joe runs `build-mondo-ingest`. There are several places ot remove.

Joe has many more notes in Workflowy.

matentzn:
Did you use my branch? There were quite a few run errors like GARD, also do. I am systematically eliminating them atm, most of them are rooted in the SSSOM py migration to curies converters. In any case, don't worry now, I will get this to work.

joeflack4:
> Did you use my branch?
No. I was checking on the results of `build-mondo-ingest` on my branch on Tuesday. 

> don't worry now, I will get this to work.
I won't, but I feel bad man, that you have to be the one to do this, especially what with how you're trying to allocate your time now!

Trish and I chatted briefly again on how to run `build-mondo-ingest` in a modular way that can pick up where it left off. I don't know when we can prioritize that, but it would be nice.

joeflack4:
@matentzn Just want to confirm. Sorry I haven't gotten to this yet; I'm still good to remove this, yes?

matentzn:
Except what is in externally managed content! From the regular pipeline, yea

PR #600, opened 2024-07-01
Exact synonym updates

joeflack4:
## Overview
- remaps `IAO:0000118` (alt label) to `oboInOwl:exactSynonym`
- adds `oboInOwl:exactSynonym` for all `rdfs:label` on components

### Documentation
- [ ] Yes

### QC
- [x] Yes
   
Build:
(I think #601 is an older run that did not finish. #609 is really what we should be looking at. I don't know why it didn't get pushed. I might have just accidentally left out `--force`).
- #601 
- #609

### New Packages
- [x] No

- [x] No

- [x] Yes

### DIFF
--- a/src/ontology/config/property-map.sssom.tsv
+++ b/src/ontology/config/property-map.sssom.tsv
@@ -6,3 +6,4 @@ http://www.ebi.ac.uk/efo/definition	http://purl.obolibrary.org/obo/IAO_0000115
 http://www.ebi.ac.uk/efo/reason_for_obsolescence	rdfs:comment
 http://www.ebi.ac.uk/efo/alternative_term	oboInOwl:hasExactSynonym
 http://purl.obolibrary.org/obo/ECO_0000218	oboInOwl:source
+http://purl.obolibrary.org/obo/IAO_0000118	oboInOwl:hasExactSynonym
--- a/src/sparql/exact_syn_from_label.ru
+++ b/src/sparql/exact_syn_from_label.ru
@@ -1,0 +1,11 @@
+# Create an exact synonym copy for all labels
+PREFIX oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
+PREFIX owl: <http://www.w3.org/2002/07/owl#>
+PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+
+INSERT {
+  ?cls oboInOwl:hasExactSynonym ?label .
+} WHERE {
+  ?cls rdfs:label ?label .
+  FILTER NOT EXISTS { ?cls owl:deprecated ?deprecated . }
+}
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -70,6 +70,7 @@ $(COMPONENTSDIR)/omim.owl: $(TMPDIR)/omim_relevant_signature.txt | component-dow
 			--update ../sparql/fix_omimps.ru \
 			--update ../sparql/fix-labels-with-brackets.ru \
 			--update ../sparql/fix_illegal_punning_omim.ru \
+			--update ../sparql/exact_syn_from_label.ru \
 		annotate --ontology-iri $(URIBASE)/mondo/sources/omim.owl --version-iri $(URIBASE)/mondo/sources/$(TODAY)/omim.owl -o $@; fi
 
 # todo: See #1 at top of file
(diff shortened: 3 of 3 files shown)


src/ontology/config/property-map.sssom.tsv
```diff
@@ -6,3 +6,4 @@ http://www.ebi.ac.uk/efo/definition	http://purl.obolibrary.org/obo/IAO_0000115
 http://www.ebi.ac.uk/efo/reason_for_obsolescence	rdfs:comment
 http://www.ebi.ac.uk/efo/alternative_term	oboInOwl:hasExactSynonym
 http://purl.obolibrary.org/obo/ECO_0000218	oboInOwl:source
+http://purl.obolibrary.org/obo/IAO_0000118	oboInOwl:hasExactSynonym
```
joeflack4, line 9: ## Remap 'alt label' to 'exact synonym'

src/sparql/exact_syn_from_label.ru
```diff
@@ -0,0 +1,11 @@
+# Create an exact synonym copy for all labels
```
joeflack4, line 1: ## Create an exact synonym copy for all labels

src/sparql/exact_syn_from_label.ru
joeflack4, line 1: ## QC build
Going to run build and also make sure the query works as expected.

src/sparql/exact_syn_from_label.ru
joeflack4: - [ ] @matentzn and/or @twhetzel Is this really the best way to do this?

My builds are taking a very long time and not yet completing. It's seemed to freeze twice. I think this is because the laptop accidentally went to sleep both times.

But it made me think of something: Maybe this change has unintended consequences.

It would be very easy for me to handle this `label` --> `exactSynonym` thing in the subclass sync Python script, without needing to modify these components.
I am wondering if adding all these extra synonyms is resulting in any of: (i) potentially causing lexmatch to take longer, (ii) might cause other lexmatch issues, (iii) might cause other unforeseen issues.

When the build (#601) is ready, let's look at the outputs and see about (ii) and (iii).
However if the build is really taking longer because of this, I recommend we do this in Python instead.

twhetzel: Given this comment and the Slack comment, I will also try to run a build from this feature branch.

src/ontology/profile.txt
matentzn, line 1: I deactived all QC for MOndo ingest, since we are not acting on it anyways. See if that removes the built bottleneck

joeflack4: I created a PR that mirrors this commit:
- #605 

It addresses:
- #602 

Has been tested successfully in:
- #601

src/sparql/exact_syn_from_label.ru
joeflack4: Scratch my comment earlier about "changing the SPARQL to a Python implementation", as [Nico said](https://monarchinitiative.slack.com/archives/C03AW548J8M/p1720094791335579?thread_ts=1719948621.736089&cid=C03AW548J8M):
> ...the general pattern (taking the label and asserting as a synonym) is used for lexical matching...

Basically if this not only works for synonym sync, but not only doesn't hurt, but actually helps with lexmatch, we should keep this implementation.

src/sparql/exact_syn_from_label.ru
```diff
@@ -0,0 +1,11 @@
...
+prefix oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
+PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+
+INSERT {
+  ?cls oboInOwl:hasExactSynonym ?label .
+}
+WHERE 
+{
+  ?cls rdfs:label ?label .
+}
```
twhetzel, line 11: I would exclude obsolete terms from having their label added as a synonym.

twhetzel (changes requested):

twhetzel:
Does [GARD](https://github.com/monarch-initiative/mondo-ingest/pull/600/files#diff-36e739e129da51616de07dd1ff7ab0c12aa9b5dd45b7f110c9a025fad909ecc4R172) also need to have the `rdfs:label` --> to `oboInOwl:exactSynonym` change?

src/ontology/config/property-map.sssom.tsv
joeflack4, line 1: Trish wrote:
> Does [GARD](https://github.com/monarch-initiative/mondo-ingest/pull/600/files#diff-36e739e129da51616de07dd1ff7ab0c12aa9b5dd45b7f110c9a025fad909ecc4R172) also need to have the `rdfs:label` --> to `oboInOwl:exactSynonym` change?

Nope, it just needs to be removed:
- #399

src/sparql/exact_syn_from_label.ru
joeflack4: Hmm, alright. That might put this on pause till later this week but I can do that.

joeflack4: Hmm, alright. That might put this on pause till later this week but I can do that.

joeflack4: @twhetzel The change is now done. I ran the query afterwards and checked an example case, and the synonym was no longer added for the deprecated term I looked at.

Otherwise this and the build have been approved; just let me know if it's still mergable after you take a look.

twhetzel: Thanks for the update.

twhetzel (approved):
lgtm

PR #594, opened 2024-06-28
Refactor git refresh system (small refactor)

matentzn:
This is change the Mondo build _inside_ the Mondo ingest pipeline into a method and folds mondo.owl and mondo.sssom.tsv dependency back into the normal Make dependency chain.

## Overview

This PR:
- Creates a make function to run the mondo build
- Adds the file depndencies generated by that build back into the normal Mondo dependency chain

### Documentation
Was the documentation added/updated under `docs/`?
- [ ] Yes
- [x] No, updates to the docs were not necessary after careful consideration

### QC
Was the full pipeline run before submitting this PR using `sh run.sh make build-mondo-ingest` on this branch (after 
`docker pull obolibrary/odkfull:dev`), and no errors occurred?
- [x] Yes
- [ ] No, there are no functional (code-related) changes to the pipeline in the PR, so no re-run is necessary
   
### New Packages
Were any new *Python packages* added?
- [ ] Yes, [requirements.txt.full](https://github.com/INCATools/ontology-development-kit/blob/master/requirements.txt.full) was updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

Were any other *non-Python packages* added?
- [ ] Yes, [Dockerfile](https://github.com/INCATools/ontology-development-kit/blob/master/Dockerfile) updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

### PR Review and Conversations Resolved
Has the PR been sufficiently reviewed by at least 1 team member of the Mondo Technical team and all threads resolved?
- [ ] Yes

PR #603, opened 2024-07-03
Fix ODK version to 1.5.2

matentzn:
To reduce suprises in the future, I suggest we move to a system where the have to deliberately change the ODK version we work with.

## Overview

This PR:
- Fixes the version of ODK used for everyone running the pipeline to 1.5.2

### Documentation
- [ ] Yes

### QC
- [X] Yes
   
### New Packages
- [X] No

- [x] No

- [ ] Yes


src/scripts/lexmatch-sssom-compare.py
```diff
@@ -303,7 +303,6 @@ def get_unmapped_df(
         new_df['confidence'] = new_df['confidence'].apply(lambda x: x if 0 <= x <= 1 else 0.5)  # Replace out of range values with default
         new_df['confidence'].fillna(0.5, inplace=True)  # Replace NaN with default value
 
-    new_df.to_csv("check_me_out.sssom.tsv", sep="\t")
```
joeflack4, line 306: Thought we got rid of this

src/ontology/run.sh.conf
```diff
@@ -1,3 +1,3 @@
 ODK_IMAGE=odkfull
-ODK_TAG=dev
+ODK_TAG=v1.5.1
```
joeflack4, line 2: Good idea

joeflack4 (approved):

src/ontology/run.sh.conf
joeflack4, line 2: Is this supposed to still be in draft?

I thought maybe this was because a build hasn't been run yet, but the "QC" checkbox is checked, indicating that it has already run. But I don't see any PRs for such a build?

matentzn: I had a failure that is oak related on another pipeline so we should make sure that 151 works first

joeflack4: Actually, it looks like you already ran it?
This says it's for #594, but it uses 1.5.1:
- #604

twhetzel: @matentzn who is "we" here :) Are you running a feature branch build for this change?

twhetzel: @matentzn I think it would good to pin this now and we can change again when 1.5.2 is released.

src/scripts/lexmatch-sssom-compare.py
twhetzel: Yes, it was removed from both `main` and `develop` https://github.com/monarch-initiative/mondo-ingest/commit/259725ba558eca8bbce4f392b4b01b98dae3bb74

src/ontology/run.sh.conf
matentzn: > @matentzn who is "we" here :) Are you running a feature branch build for this change?

Sorry I cant at the moment - I can do it next week, or one of you does it if you have cycles free. My laptop is running other builds at the moment relatively non-stop. I can do it Monday

src/scripts/lexmatch-sssom-compare.py
matentzn: I was probably not synced. Now I am!

src/ontology/run.sh.conf
joeflack4: @twhetzel Leaving this up to you if you think we should run the build for this, and who should do that if so.

twhetzel:
NOTE: This should be changed to 1.5.2 next week

src/ontology/run.sh.conf
matentzn:
```suggestion
ODK_TAG=v1.5.2
```

src/ontology/run.sh.conf
joeflack4, line 1: ## When/how to get this into `main`
@twhetzel Not a big deal but just wondering if we are gonna:
a. Make a separate PR off of `main` w/ this same change and have it merge into `main` directly.
b. Merge this into `develop`, and then merge `develop` into `main` at some point in the future.
c. Merge into `develop` and then do a `develop`-->`main` merge process immediately after.

I'm fine w/ anything. I suppose I lean towards 'b'.

joeflack4 (approved):

src/ontology/run.sh.conf
twhetzel: We'll sort out the details on Monday (29-Jul).

twhetzel:
The test build for this change is at: https://github.com/monarch-initiative/mondo-ingest/pull/620

The test build ran without error, but took 4:59:04 to complete, which is ~1 hr longer than previous builds. Other details in the [Mondo Ingest Build Stats](https://docs.google.com/document/d/1MBM0lTAWfLBTQ6Kz9g0yOSvthGbuTQRVdRc0m4shugo/edit) file.

matentzn:
> The test build ran without error, but took 4:59:04 to complete, which is ~1 hr longer than previous builds. Other details in the [Mondo Ingest Build Stats](https://docs.google.com/document/d/1MBM0lTAWfLBTQ6Kz9g0yOSvthGbuTQRVdRc0m4shugo/edit) file.

Nothing in my mind would explain this, as you were using `dev` previously, which more or less corresponds to the 1.5.2 release.. Sorry I am not any help!

PR #606, opened 2024-07-04
Build 2024-07-03 main

twhetzel:
Resolves #ISSUE(s).

## Overview

Build data from `main` for next week's Alignment Build Week for the Mondo release cycle.
(built as: sh run.sh make build-mondo-ingest -B > output_2024-07-03.log 2>&1)

### Documentation
Was the documentation added/updated under `docs/`?
- [ ] Yes
- [x] No, updates to the docs were not necessary after careful consideration

### QC
Was the full pipeline run before submitting this PR using `sh run.sh make build-mondo-ingest` on this branch (after 
`docker pull obolibrary/odkfull:dev`), and no errors occurred?
- [x] Yes
- [ ] No, there are no functional (code-related) changes to the pipeline in the PR, so no re-run is necessary
   
### New Packages
Were any new *Python packages* added?
- [ ] Yes, [requirements.txt.full](https://github.com/INCATools/ontology-development-kit/blob/master/requirements.txt.full) was updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

Were any other *non-Python packages* added?
- [ ] Yes, [Dockerfile](https://github.com/INCATools/ontology-development-kit/blob/master/Dockerfile) updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

### PR Review and Conversations Resolved
Has the PR been sufficiently reviewed by at least 1 team member of the Mondo Technical team and all threads resolved?
- [ ] Yes

PR #607, opened 2024-07-05
Set the source of subclass sync to a merged version of edit and owl

matentzn:
I had a dream last night that this might not be necessary because of the way that semsql works. I dont know exactly what semsql does though, and now that Chris is gone it may be not possible to find out.

Here my dream:

1. semsql is using relationgraph under the hood (that is a fact)
2. RG may, all by itself, remove the red subclass axiom in the figure below, rendering the change in this PR ineffectual

image

In my view, nothing bad will come of this change, but we need to check manually, later down the line, if subclass assertions in mondo-edit.owl do not get support, even though it exists. We can stem that in a second round I think

## Overview
This PR:
- Changes the source of the subclass sync Mondo DB to a version that contains both the release and edit file axioms
- This was discussed and agreed in the last tech call

### Documentation
- [ ] Yes

### QC
- [X] Yes
   
### New Packages
- [X] No

- [X] No

- [ ] Yes
- [X] No


src/ontology/mondo-ingest.Makefile
```diff
@@ -371,6 +371,12 @@ $(TMPDIR)/mondo.sssom.tsv: $(TMPDIR)/mondo_repo_built
 $(TMPDIR)/mondo.owl: $(TMPDIR)/mondo_repo_built
 	cp $(TMPDIR)/mondo/src/ontology/mondo.owl $@
 
+$(TMPDIR)/mondo-edit.owl: $(TMPDIR)/mondo_repo_built
+	$(ROBOT) convert -i $(TMPDIR)/mondo/src/ontology/mondo-edit.obo -o $@
+
+$(TMPDIR)/mondo-edit-mondo-owl-merged.owl: $(TMPDIR)/mondo-edit.owl $(TMPDIR)/mondo.owl
+	$(ROBOT) merge -i $(TMPDIR)/mondo-edit.owl -i $(TMPDIR)/mondo.owl convert -o $@
```
joeflack4, line 378: Ah I see, you merged them, not pass both as inputs. That makes sense.

src/ontology/mondo-ingest.Makefile
joeflack4, line 1: > I had a dream last night... 

Haha

> RG may, all by itself, remove the red subclass axiom in the figure below, rendering the change in this PR ineffectual

Mmm I see.

> In my view, nothing bad will come of this change

Only negative is extra build time. Not a huge problem, but more a problem now than before, given how frequently we run them.

> but we need to check manually, later down the line

Hmm, we can't check now?
Wouldn't we expect to see an increased number of new subclass relations in the `-added` template, with this change?

- [ ] GH issue: If we can't check it now for whatever reason, we should make an issue to check on this later; that redundant ancestor subclass relations/axioms are being added.

joeflack4 (approved):

src/ontology/mondo-ingest.Makefile
matentzn: > Only negative is extra build time

I think the extra build time is not so bad as the only redundant step is the robot merge step, which is fast (less than 8 seconds).

> Hmm, I checked doid.subclass.added.robot.tsv in the build for this PR (https://github.com/monarch-initiative/mondo-ingest/pull/608), and compared it to another recent build (https://github.com/monarch-initiative/mondo-ingest/pull/606), and it had the same number of added subclass relations.

Thank you for checking that. In this case, lets just hold of on this PR for now and go ahead with what we previously had. Relieves the pressure a bit. I will

1. Set this PR to draft
2. Update the OP with the result of your investigation
3. Schedule the PR for total closing in 2 months if we determine it is actually not needed.

src/ontology/mondo-ingest.Makefile
joeflack4: @matentzn Coming back to this. Trying to remember fundamentally why we ought to merge these.

I know that `mondo.owl` ends up removing redundant subclass axioms, and we want to access `mondo-edit` so that we have access to those.

But what does `mondo.owl` add that we need to use for the subclass sync which `mondo-edit` does not already have?

src/ontology/mondo-ingest.Makefile
joeflack4: > the only redundant step is the robot merge step

Also the building of the `.db`, which I just checked; only adds 2.5 minutes.

src/ontology/mondo-ingest.Makefile
matentzn: > But what does mondo.owl add that we need to use for the subclass sync which mondo-edit does not already have?

Inferred subclass relations based on equivalence class axioms for starters!

joeflack4: right, dumb qustion

src/ontology/mondo-ingest.Makefile
joeflack4: ## What this PR can / should achieve, if anything
This text was in the OP but I moved it here:

> Ensure this PR is actually useful & doing what it is supposed to
> As @joeflack4 determine in [his comment](https://github.com/monarch-initiative/mondo-ingest/pull/607#discussion_r1667046860) this PR may, after all, be redundant. We need to figure out the following:

> 1. If a redundant subclass axiom exists in Mondo (red line), will we gather evidence for it?
> 2. We may have to adjust the code to handle this case if it is not, but we can put this on the list of priorities in the next quarter

---

**TLDR**: So far I think that this PR is redundant, and we should work on incorporating the case "SCR is direct in source, but indirect Mondo", and any other [cases](https://docs.google.com/document/d/1H8fJKiKD-L1tfS-2LJu8t0_2YXJ1PQJ_7Zkoj7Vf7xA/edit?tab=t.0) that involve indirect Mondo SCRs, by adding additional Python code rather than changing the `mondo.owl` dependency to `mondo-edit-mondo-owl-merged.owl`.

Consider this example "red line" (indirect) case from #708: `('MONDO:0000248', 'rdfs:subClassOf', 'MONDO:0005502')`

Before this PR, we were using `mondo.db`, and the subclass sync did detect this edge/relation, captured within the variable `rels_indirect_mondo_mondo`.

Using this PR, which uses `mondo-edit-mondo-owl-merged.db`, it detects this edge/relation, but it is now captured within `rels_direct_mondo_mondo`.

Additionally, these "SCR is direct in source, but indirect Mondo" cases are now appearing in `-confirmed`.

I don't like this, though, because now there will be lots of Python variables & comments that will be incorrectly named "direct" because they now include indirect relations. And this will make it impossible for us to tease them apart, and possibly cause bugs that I haven't yet thought of. We should instead address "SCR is direct in source, but indirect Mondo" by adding some more lines of Python. I might be able to get a PR for this this week, but due to the holidays it would likely come on Sunday.

As for this old comment about `-added` ("SCR is direct in the source, but not at all in Mondo"):
> Edit: Hmm, I checked `doid.subclass.added.robot.tsv` in the build for this PR (https://github.com/monarch-initiative/mondo-ingest/pull/608), and compared it to another recent build (https://github.com/monarch-initiative/mondo-ingest/pull/606), and it had the same number of added subclass relations.

IDK why I was surprised by this then. Before this PR, these indirect relations were present. This PR did not change that, so it would not have an effect on `-added`.

matentzn: I agree this PR does not work, hence we didn't merge it; however, my point of confusion is that #708 seems is _not_ a case where the relation is indirect in mondo _edit_, which makes the case here a bit different, as mondo edit and mondo.owl have overlapping but distinct conceptions of which subclass relation is redundant. So the intuition behind this PR still stands (although admittedly, the solution is wrong): we want to mondo.owl PLUS all redundant subclass of axioms as inputs to the confirmation pipeline. Anyways, lets talk Monday, just dumping thought here while still fresh in my mind.

joeflack4: OK, I'll add to the agenda.

I'm tempted to ask more questions now but I'll add to the agenda notes.

joeflack4: Edit: We discussed on 12/02 and determined that we would not use this approach and instead go with the approach in #714.

@matentzn Shall we close this PR?

matentzn: Yes thanks

PR #601, opened 2024-07-01
Exact synonym updates - Build 1

joeflack4:
Build for:
- #600

PR #609, opened 2024-07-08
Exact synonym updates - Build 2

joeflack4:
Build for:
- #600


src/ontology/lexmatch/README.md
joeflack4, line 1: ## FYI: this is the newest build for #600
Since this build initially froze, I had to run it several times. It looks like this PR (#609) is the final build that ran. I just pushed it now after noticing that my local files in the clone where I ran the build were different than what was in #601 (build 1). Note that there are 21 more files changed in this PR.

I probably just forgot to `--force` push when finishing that final build. Sorry.

src/mappings/doid.sssom.tsv
```diff
@@ -275,13 +273,10 @@ DOID:0050427	xeroderma pigmentosum	oboInOwl:hasDbXref	MESH:D014983	semapv:Unspec
...
-DOID:0050428	nonepidermolytic palmoplantar keratoderma	oboInOwl:hasDbXref	NCI:C3147	semapv:UnspecifiedMatching
-DOID:0050428	nonepidermolytic palmoplantar keratoderma	oboInOwl:hasDbXref	OMIM:600962	semapv:UnspecifiedMatching
-DOID:0050428	nonepidermolytic palmoplantar keratoderma	oboInOwl:hasDbXref	ORDO:2337	semapv:UnspecifiedMatching
-DOID:0050428	nonepidermolytic palmoplantar keratoderma	oboInOwl:hasDbXref	UMLS:C0022584	semapv:UnspecifiedMatching
-DOID:0050428	nonepidermolytic palmoplantar keratoderma	oboInOwl:hasDbXref	UMLS:C3489771	semapv:UnspecifiedMatching
-DOID:0050428	nonepidermolytic palmoplantar keratoderma	skos:exactMatch	OMIM:600962	semapv:UnspecifiedMatching
+DOID:0050428	nonepidermolytic palmoplantar keratoderma	oboInOwl:hasDbXref	MESH:C563422	semapv:UnspecifiedMatching
+DOID:0050428	nonepidermolytic palmoplantar keratoderma	oboInOwl:hasDbXref	UMLS:C1833030	semapv:UnspecifiedMatching
+DOID:0050428	nonepidermolytic palmoplantar keratoderma	skos:exactMatch	MESH:C563422	semapv:UnspecifiedMatching
+DOID:0050428	nonepidermolytic palmoplantar keratoderma	skos:exactMatch	UMLS:C1833030	semapv:UnspecifiedMatching
```
twhetzel, line 279: What does `skos:exactMatch` mean in the context of this sssom file and why are there now `oboInOwl:hasDbXref` and `skos:exactMatch`?

twhetzel (approved):
Overall, looks fine. I did add a comment in #600 about updating the SPARQL query to exclude deprecated terms from the label-->synonym update.

src/mappings/doid.sssom.tsv
joeflack4: Hmm, that's a good question. I'll have to look into that. I believe these files are from the lexmatch pipeline that Harshad wrote. I don't know if it is a proposed match, or a match that is actually declared in the source.

I'm not sure why`hasDbXref` is in here either, but I looked and my development environment also has these, so this predates this PR.

matentzn: When you have questions / hypotheses like this its good to consult the `Makefile` first:

https://github.com/monarch-initiative/mondo-ingest/blob/52ec97cfe3d64c5aa49f49393fe2735c2a45ca78/src/ontology/mondo-ingest.Makefile#L187

These `src/mapping/` files have nothing to do with `lexmatch`, they are just 1:1 extracts from the components (in this case DOID). So whatever DO says, will appear in this file.

PR #616, opened 2024-07-24
OMIM: Annotated 'included' synonyms using a different property

joeflack4:
Related:
- https://github.com/monarch-initiative/omim/issues/116
- https://github.com/monarch-initiative/omim/issues/117
- Build PR: #617

## Overview
Update: `properties.txt` to list this new annotation property being used by OMIM to denote these special synonyms. This ensures that these properties are not removed during the component OWL creation process.

### Documentation
Was the documentation added/updated under `docs/`?
- [ ] Yes
- [x] No, updates to the docs were not necessary after careful consideration

### QC
Was the full pipeline run before submitting this PR using `sh run.sh make build-mondo-ingest` on this branch (after 
`docker pull obolibrary/odkfull:dev`), and no errors occurred?
- [x] Yes
- [ ] No, there are no functional (code-related) changes to the pipeline in the PR, so no re-run is necessary
   

Build PR:
- #617 

### New Packages
Were any new *Python packages* added?
- [ ] Yes, [requirements.txt.full](https://github.com/INCATools/ontology-development-kit/blob/master/requirements.txt.full) was updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

Were any other *non-Python packages* added?
- [ ] Yes, [Dockerfile](https://github.com/INCATools/ontology-development-kit/blob/master/Dockerfile) updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

### PR Review and Conversations Resolved
Has the PR been sufficiently reviewed by at least 1 team member of the Mondo Technical team and all threads resolved?
- [x] Yes

PR #617, opened 2024-07-24
DO NOT MERGE - OMIM: Annotated 'included' synonyms using a different property - build

joeflack4:
> [!warning]
> Build PR; don't merge.

## Overview
Build for:
- #616

### Documentation
- [ ] Yes

### QC
- [x] Yes
   
### New Packages
- [x] No

- [x] No

- [x] Yes


src/ontology/imports/omo_import.owl
joeflack4, line 1: ## Build running
I'll notify when it is complete.

src/ontology/lexmatch/README.md
Overall, LGTM.

I was expecting that possibly that there would be (more) lines removed in the OMIM lexmatch files as there should now be less synonyms to match on, not as a result of this PR, but as a result of the latest OMIM release having removed those 'included' synonyms out in favor of this new annotation property. However I'm not seeing such deletions / negative deltas.

Otherwise, I wouldn't expect any side effects / changes based on #616 to show up anywhere in this build or any others, unless we write some new logic to process this new annotation property.

Also thanks @Trish for putting in draft mode; I forgot.

twhetzel (approved):
LGTM

PR #620, opened 2024-07-25
DO NOT MERGE - Test Build from branch `fix-odk-version`

twhetzel:
Resolves #ISSUE(s).

## Overview

Feature branch build for: https://github.com/monarch-initiative/mondo-ingest/pull/603

### Documentation
- [ ] Yes

### QC
- [x] Yes
   
### New Packages
- [x] No

- [ ] No

- [ ] Yes


src/ontology/lexmatch/README.md
Looks overall fine.

I do notice a `broadmatch` file removed. Per the SOP, deleted or added files is something to look out for but maybe that should be updated to worry less about certain kinds of files, like these close/broad/narrowmatch files. I'm guessing this file being deleted is not an issue. I do see 2 new closemach files and a new broadmatch file as well.

@twhetzel Can you change the base to the branch for #603 instead of `develop`?
The number of files in the diff, as well as the diff sizes, are a bit bigger than I expected since I got 90 files changed in the build I ran yesterday, and I think the idffs were smaller. Maybe changing the base will alleviate.
It might also alleviate the conflicts, which I am surprised to see.

joeflack4 (approved):

src/ontology/lexmatch/README.md
twhetzel: Good points. I am not concerned about `src/ontology/lexmatch/split-mapping-set/mondo_broadmatch_doid.tsv` being removed since this is a subset file of all DOID matches, there are other DOID lexmatch files, and the entities removed in this removed file do exist in Mondo as xrefs.

I did change the base branch to the feature branch this build is from. Still some larger diffs, but no longer conflicts as a sanity check.

@joeflack4 since you approved this, I will merge #603.

PR #622, opened 2024-07-31
ODK 1.5.3 & Replace jinja2 with Jinjanator

matentzn:
In a future release of ODK, we will have to replace the unmaintained j2 with a new tool we are deploying in ODK. This PR removes all dependencies on j2 with Jinjanator, which reflects that change.

For more context of that see this ODK PR, which switches the BASE system in ODK to a new Ubuntu, and upgraded python to version 12 https://github.com/INCATools/ontology-development-kit/pull/1076

## Overview

This PR:
- Replaces the use of the j2 package for instantiating jinja templates with jinjanator (https://github.com/kpfleming/jinjanator)

### Documentation
- [ ] Yes

### QC
- [X] Yes

Build:
- #683   

### New Packages
- [X] No (This new package is added to ODK, not to Mondo Ingest)

- [X] No

- [x] Yes


src/ontology/mondo-ingest.Makefile
```diff
@@ -18,7 +18,7 @@ SKIP_HUGE=false
 ####################################
 .PHONY: pip-%
 pip-%:
-	 python3 -m pip install --upgrade $*
+	 python3 -m pip install --upgrade $* --break-system-packages
```
joeflack4, line 21: ## `--break-system-packages`
> The `--break-system-packages` flag in pip is a potentially dangerous option that allows pip to modify or replace packages that are part of your system's Python installation.

What is the rationale for using this here? Is this a temporary change?

If it is a permanent change, does it need to be bundled in this PR or can it be added separately?

src/ontology/mondo-ingest.Makefile
```diff
@@ -318,14 +318,10 @@ metadata/%-metrics.json: $(COMPONENTSDIR)/%.owl
 .PRECIOUS: metadata/%-metrics.json
 
 ../../docs/metrics/%.md: metadata/%-metrics.json | ../../docs/metrics/
-	j2 "$(SOURCE_METRICS_TEMPLATE)" metadata/$*-metrics.json > $@
-
-.PHONY: j2
-j2:
-	pip install j2cli j2cli[yaml]
```
joeflack4, line 325: ## Rationale for this change
So I guess that the `j2` Jinja CLI was never part of ODK, but `jinjanate` will be. So that is the reason we are going to start using `jinjanate`?
I assume it doesn't otherwise offer any special benefits?

src/ontology/mondo-ingest.Makefile
matentzn: In the future this should not be necessary anymore, as we should rarely if ever need to do pip install. Unfortunately, this is a sideeffect of migrating to python 3.12.. its just needed now to add this option if you are installing into the global (rather than a virtual) environment. the next odk version (already on `dev`) will be on python 3.12 and need this.

src/ontology/mondo-ingest.Makefile
matentzn: j2 was actually part of ODK for a while now (not when this was added to the makefile), but jinjanate is now part of the base packages. Already installed on `dev`.

src/ontology/mondo-ingest.Makefile
joeflack4, line 1: @matentzn Looks like it's time to revive this!

matentzn: Agreed!

matentzn: Once ODK 1.5.3 migration is complete, you can merge this as well!

matentzn: Actually, probably you should do the ODK 1.5.3 migration in this very PR as all the changes here are requirements for the 1.5.3 upgrade!

joeflack4: @matentzn I don't mind, but what would you have me do?
- [ ] @joeflack4 Fix merge conflicts
- [ ] Do we need to add `jinjanator` to the `requirements.txt`s?
- Anything else remaining to be done?

matentzn: No requirements txt necessary- it's all in ODK. Search for the string 1.5.2 in the repo and replace by 1.5.3! At least run.sh.conf

joeflack4: @matentzn OK. That was originally what I was going to check but I forgot. I thought we actually had `mondo-ingest` set to `latest` and not a fixed version, but I do see that on 7/25 you set fixed it to `ODK_TAG=v1.5.2`. I'll update that.

twhetzel: I thought I already had a separate ticket to update `mondo-ingest` to 1.5.3. I am missing why these two updates need to happen together.

matentzn: This PR contains changes that are needed to run OdK 1.5.3 - you will run into errors if you don't use these. For example odk no longer has j2 which is needed by the pipeline, it uses jinjanator now.

joeflack4: @matentzn Alright I fixed merge conflicts, updated the version, and ran `update_repo`. You may want to review~

matentzn (commented):
Approved!

joeflack4 (approved):

joeflack4:
@matentzn You had this on `main` but I'm merging now and feel putting this to `develop` would be better right now.

CC @twhetzel

PR #630, opened 2024-08-07
Benign attempt to resolve strange `LF` --> `CRLF` issue on Joe's machine

joeflack4:
## Overview
    Add .gitattributes to .gitignore
    This allows individual contributors to add a .gitattributes for their needs, but opts for individualized .gitattributes per contributor.
    Use case: One contributor is having an issue where, despite global git configuration, a lot of the line endings for files are getting converted to CRLF (Windows) rather than LF (Unix). The .gitattributes solution attempts to solve this, but since others aren't having this issue, it may be better to not bloat the repo by adding a file that it doesn't generally need.

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [x] No

- [x] No

- [ ] Yes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ semantic.cache/
 .vscode/*
 
 # - static
+.gitattributes
 .DS_Store
 
 # Specific files, patterns, and directories


joeflack4:
Will bring this up at next 1:1 meeting.

twhetzel:
@joeflack4 do you have a follow-up commit somewhere to show the before and after effect of using this to specify the end of line issue you're encountering?

joeflack4:
I expect no before/after effect on the git history. If it actually helps, I should only notice the effect on on my end.

I still don't understand why this is happening all of a sudden and only in some of my repos, why it doesn't listen to my global configs, and also sometimes when I pushed, I did see line end differences, but I haven't seen that for a while. If I commit and I do see differences, I'll revert and abandon this `.gitattributes` solution.

matentzn:
After reflecting on this:

While it is uncommon to do this, this allows @joeflack4 to configure his repo locally in whatever way he wants. It does not affect the other team members. The only reason to _not_ merge this is if we believe that we have to figure out the root cause of the issue, but this is the plaster that stops the bleeding and stopping the bleeding is good enough, I think we should merge this and move on. It is indeed a totally benign change for other team members!

joeflack4:
@twhetzel I'm going to just close this for now since this problem hasn't surfaced in a couple months. If it does, you've already approved this in a prior meeting, we can come back and try it out.

PR #632, opened 2024-08-12
`fatal: detected dubious ownership in repository`

### LINKED ISSUE #633: `fatal: detected dubious ownership in repository`
## Overview
This is a new error that is occurring sometimes during the "`tmp/mondo/` artefacts pipeline", i.e. when we are refreshing `mondo.owl`, `mondo-edit.owl`, and `mondo.sssom.tsv`.

### Background
It first occurred, and may only be occurring in, the `sync-synonyms` branch.

This error doesn't happen every time.

Even though this happened after we added `mondo-edit.owl` to the pipeline, it does not seem to be caused by that addition, because it does not happen every time.

## The error
`fatal: detected dubious ownership in repository at '/work/src/ontology/tmp/mondo'`

#### Instance 1: Failure during `quick-check.pl`
```
../utils/quick-check.pl mondo-edit.obo && robot --catalog catalog-v001.xml convert -i mondo-edit.obo -o mondo-edit.owl
make[2]: Leaving directory '/work/src/ontology/tmp/mondo/src/ontology'
fatal: detected dubious ownership in repository at '/work/src/ontology/tmp/mondo'
To add an exception for this directory, call:

	git config --global --add safe.directory /work/src/ontology/tmp/mondo
make[1]: *** [mondo-ingest.Makefile:373: tmp/mondo_repo_built] Error 128
```

#### Instance 2: Failure after not being able to `cat` a file
```
cat: tmp/mondo_repo_built: No such file or directory
fatal: detected dubious ownership in repository at '/work/src/ontology/tmp/mondo'
To add an exception for this directory, call:

	git config --global --add safe.directory /work/src/ontology/tmp/mondo
fatal: detected dubious ownership in repository at '/work/src/ontology/tmp/mondo'
To add an exception for this directory, call:

	git config --global --add safe.directory /work/src/ontology/tmp/mondo
```

I am pretty sure that the error occurs not as a result of the failed `cat`, but on one of the following 2 `git` lines as you can see in this snippet (I think it is the `rev-parse` line).
```make
		current_hash=$$(cat tmp/mondo_repo_built); \
		cd tmp/mondo; \
		git fetch origin; \
		latest_hash=$$(git rev-parse origin/master); \
```

#### Logs
[log1.txt](https://github.com/user-attachments/files/16603270/log1.txt)
Log 1 shows 1 instance of "error instance 1", followed by several of "error instance 2".
What's going on here is that I'm running the synonym sync for all sources. At the start of every run of the script, it checks to make sure that `tmp/mondo/` is up-to-date. The first time it does that, it fails on "error instance 1". Because of that, it is not able to create a new `tmp/mondo_repo_built` file (which contains the latest commit hash from `main`), so that file is now missing. So when the script runs subsequent times for other sources, it errors on "instance 2".

I am actually confused why instance 2, erroring after the failed `cat`, seems to happen 100% of the time after a failed `cat`, but not otherwise. Even though the refresh of `tmp/mondo/` fails after this, it does not cause an error which results in the `make` run exiting. Since the file isn't there, `curent_hash` is set to empty string. And then, because of the "dubious error" message, it appears that it is not able to set `latest_hash` either. So, both are empty strings. So when it compares the hashes, it sees that they are the same, and it determines that "no refresh is needed". Then `make` just continues.

[log2.txt](https://github.com/user-attachments/files/16603272/log2.txt)
Log 2 shows 1 instance of "error instance 2".
What's happening here is that I was doing a test. I ran the refresh of `tmp/mondo/`, but then I deleted `tmp/mondo_repo_built` manually and ran it again to see if I could replicate that error instance.
I was able to replicate it.

## Additional info
I haven't figured out why this error started happening.

It appears that errors like in instance 2 (erroring during the `tmp/mondo/` refresh, for whatever reason), cause errors like instance 1 (errors during the steps before it runs `tmp/mondo/`.

At first I tried a few things that failed. But I think I may have found a solution:
- #632

joeflack4:
Resolves #633

## Overview
Bug fix for 'fatal: detected dubious ownership in repository'
- Updated 2 related goals in the 'tmp/mondo/ artefacts pipeline'. Added lines which will prevent this error from occurring.

### Documentation
- [ ] Yes

### QC
- [ ] Yes

Build:
- #636

### New Packages
- [x] No

- [x] No

- [ ] Yes

### DIFF
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -373,6 +373,7 @@ USE_MONDO_RELEASE=false
 
 # Builds tmp/mondo/ and rebuilds mondo.owl and mondo.sssom.tsv, and stores hash of latest commit of mondo repo main branch in tmp/mondo_repo_built
 tmp/mondo_repo_built:
+	git config --global --add safe.directory /work/src/ontology/tmp/mondo
 	if [ $(USE_MONDO_RELEASE) = true ]; then wget http://purl.obolibrary.org/obo/mondo.owl -O $@; else cd $(TMPDIR) &&\
 		rm -rf ./mondo/ &&\
 		git clone --depth 1 https://github.com/monarch-initiative/mondo &&\
@@ -385,7 +386,8 @@ tmp/mondo_repo_built:
 # Triggers a refresh of tmp/mondo/ and a rebuild of mondo.owl and mondo.sssom.tsv, only if mondo repo main branch has new commits
 .PHONY: refresh-mondo-clone
 refresh-mondo-clone:
-	@if [ ! -d tmp/mondo ]; then \
+	git config --global --add safe.directory /work/src/ontology/tmp/mondo
+	if [ ! -d tmp/mondo ]; then \
 		$(MAKE) tmp/mondo_repo_built -B; \
 	else \
 		current_hash=$$(cat tmp/mondo_repo_built); \


src/ontology/mondo-ingest.Makefile
```diff
@@ -373,6 +373,7 @@ USE_MONDO_RELEASE=false
 
 # Builds tmp/mondo/ and rebuilds mondo.owl and mondo.sssom.tsv, and stores hash of latest commit of mondo repo main branch in tmp/mondo_repo_built
 tmp/mondo_repo_built:
+	git config --global --add safe.directory /work/src/ontology/tmp/mondo
```
joeflack4, line 376: ## About the bug & the fix
@twhetzel @matentzn This is a weird bug. I don't know if this is only causing a problem for the synonym sync. It's the only time the bug has appeared. But at the same time, I don't know why it would only be related to the synonym sync.

We have a meeting coming up shortly; I think it will be easier to discuss there.

PR #634, opened 2024-08-13
Bug fix: ICD11 "label not found" `RuntimeError`

joeflack4:
## Overview
Fixed bug in the problematic xrefs pipeline where an error would be thrown if there was no ICD11 label. The problem was that (i) the SPARQL query only checked for labels on owl:Class, but not also rdfs:Description, and (ii) `tmp/component-download-*.ow.owl` was being used instead of `components/*.owl`.

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
Build
- #635 
  - Implementation changed so we'll need to do a new build.

### New Packages
- [x] No

- [x] No

- [ ] Yes

### DIFF
--- a/src/analysis/problematic_exclusions.py
+++ b/src/analysis/problematic_exclusions.py
@@ -31,13 +31,12 @@ SIGNATURE_FILES_DIR = REPORTS_DIR
 OUTDIR = REPORTS_DIR
 SPARQL_STR_GET_LABELS = """
 {% for sparql_prefix_line in prefixes %}{{ sparql_prefix_line }}{% endfor %}
-prefix owl:  <http://www.w3.org/2002/07/owl#>
-prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+PREFIX owl:  <http://www.w3.org/2002/07/owl#>
+PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
 
-select ?term_id ?term_label where {
+SELECT ?term_id ?term_label WHERE {
     VALUES ?term_id { {{values}} }
-    ?term_id a owl:Class;
-      rdfs:label ?term_label .
+    ?term_id rdfs:label ?term_label .
 }"""
 # # Config
 CONFIG = {
@@ -72,9 +71,8 @@ def load_and_format_tsv(path: str, prefix_map: Dict[str, str]) -> pd.DataFrame:
     return df2
 
 
-# TODO: refactor to use utils.py get_labels()
+# todo: refactor to use utils.py get_labels()
 # todo: In the future, refactor to use OAK.
-#
 # noinspection DuplicatedCode
 def populate_labels(
     df: pd.DataFrame, onto_path: str, prefix_map: Dict[str, str], template_str=SPARQL_STR_GET_LABELS,
@@ -82,7 +80,7 @@ def populate_labels(
 ) -> pd.DataFrame:
     """Get labels for terms
 
-    # TODO: use this when OAK ready
+    # todo: use this when OAK ready
     # from oaklib.resource import OntologyResource
     # from oaklib.implementations.sqldb.sql_implementation import SqlImplementation
     # # https://incatools.github.io/ontology-access-kit/introduction.html#basic-python-example
@@ -105,7 +103,7 @@ def populate_labels(
     terms = [x for x in terms if any([x.startswith(y) for y in prefix_map.keys()])]
 
     # Instantiate template
-    prefix_sparql_strings = [f'prefix {k}: <{v}>' for k, v in prefix_map.items()]
+    prefix_sparql_strings = [f'prefix {k}: <{v}>\n' for k, v in prefix_map.items()]
     template_obj = Template(template_str)
     instantiated_str = template_obj.render(
         prefixes=prefix_sparql_strings,
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -275,7 +275,7 @@ $(REPORTDIR)/%_excluded_terms_in_mondo_xrefs.tsv $(REPORTDIR)/%_excluded_terms_i
 	$(MAKE) up-to-date-mondo.sssom.tsv
 	python3 $(RELEASEDIR)/src/analysis/problematic_exclusions.py \
 	--mondo-mappings-path $(TMPDIR)/mondo.sssom.tsv \
-	--onto-path $(TMPDIR)/component-download-$*.owl.owl \
+	--onto-path $(COMPONENTSDIR)/$*.owl \
 	--onto-config-path metadata/$*.yml \
 	--mirror-signature-path $(REPORTDIR)/mirror_signature-$*.tsv \
 	--component-signature-path $(REPORTDIR)/component_signature-$*.tsv \
@@ -283,7 +283,7 @@ $(REPORTDIR)/%_excluded_terms_in_mondo_xrefs.tsv $(REPORTDIR)/%_excluded_terms_i
 
 # Exclusions: all artefacts for single ontology
 .PHONY: exclusions-%
-exclusions-%: reports/%_exclusion_reasons.ttl reports/%_excluded_terms_in_mondo_xrefs.tsv  $(REPORTDIR)/%_term_exclusions.txt 
+exclusions-%: reports/%_exclusion_reasons.ttl reports/%_excluded_terms_in_mondo_xrefs.tsv  $(REPORTDIR)/%_term_exclusions.txt
 	@echo "$@ completed"
 
 exclusions-all: $(foreach n,$(ALL_COMPONENT_IDS), exclusions-$(n))


src/analysis/problematic_exclusions.py
joeflack4, line 1: ## Pending review
Still doing a couple runs, comparing changes to make sure they are correct.

Will add a review when done.

src/ontology/mondo-ingest.Makefile
```diff
@@ -275,15 +275,15 @@ $(REPORTDIR)/%_excluded_terms_in_mondo_xrefs.tsv $(REPORTDIR)/%_excluded_terms_i
 	$(MAKE) up-to-date-mondo.sssom.tsv
 	python3 $(RELEASEDIR)/src/analysis/problematic_exclusions.py \
 	--mondo-mappings-path $(TMPDIR)/mondo.sssom.tsv \
-	--onto-path $(TMPDIR)/component-download-$*.owl.owl \
+	--onto-path $(COMPONENTSDIR)/$*.owl \
```
joeflack4, line 278: ## Bug fix 2
The ICD11 `component-download` had `prefLabel` and `altLabel`, but not `rdfs:label`, which is what it needed. I changed to the processed component to fix that.

I was honestly surprised that this wasn't a lurker bug affecting other ontologies. I had thought that there might be other ones where we do the `prefLabel` --> `rdfs:label`

I'm not sure why for this script we originally went with `component-download` vs the processed component like we usually do.

src/analysis/problematic_exclusions.py
```diff
@@ -31,13 +31,12 @@
...
-prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+PREFIX owl:  <http://www.w3.org/2002/07/owl#>
+PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
 
-select ?term_id ?term_label where {
+SELECT ?term_id ?term_label WHERE {
     VALUES ?term_id { {{values}} }
-    ?term_id a owl:Class;
-      rdfs:label ?term_label .
+    ?term_id rdfs:label ?term_label .
```
joeflack4, line 39: ## Bug fix 1
I'm not sure why this is; maybe @matentzn knows, but there are some terms (at least this 1 term `icd11.foundation:430573082`) that are `owl:Class` in the `component-download`, but when that file gets converted to the processed component (in the `components/` dir, it no longer appears in a class, but there is a `rdf:Description` for it.

<details><summary>tmp/component-download-icd11foundation.owl.owl vs components/icd11foundation.owl</summary>
<p>
`tmp/component-download-icd11foundation.owl.owl`:

```owl
    <owl:Class rdf:about="http://id.who.int/icd/entity/430573082">
        <rdfs:subClassOf rdf:resource="http://id.who.int/icd/entity/1029961628"/>
        <rdfs:subClassOf rdf:resource="http://id.who.int/icd/entity/1661422782"/>
        <schema:browserUrl rdf:resource="https://icd.who.int/dev11/f/en#/http%3A%2F%2Fid.who.int%2Ficd%2Fentity%2F430573082"/>
        <schema:inclusion xml:lang="en">Adult T-cell lymphoma</schema:inclusion>
        <skos:altLabel xml:lang="en">Adult T-cell leukaemia</skos:altLabel>
        <skos:altLabel xml:lang="en">Adult T-cell lymphoma</skos:altLabel>
        <skos:altLabel xml:lang="en">Adult T-cell lymphoma/leukaemia</skos:altLabel>
        <skos:prefLabel xml:lang="en">Adult T-cell leukaemia/lymphoma (HTLV-1 positive)</skos:prefLabel>
    </owl:Class>
```

`components/icd11foundation.owl`:
```owl
<rdf:Description rdf:about="http://id.who.int/icd/entity/430573082">
        <oboInOwl:hasExactSynonym xml:lang="en">Adult T-cell leukaemia</oboInOwl:hasExactSynonym>
        <oboInOwl:hasExactSynonym xml:lang="en">Adult T-cell leukaemia/lymphoma (HTLV-1 positive)</oboInOwl:hasExactSynonym>
        <oboInOwl:hasExactSynonym xml:lang="en">Adult T-cell lymphoma</oboInOwl:hasExactSynonym>
        <oboInOwl:hasExactSynonym xml:lang="en">Adult T-cell lymphoma/leukaemia</oboInOwl:hasExactSynonym>
        <oboInOwl:hasExactSynonym>adult t-cell leukaemia/lymphoma</oboInOwl:hasExactSynonym>
        <rdfs:label xml:lang="en">Adult T-cell leukaemia/lymphoma (HTLV-1 positive)</rdfs:label>
    </rdf:Description>
    <owl:Axiom>
        <owl:annotatedSource rdf:resource="http://id.who.int/icd/entity/430573082"/>
        <owl:annotatedProperty rdf:resource="http://www.geneontology.org/formats/oboInOwl#hasExactSynonym"/>
        <owl:annotatedTarget>adult t-cell leukaemia/lymphoma</owl:annotatedTarget>
        <oboInOwl:hasSynonymType rdf:resource="http://purl.obolibrary.org/obo/mondo#GENERATED"/>
    </owl:Axiom>
```

</p>
</details> 

In order to fix this problem, I made the SPARQL query less strict, dropping the `owl:Class` requirement.

The results are that this fixes the ICD11 issue we're getting, but otherwise everything else is unaffected.

src/analysis/problematic_exclusions.py
joeflack4, line 1: ## Overall review
I made 2 changes that together fix the bug we were having, caused when an excluded terms with a Mondo xref is being reported, but there was a failure in fetching its label.

<details><summary>Error log showing bug</summary>
<p>

```
python3 ../../src/analysis/problematic_exclusions.py \
--mondo-mappings-path tmp/mondo.sssom.tsv \
--onto-path tmp/component-download-icd11foundation.owl.owl \
--onto-config-path metadata/icd11foundation.yml \
--mirror-signature-path reports/mirror_signature-icd11foundation.tsv \
--component-signature-path reports/component_signature-icd11foundation.tsv \
--outpath reports/icd11foundation_excluded_terms_in_mondo_xrefs.tsv
Traceback (most recent call last):
  File "/work/src/ontology/../../src/analysis/problematic_exclusions.py", line 140, in populate_labels
    df2 = pd.read_csv(results_path).fillna('')
  File "/usr/local/lib/python3.10/dist-packages/pandas/io/parsers/readers.py", line 1026, in read_csv
    return _read(filepath_or_buffer, kwds)
  File "/usr/local/lib/python3.10/dist-packages/pandas/io/parsers/readers.py", line 620, in _read
    parser = TextFileReader(filepath_or_buffer, **kwds)
  File "/usr/local/lib/python3.10/dist-packages/pandas/io/parsers/readers.py", line 1620, in __init__
    self._engine = self._make_engine(f, self.engine)
  File "/usr/local/lib/python3.10/dist-packages/pandas/io/parsers/readers.py", line 1898, in _make_engine
    return mapping[engine](f, **self.options)
  File "/usr/local/lib/python3.10/dist-packages/pandas/io/parsers/c_parser_wrapper.py", line 93, in __init__
    self._reader = parsers.TextReader(src, **kwds)
  File "parsers.pyx", line 581, in pandas._libs.parsers.TextReader.__cinit__
pandas.errors.EmptyDataError: No columns to parse from file

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/work/src/ontology/../../src/analysis/problematic_exclusions.py", line 288, in <module>
    cli()
  File "/work/src/ontology/../../src/analysis/problematic_exclusions.py", line 283, in cli
    return problematic_exclusions(**d)
  File "/work/src/ontology/../../src/analysis/problematic_exclusions.py", line 233, in problematic_exclusions
    df = populate_labels(df=df, onto_path=onto_path, prefix_map=prefix_uri_map)
  File "/work/src/ontology/../../src/analysis/problematic_exclusions.py", line 145, in populate_labels
    raise RuntimeError('No results found when trying to get labels for terms.')
RuntimeError: No results found when trying to get labels for terms.
make[1]: *** [mondo-ingest.Makefile:276: reports/icd11foundation_excluded_terms_in_mondo_xrefs.tsv] Error 1
make[1]: Leaving directory '/work/src/ontology'
make: *** [mondo-ingest.Makefile:348: build-mondo-ingest] Error 2
Command exited with non-zero status 2
```

</p>
</details> 

I'm running the build now, but I did a first pass before and after the changes to ensure that this fix worked and that no additional bugs appeared:

```
diff excluded_terms_in_mondo_xrefs1.tsv excluded_terms_in_mondo_xrefs2.tsv
10a11
> icd11.foundation:430573082	Adult T-cell leukaemia/lymphoma (HTLV-1 positive)	True	False	True	True
joeflack4@MacBook-Pro Desktop % diff excluded_terms_in_mondo_xrefs1.tsv excluded_terms_in_mondo_xrefs.tsv
10a11
````

This is what I expected to see. There was only 1 such case in ICD11, and now it is appearing in the results. Further, no other rows were added or removed for other sources. [files.zip](https://github.com/user-attachments/files/16594030/files.zip)

File: src/ontology/mondo-ingest.Makefile (hunk 1 above)
Comment (matentzn) [reply src/ontology/mondo-ingest.Makefile:278]: Makes sense! This seems to be the root cause of the issue, and would make sense to me. Good detective work!

File: src/analysis/problematic_exclusions.py (hunk 2 above)
Comment (matentzn) [reply src/analysis/problematic_exclusions.py:39]: I don't agree with this fix - it's should be totally ineffectual; if this is needed, we have a much bigger problem.. are you saying the ICD11 ingest owl has no rdf:type owl:Class assertions? If so, we should load safe it with robot once before processing. 

Minor comment: I would suggest always to avoid spurious diffs; this has obviously been cycled to some AI tool, and out of the 4 changes in this file only 1 matters. Same with the Python script below..

**Review by matentzn — COMMENTED** [review]
See comments

File: src/analysis/problematic_exclusions.py (hunk 2 above)
Comment (twhetzel) [reply src/analysis/problematic_exclusions.py:39]: The "tmp" file uses `owl:Class` and the "component" file uses `rdf:Description` for some entities and `icd11.foundation:430573082` is one of those that changes between the two files.

File: src/ontology/mondo-ingest.Makefile (hunk 1 above)
Comment (twhetzel) [reply src/ontology/mondo-ingest.Makefile:278]: I'm trying to understand how changing the makefile to use the component directory instead of tmp for "onto-path" fixes the issue if a problematic exclusion is defined as a curie that exists in mondo.sssom.tsv and the mirror(tmp) file, but not the component file based on this [code section](https://github.com/monarch-initiative/mondo-ingest/blob/main/src/analysis/problematic_exclusions.py#L216-L220). If the "problematic" entity is not in the component file, then how does [querying the file in the component directory](https://github.com/monarch-initiative/mondo-ingest/blob/main/src/analysis/problematic_exclusions.py#L233) to get it's label work?

Also, what happens to these "problematic" xrefs and are these icd11.foundation xrefs actually problematic since theses curies are in the component file but use rdf:Description instead?

Comment (matentzn) [reply src/ontology/mondo-ingest.Makefile:278]: It was my understanding that this is the solution because it has all the preprocessing in it, in particular, it changes skos:prefLabel to rdfs:label so that it can "find a label"?

Comment (matentzn) [reply src/ontology/mondo-ingest.Makefile:278]: I cannot verify this (neither negatively nor positively).

Eyeballing the goal, it is not clear to me which ontology is needed. As this target seems to be about "exclusions", it may well _need_ the mirror (but I do not know that).

So my suggestion is:

Review the code (python script), determine what  the code gets from the ontology exactly, and based on that, decide:

1. Should the information be extracted from the component (where all relationships are correctly harmonised)
2. Should the information be extract from the mirror (which includes all classes, including the ones _excluded_ from the component
3. Or, god forbid, the information should be extracted from a _harmonized mirror_ which would be... A bit more work. I can do it though if we get to this determination.

File: src/analysis/problematic_exclusions.py (hunk 2 above)
Comment (matentzn) [reply src/analysis/problematic_exclusions.py:39]: 😱 looking into this now

Comment (matentzn) [reply src/analysis/problematic_exclusions.py:39]: Found it..

```
		remove -T $(TMPDIR)/icd11foundation_relevant_signature.txt --select complement --select "classes individuals" --trim false \
```

(problem is --trim false). Drilling deeper.

File: src/ontology/mondo-ingest.Makefile (hunk 1 above)
Comment (twhetzel) [reply src/ontology/mondo-ingest.Makefile:278]: Sure, I have been reviewing the script. Based on the initial error reported I would be inclined to change the [SPARQL query](https://github.com/monarch-initiative/mondo-ingest/blob/main/src/analysis/problematic_exclusions.py#L32-L41) to:
```
prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
prefix skos: <http://www.w3.org/2004/02/skos/core#>

    ?term_id a owl:Class;
    OPTIONAL { ?term_id rdfs:label ?term_label . }
    OPTIONAL { ?term_id skos:prefLabel ?term_label . }
    FILTER(BOUND(?term_label))
```

so that the term label can be found regardless of whether the mirror uses rdfs:label or skos:prefLabel (which icd11.foundation and ICD10CM use). 

This change also allows the `reports/icd11foundation_excluded_terms_in_mondo_xrefs.tsv` to complete successfully.

src/analysis/problematic_exclusions.py
matentzn: Ok. 

Removing --trim false from the line in the comment above will sort this issue, and no need to change the SPARQL query. This is not related to the other issue (the question of wether to use the mirror or the component).

The disease you named (430573082 ) is an extension code and therefore _not_ in the list of entities of interest.

src/ontology/mondo-ingest.Makefile
matentzn: I agree but this does not answer the question fully which ontology (component or mirror) _should_ be used here. Also, it creates the risk of having two labels, as some ontologies have rdfs:label and skos:prefLabel at the same time..

twhetzel: A solution from Nico is [mentioned above](https://github.com/monarch-initiative/mondo-ingest/pull/634/files#r1715508237) and no change in the SPARQL query is needed.

twhetzel: Running some tests on the options that have been proposed so far this is what I am seeing.

TEST: Run "reports" make goal as: 
`sh run.sh make reports/icd11foundation_excluded_terms_in_mondo_xrefs.tsv 2>&1 | tee icd11foundation_logs.txt`

Option 1: 
- Change only the make goal "$(COMPONENTSDIR)/icd11foundation.owl" to remove "--trim False"

-> Result: 
The set of problematic excluded terms is now larger.
Fails with "RuntimeError: No results found when trying to get labels for terms."


Option 2:
- Change the make goal "$(COMPONENTSDIR)/icd11foundation.owl" to remove "--trim False"
- Change make goal "$(REPORTDIR)/%_excluded_terms_in_mondo_xrefs.tsv" to use "--onto-path $(COMPONENTSDIR)/$*.owl \"

-> Result: 
The set of problematic excluded terms is now larger.
Fails with RuntimeError: No results found when trying to get labels for terms.


Option 3:
- Change the make goal "$(COMPONENTSDIR)/icd11foundation.owl" to remove "--trim False"
- Change the SPARQL query in the script to account for labels using rdfs:label or skos:prefLabel in the mirror files as:
SELECT ?term_id ?term_label WHERE {
    ?term_id a owl:Class .
    OPTIONAL { ?term_id rdfs:label ?term_label . }
    OPTIONAL { ?term_id skos:prefLabel ?term_label . }
    FILTER(BOUND(?term_label))
}

-> Result: This completes successfully.

I thought this SPARQL query would look for rdfs:label and use that if it's found and only if not found then look for skos:prefLabel, therefore reducing the risk if a class uses two label properties.


---
I am still not clear on Joe's logic to change TMPDIR to COMPONENTSDIR when it looks like a problematic exclusion is defined as a curie that exists in mondo.sssom.tsv _and_ the mirror(tmp) file, but not the component file based on this [code section](https://github.com/monarch-initiative/mondo-ingest/blob/main/src/analysis/problematic_exclusions.py#L216-L220).

joeflack4: So much going on here. Good work though both of you.

I ran a test yesterday and got the same results when changing from the mirror to the component, but I agree we should / may have to stick to the mirror.

> Joe's logic to change TMPDIR to COMPONENTSDIR 

@twhetzel The reason I did this is because the component always uses `rdfs:label` whereas the mirror doesn't. I was unsure whether or not `prefLabel` was the only alternative label property that might ever appear in the mirror, either. But I knew that whatever label properties might ever exist now or in the future, we'll always standardize them to `rdfs:label` in the component. So I decided to try this change first and make sure I got good results. I ran script and the results were the same, so I went with this as the first implementation.

joeflack4: Adding to @matentzn's list of options, "should we query the mirror, component, or harmonized mirror", I would add:
4. If needed, we could use both the mirror and component for different queries.

joeflack4: > Option 3: ... remove `--trim False` ... `rdfs:label` or `skos:prefLabel`...

@twhetzel I like this solution.
The `--trim False` is independent of your query since it applies to the component and not the mirror that you are querying in option 3.
As I mentioned above, I was worried there might be additional label properties now or in the future, but you and Nico may know better than I that this isn't a concern. Also if that does happen, we could just add another `OPTIONAL` clause to your query.

@matentzn thanks for the `--trim False` fix. I wonder if this will affect only "problematic exclusions" or might have effects elsewhere.

src/analysis/problematic_exclusions.py
joeflack4: > The disease you named (430573082) is an extension code and therefore not in the list of entities of interest.

I suppose that is why it is an exclusion. The point of the script is to report on excluded terms that have mondo xrefs. So it makes sense that this disease is appears here.
Are you mentioning this just as an FYI? Or is there something wrong that needs to be changed?

joeflack4: > avoid spurious diffs; this has obviously been cycled to some AI tool

@matentzn No, that's not the case; these are all intentional, manual changes. I can explain any/each of these changes if you like. I've heard from you before that basically I don't need a separate PR for codestyle changes that are closely related to what I am working on in the PR. If I noticed codestyle issues in something that was not "problematic exclusions", I would put them in a separate PR.

The only reason I didn't explain why I made these changes is because:
- I'm trying to write less overall / keep to what is important/necessary.
- It was suggested that I don't need to comment about some obvious things because they can clearly be seen in the diff.

src/ontology/mondo-ingest.Makefile
twhetzel: @joeflack4 what results did you get when you say "got the same results when changing from the mirror to the component"?

joeflack4: From: [Overall Review](https://github.com/monarch-initiative/mondo-ingest/pull/634#discussion_r1714622167)
> This is what I expected to see. There was only 1 such case in ICD11, and now it is appearing in the results. Further, no other rows were added or removed for other sources.

src/analysis/problematic_exclusions.py
joeflack4, line 1: ## Codestyle & misc changes
> avoid spurious diffs; this has obviously been cycled to some AI tool

@matentzn No, that's not the case; these are all intentional, manual changes. I can explain any/each if you like. I know about this rule that I should keep PR changes as close to what is necessary to the task at hand and no more. But I've also heard from you before that basically I don't need a separate PR for codestyle changes that are closely related ("problematic exclusions") to what I am working on in the PR.

The only reason I didn't explain why I made these changes is because:
- I'm trying to write less overall / keep to what is important/necessary.
- It was suggested that I don't need to comment about some obvious things because they can clearly be seen in the diff.

twhetzel: From my perspective, I certainly would not expect any codestyle changes in a PR like this.

src/ontology/mondo-ingest.Makefile
twhetzel: Thanks for explaining your logic for the change TMPDIR to COMPONENTSDIR. I was trying to stay as true to the original code and and how problematic exclusions were identified so I focused on options to have the helper function that looks up term labels work against the files in TMPDIR. 

I realize I did not test the change to the COMPONENTSDIR _and_  your change in the SPARQL query. After your further explanations of choices and my digging through this code further I see why you proposed this change.

I gather it's about 50/50 to include the changes in this PR vs. Option 3. What are the risks that a term that is a problematic exclusion will not exist in the files in COMPONENTSDIR based on the make goals to create the components files?

The test build for Option 3 (just) successfully completed. Is there a PR for the build from this branch (bugfix-icd11-label)?

joeflack4: @twhetzel Let's use your approach.
Sometimes what I like to do is merge something that is currently working, and then improve it with a second PR.
I was considering suggesting that we do that w/ my implementation, but I saw your changes which I think are better, and that they giving good results. And now that your build has passed, I think we should take a look at it and if it looks correct, let's push your commit to this PR and merge that.

I did run a build using my approach, but I closed it in favor of yours:
- #635

I think build PRs should really only have 1 commit. They should be based off of 1 implementation. Your changes will produce different results, and I think it's better to have them in a different PR.

twhetzel: Ok, let me push the build I did from Option 3.

twhetzel:
I converted this to a Draft for now now so that does not accidentally get merged. I'll decide after 11am PT if this branch will be used at all and close as needed.

Issue #638, opened 2024-08-15
Update problematic_exclusions.py due to error with icd11.foundation


twhetzel:
Since icd11.foundation now exists in the mondo.sssom.tsv file, a build error was found in the problematic_exclusions.py script for icd11.foundation content. This error was [fixed in the test build branch directly](https://github.com/monarch-initiative/mondo-ingest/pull/637/files) from `main`, but is being addressed here for the `develop` branch. I also want to check the results of not using `--trim false` in the component goal for icd11.foundation (which was also included in the hotfix to main).

UPDATE: We discussed at the Tech call on 16-Aug-2024 that `--trim false` should also be included to resolve the content issue with the icd11.foundation COMPONENT so this change will also be included.

twhetzel:
Closed with https://github.com/monarch-initiative/mondo-ingest/pull/641

PR #640, opened 2024-08-16
DO NOT MERGE - Feature branch build for issue-638

twhetzel:
Resolves #ISSUE(s).

## Overview

This PR:
- Feature branch build for https://github.com/monarch-initiative/mondo-ingest/issues/638

### Documentation
Was the documentation added/updated under `docs/`?
- [ ] Yes
- [x] No, updates to the docs were not necessary after careful consideration

### QC
Was the full pipeline run before submitting this PR using `sh run.sh make build-mondo-ingest` on this branch (after 
`docker pull obolibrary/odkfull:dev`), and no errors occurred?
- [x] Yes
- [ ] No, there are no functional (code-related) changes to the pipeline in the PR, so no re-run is necessary
   
### New Packages
Were any new *Python packages* added?
- [ ] Yes, [requirements.txt.full](https://github.com/INCATools/ontology-development-kit/blob/master/requirements.txt.full) was updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

Were any other *non-Python packages* added?
- [ ] Yes, [Dockerfile](https://github.com/INCATools/ontology-development-kit/blob/master/Dockerfile) updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

### PR Review and Conversations Resolved
Has the PR been sufficiently reviewed by at least 1 team member of the Mondo Technical team and all threads resolved?
- [ ] Yes

PR #637, opened 2024-08-13
TEST Build from main branch with code updates to icd11 component goal and problematic exclusion script

twhetzel:
Resolves #ISSUE(s).

## Overview

This PR:
Includes code updates to:
- `mondo-ingest.Makefile` to remove `--trim false` in the goal `$(COMPONENTSDIR)/icd11foundation.owl`
- update to SPARQL query used for label look-up for curies identified as problematic exclusions to check for labels using first `rdfs:label` and then `skos:prefLabel`. Currently, ICD10CM and icd11.foundation both use `skos:prefLabel` in the files in TMPDIR

This PR also includes data updates based on a full build of the mondo-ingest pipeline run as: `sh run.sh make build-mondo-ingest -B 2>&1 | tee logs.txt`

NOTE: Keep in mind this is the first data build since icd11.foundation was included in "mondo.sssom.tsv" following the inclusion of xrefs to icd11.foundation for the August Mondo release and subsequent [updates in the mondo repo](https://github.com/monarch-initiative/mondo/pull/8054) to make sure the icd11.foundation xrefs were included in the "mondo.sssom.tsv" file.

PROPOSAL: Given this very special situation where this is Week 2 (Data) of the Mondo Release Cycle where lexical alignments are reviewed and generally new builds are done _and_ that this hotfix code change is needed to fix the build on `main`, I suggest that if we approve this PR and it's ok with Sabrina that both the code and data files be merged into `main`. @matentzn and @twhetzel will discuss Wednesday.
- Update: Sabrina is ok if this is merged into `main`.

### Documentation
Was the documentation added/updated under `docs/`?
- [ ] Yes
- [x] No, updates to the docs were not necessary after careful consideration

### QC
Was the full pipeline run before submitting this PR using `sh run.sh make build-mondo-ingest` on this branch (after 
`docker pull obolibrary/odkfull:dev`), and no errors occurred?
- [x] Yes
- [ ] No, there are no functional (code-related) changes to the pipeline in the PR, so no re-run is necessary
   
### New Packages
Were any new *Python packages* added?
- [ ] Yes, [requirements.txt.full](https://github.com/INCATools/ontology-development-kit/blob/master/requirements.txt.full) was updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

Were any other *non-Python packages* added?
[...]

PR #641, opened 2024-08-16
account for mirror source content with rdfs:label or skos:prefLabel

twhetzel:
Resolves #ISSUE(s).

## Overview

This PR:
- Updates the SPARQL query to account for mirror source content that uses skos:prefLabel
- Feature branch build https://github.com/monarch-initiative/mondo-ingest/pull/640

- Closes https://github.com/monarch-initiative/mondo-ingest/issues/638

~NOTE: This PR only includes the change to the SPARQL query and not also the change to the icd11.foundation component goal that [removed `--trim false`](https://github.com/monarch-initiative/mondo-ingest/pull/637/files#diff-36e739e129da51616de07dd1ff7ab0c12aa9b5dd45b7f110c9a025fad909ecc4R165).~

We discussed at the Tech call on 16-Aug-2024 that `--trim false` should be removed. This change is now included.

### Documentation
- [ ] Yes

### QC
- [x] Yes
   
### New Packages
- [x] No

- [x] No

- [x] Yes

### DIFF
--- a/src/analysis/problematic_exclusions.py
+++ b/src/analysis/problematic_exclusions.py
@@ -33,11 +33,16 @@ SPARQL_STR_GET_LABELS = """
 {% for sparql_prefix_line in prefixes %}{{ sparql_prefix_line }}{% endfor %}
 prefix owl:  <http://www.w3.org/2002/07/owl#>
 prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+prefix skos: <http://www.w3.org/2004/02/skos/core#>
 
 select ?term_id ?term_label where {
     VALUES ?term_id { {{values}} }
     ?term_id a owl:Class;
-      rdfs:label ?term_label .
+    OPTIONAL {
+        VALUES ?label_property { rdfs:label skos:prefLabel } .
+        ?term_id ?label_property ?term_label .
+        }
+    FILTER(BOUND(?term_label))
 }"""
 # # Config
 CONFIG = {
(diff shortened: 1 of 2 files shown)


src/ontology/mondo-ingest.Makefile
joeflack4, line 1: ## Some missing code changes
There were some [code changes](https://github.com/monarch-initiative/mondo-ingest/pull/640#discussion_r1720838868) in the build PR that I believe are missing from this feature branch.

src/analysis/problematic_exclusions.py
```diff
@@ -33,11 +33,16 @@
...
 prefix owl:  <http://www.w3.org/2002/07/owl#>
 prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+prefix skos: <http://www.w3.org/2004/02/skos/core#>
 
 select ?term_id ?term_label where {
     VALUES ?term_id { {{values}} }
     ?term_id a owl:Class;
-      rdfs:label ?term_label .
+    OPTIONAL {
+        VALUES ?label_property { rdfs:label skos:prefLabel } .
```
joeflack4, line 42: ## Possible future proofing
I consulted LLM to see about any additional primary label properties that might exist, in case we want to add them here.

It listed off about 6, and out of those, these 3 looked like they would apply generally:
- `dc:title`
- `dct:title`
- `schema:name`

I think `dc:title` is the only one I've seen before, but in any case it doesn't seem like these are really used by biomedical ontologies. Still, I suppose it couldn't hurt to add them, given the `OPTIONAL` clause. I'm somewhat ambivalent, though.

matentzn: Title would never be used for ontology / vocabulary terms. schema:name is more likely, but not in the clinical domain.. I think its probably ok the way it is!

twhetzel: Nico and I are very familiar with how ontologies are structured. While there is a small chance some new source will use a different label property, the ingest documentation can include that if a different property is used this query needs to be updated.

matentzn (approved):

joeflack4 (approved):

PR #642, opened 2024-08-17
Problematic exclusions codestyle updates

joeflack4:
## Overview
Just some codestyle fixes that didn't make their way in w/ close of #634: added / removed whitespace, standardized todo's related to level of urgency.

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [x] No

- [x] No

- [x] Yes

### DIFF
--- a/src/analysis/problematic_exclusions.py
+++ b/src/analysis/problematic_exclusions.py
@@ -72,9 +72,8 @@ def load_and_format_tsv(path: str, prefix_map: Dict[str, str]) -> pd.DataFrame:
     return df2
 
 
-# TODO: refactor to use utils.py get_labels()
+# todo: refactor to use utils.py get_labels()
 # todo: In the future, refactor to use OAK.
-#
 # noinspection DuplicatedCode
 def populate_labels(
     df: pd.DataFrame, onto_path: str, prefix_map: Dict[str, str], template_str=SPARQL_STR_GET_LABELS,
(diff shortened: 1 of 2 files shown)


src/analysis/problematic_exclusions.py
```diff
@@ -105,7 +104,7 @@ def populate_labels(
     terms = [x for x in terms if any([x.startswith(y) for y in prefix_map.keys()])]
 
     # Instantiate template
-    prefix_sparql_strings = [f'prefix {k}: <{v}>' for k, v in prefix_map.items()]
+    prefix_sparql_strings = [f'prefix {k}: <{v}>\n' for k, v in prefix_map.items()]
```
matentzn, line 107: why isnt this solved on the level of the template? This is a list of prefix strings, right? The template engine should be able to turn a list of strings into a newline separated string output?

joeflack4: We can. I think it will look something like `{% for string in prefix_sparql_strings %}{{string}}{-% endfor %}` which makes the thing harder to read, so in this particular case I like this better.

Also for time purposes I'd like to leave as is, but after looking at my explanation, if you you do still want this logic moved into the Jinja template, let me know and I can do.

matentzn: Ok, its fine; I am just noting that, from a software engineering perspective, to ensure seperation of concerns, the layer that is responsible for defining how the final string (the whole sparql query) should be represented is the template, while the prefix_maps iteration is still on the data side. In any case, technically the whole list comprehension could be in the template, but from my POV this is just academic. I am satisfied with the explanation given that the hole list comprehension is conceptually part of the templating system.

matentzn (approved):

twhetzel (approved):
lgtm

PR #93, opened 2022-10-06
Synchronization: Synonyms

joeflack4:
Resolves #88

## Overview
Keeps synonym for all terms from `mondo-ingest` sources synchronized in Mondo.

### Documentation
Was the documentation added/updated under `docs/`?
- [x] Yes
- [ ] No, updates to the docs were not necessary after careful consideration

### QC
Was the full pipeline run before submitting this PR using `sh run.sh make build-mondo-ingest` on this branch (after 
`docker pull obolibrary/odkfull:dev`), and no errors occurred?
- [x] Yes
- [ ] No, there are no functional (code-related) changes to the pipeline in the PR, so no re-run is necessary
   
Build PR:
- #689

There have been no changes to this PR since #671. It has already run a successful build. Once #671 is merged into this branch, then it is effectively the same as having run the build on this branch.

### New Packages
Were any new *Python packages* added?
- [ ] Yes, [requirements.txt.full](https://github.com/INCATools/ontology-development-kit/blob/master/requirements.txt.full) was updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

Were any other *non-Python packages* added?
- [ ] Yes, [Dockerfile](https://github.com/INCATools/ontology-development-kit/blob/master/Dockerfile) updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

### PR Review and Conversations Resolved
Has the PR been sufficiently reviewed by at least 1 team member of the Mondo Technical team and all threads resolved?
- [x] Yes

## Subtasks
See **[Google Doc](https://docs.google.com/document/d/1xdSROYkk16qoQY0jGQap6ihNouG04OLH_jE3uAMpU20/edit)** or the issue (#88).

## Additional information
- [Google sheet](https://docs.google.com/spreadsheets/d/1ftEZ47-JDe-1Z_fRWyO5NS39qT_-8MZsHXHEKz9aZ9Q/edit?gid=978570179#gid=978570179)

### Change details
    Synonym synchronization
    - Add: make goals
    - Add: sync_synonym.py

    - Update: Changed method of querying synonym source axioms from OAK to SPARQL.
    - Update: 4 output cases: -added, -confirmed, -deleted, -updated
    - Update: Handled lots of edge cases, including: filtering obsoletions, filtering deletions, adding labels.

PR #646, opened 2024-08-22
Add draft reference for excluded values

matentzn:
For various different projects, we need a reference for excluded values. @twhetzel and I have discussed to remove "excluded synonyms" from mondo-edit.obo and move them here, and curate them. The reason I chose YAML is because I want to be able to flexibly store a variety of different excluded data types here with full flexibility on structure.

What do you think @twhetzel and @joeflack4?

## Overview

This PR:
- Adds a draft registry for _excluded values_ to Mondo ingest, which will be used by multiple pipelines.

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [x] No

- [x] No

- [ ] Yes


joeflack4 (approved):
## General review
This is great. Trish just filled me in on the plan here.

This will allow me to simplify this part of the synonym sync pipeline, which should also should remove the `git` 'dubious ownership' issue.

Some observations:
- This is many more exclusions than what I had queried recently (12 rows, 8 unique synonyms). I assume this is the result of some recent curation.
- This excludes these synonyms 'generally', as opposed to excluding them on a particular Mondo ID.

src/ontology/config/mondo-excluded-values.yml
joeflack4, line 1: ## Merge into `develop` instead?
I need to get this config into my synonym sync branch. But I can't merge `main` into that branch. Can we change the base to `develop` or also make a PR there?

matentzn: Done, thanks!

twhetzel (commented):
This file does not include the [synonyms identified as "excluded" in the Synonym Sync PR](https://github.com/monarch-initiative/mondo-ingest/pull/93/files#diff-6b6bca58ff4e8b1e979e9791da289a3705df30f7d0058c067324e8c8290e38b4). There are ~8 synonyms now in mondo-edit.obo that are annotated with "excluded_synonym". 

How were these "excluded" synonyms identified?

matentzn:
How were these "excluded" synonyms identified?

```
grep EXCLUDE mondo-edit.obo | tail
```

I harmonise these 8 ones the, what seems to me, correct representation:

https://github.com/monarch-initiative/mondo/pull/8105

let me know what you think!

matentzn:
P.S. I also now added these 8 missing values: https://github.com/monarch-initiative/mondo-ingest/pull/646/commits/14916c5f558a98ffa2ef610ea6b50dd933c5b2c5

src/ontology/config/mondo-excluded-values.yml
joeflack4, line 1: ## Values in this PR vs synonym sync branch
Just making a thread off the comments about this. Trish and I are observing that there was a difference.

[Nico asked](https://github.com/monarch-initiative/mondo-ingest/pull/646/#issuecomment-2307774411):
> How were these "excluded" synonyms identified?

A: Via [this SPARQL query](https://github.com/monarch-initiative/mondo-ingest/blob/sync1-synonyms/src/sparql/mondo-excluded-synonyms.sparql).

joeflack4: I can confirm as well the previously missing 8 values are there! Thanks.

Also wondering how this PR has more than what we were able to query via that SPARQL query. I would have expected the list to be the same.

twhetzel: Excluded synonyms were represented in two different ways. Joe you identified the less common way and Nico used the more common option.

twhetzel (approved):
lgmt

PR #648, opened 2024-08-24
Replace `bioregistry` dependency with `curies`

### LINKED ISSUE #584: Replace `bioregistry` dependency with `curies`
## Overview
`mondo-ingest` is currently installing `bioregistry`. `match-mondo-sources-all-lexical.py` depends on it. However, whatever functionality being used by that package should also be found in `curies`, which comes with ODK. We should migrate to `curies`.

## Sub-tasks
- [ ] 1. `match-mondo-sources-all-lexical.py`: Swap out whatever code is using `bioregistry` to use `curies`
- [ ] 2. `mondo-ingest.Makefile`: Drop `$(MAKE) pip-bioregistry`

joeflack4:
Resolves #584

## Overview
Replace any bioregistry usage with curies instead, and remove any installations of bioregistry.

### Documentation
- [ ] Yes

### QC
- [x] Yes
   
Build:
- #649 

### New Packages
- [x] No

- [x] No

- [ ] Yes

## Additional info
Didn't find any actual current usages of bioregistry.
There were some comments/todo's referencing bioregistry; updated those to curies.

### DIFF
--- a/python-requirements.txt
+++ b/python-requirements.txt
@@ -11,7 +11,6 @@ attrs==21.4.0
 Babel==2.10.1
 bcp47==0.0.4
 beautifulsoup4==4.12.2
-bioregistry==0.6.99
 bleach==5.0.0
 cattrs==22.2.0
 certifi==2021.10.8
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -461,7 +461,6 @@ tmp/merged.owl: mondo-ingest.owl tmp/mondo.sssom.ttl tmp/mondo.owl
 $(MAPPINGSDIR)/mondo-sources-all-lexical.sssom.tsv: $(SCRIPTSDIR)/match-mondo-sources-all-lexical.py tmp/merged.db $(MAPPINGSDIR)/rejected-mappings.tsv
 	rm -f $(MAPPINGSDIR)/mondo-sources-all-lexical.sssom.tsv
 	rm -f $(MAPPINGSDIR)/mondo-sources-all-lexical-2.sssom.tsv
-	$(MAKE) pip-bioregistry
 	python $(SCRIPTSDIR)/match-mondo-sources-all-lexical.py run tmp/merged.db \
 		-c metadata/mondo.sssom.config.yml \
 		-r config/mondo-match-rules.yaml \
--- a/python-requirements-apple-silicon.txt
+++ b/python-requirements-apple-silicon.txt
@@ -6,7 +6,6 @@ attrs==23.1.0
 Babel==2.13.1
 bcp47==0.0.4
 beautifulsoup4==4.12.2
-bioregistry==0.10.110
 cachetools==5.3.2
 cattrs==23.2.2
 certifi==2023.11.17
--- a/python-requirements-unlocked.txt
+++ b/python-requirements-unlocked.txt
@@ -1,4 +1,3 @@
-bioregistry
 curies
 jinja2
 oaklib>=0.5.20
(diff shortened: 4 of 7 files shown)


matentzn (approved):
Thank you!! I confirmed that the change to src/scripts/match-mondo-sources-all-lexical.py is indeed non-functional, but I will say that for the cautious person I am that, in theory, the change to `src/ontology/mondo-ingest.Makefile` is not _not functional_; I would make it a habit before you close for the day to kick off a run on a branch like this, because sorting out an (albeit extremely unlikely, hence by approval) issue later may be unnecessarily time consuming.

EDIT: as an explanation why it is not "NOT functional" is the `--upgrade` option in

```
.PHONY: pip-%
pip-%:
```

This will also upgrade dependencies if necessary, which means that in very theory, the code may have been wrong on slightly different python packages then are installed by default in the docker image.

src/ontology/mondo-ingest.Makefile
```diff
@@ -461,7 +461,6 @@ tmp/merged.owl: mondo-ingest.owl tmp/mondo.sssom.ttl tmp/mondo.owl
 $(MAPPINGSDIR)/mondo-sources-all-lexical.sssom.tsv: $(SCRIPTSDIR)/match-mondo-sources-all-lexical.py tmp/merged.db $(MAPPINGSDIR)/rejected-mappings.tsv
 	rm -f $(MAPPINGSDIR)/mondo-sources-all-lexical.sssom.tsv
 	rm -f $(MAPPINGSDIR)/mondo-sources-all-lexical-2.sssom.tsv
-	$(MAKE) pip-bioregistry
```
joeflack4, line 464: ## `pip install --upgrade`
@matentzn I see what you're saying in [this comment](https://github.com/monarch-initiative/mondo-ingest/pull/648#pullrequestreview-2258517161). Normally we do a build in these PRs, but it's possible we may have opted to skip that here if we determined these were all non-functional changes.

What you're saying is that, since `$(MAKE) pip-bioregistry` was added, likely upgraded some of its dependencies, which may have been existing dependencies that we use elsewhere in `mondo-ingest`, and that these upgrades may have different behavior than their non-uprgraded versions. So when removing `$(MAKE) pip-bioregistry`, this could result things behaving differently--possibly differences in outputs, or even errors.

So to be sure, we should run a build. I'll do that now.

joeflack4: Build is ready for review:
- #649

matentzn: Awesome, thanks! All good!

python-requirements.txt
```diff
@@ -11,7 +11,6 @@ attrs==21.4.0
 Babel==2.10.1
 bcp47==0.0.4
 beautifulsoup4==4.12.2
-bioregistry==0.6.99
```
twhetzel, line 14: @joeflack4 why is the `curies` package not included in the python-requirements.txt file in addition to the python-requirements-unlocked.txt file?

twhetzel (commented):
comment inline

python-requirements.txt
joeflack4: Ah. I don't maintain this file anymore. It doesn't say it (maybe it should), but this requirements file applies to Intel / x86 architecture.

After I got my new computer last December, I'm now on ARM64 Apple Silicon. I wasn't able to install via this requirements file. `pandas` (`numpy`) may have been the only thing that wouldn't install. Since then, I created and started maintaining a separate `python-requirements-apple-silicon.txt` (it should really say `-arm64`).

Even though I can't `pip freeze > python-requirements.txt`, I could manually insert the `curies` line in there.

Perhaps we should make an issue about the best way to handle these multiple `requirements.txt` files. Perhaps we should just drop `requirements.txt` since nobody is maintaining it anymore, and rename the other one to `-arm64.txt`, and be done with it for now.I've never seen a Python project before with multiples of them before, though I have seen the distinction in more sophisticated setups e.g. Poetry.

- [ ] @twhetzel Decide: (a) drop `requirements.txt`, (b) manually add `curies` to this file, (c) something else.

twhetzel: The `curies` package is listed in this file. I missed that earlier.

PR #650, opened 2024-08-28
Add exclusion rules to mondo-excluded-values.yml

matentzn:
Resolves #ISSUE(s).

## Overview

This PR:
- One
- Two
- Three

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [ ] No

- [ ] No

- [ ] Yes

### DIFF
--- a/src/ontology/config/mondo-excluded-values.yml
+++ b/src/ontology/config/mondo-excluded-values.yml
@@ -1,4 +1,9 @@
 exclusions:
+  synonym_exclusion_rules:
+    - rule: ".*otherwise specified"
+      reason: "We don't include otherwise specified"
+    - rule: ".*, NOS"
+      reason: "We don't include otherwise specified"
   synonyms:
     - id: "MONDO:0008471" 
       scope: "oio:hasExactSynonym"


src/ontology/config/mondo-excluded-values.yml
```diff
@@ -1,4 +1,9 @@
 exclusions:
+  synonym_exclusion_rules:
+    - rule: ".*otherwise specified"
```
joeflack4, line 3: ## Regex-based lexical exclusion rules
I'm fine w/ this implementation, and with just adding rules incrementally as we go, starting with these.

src/ontology/config/mondo-excluded-values.yml
joeflack4, line 1: When would we like to start using this in the synonym sync?

joeflack4 (commented):
## How would we like to use these rules?
1. In cases of `-confirmed` or `-updated`, do we simply want to add the [`EXCLUDE` synonym type](https://mondo.readthedocs.io/en/latest/editors-guide/f-entities/)?
2. In `-added` cases, do we want to (a) add the synonym and add `EXCLUDE`? or (b) Not put these cases in the `-added` robot template, and instead output them into some file that shows excluded synonyms we are choosing not to add?
3. Is there more we'd like to do w/ these rules?

src/ontology/config/mondo-excluded-values.yml
joeflack4, line 1: When would we like to start using this in the synonym sync?

src/ontology/config/mondo-excluded-values.yml
joeflack4, line 3: ## Regex-based lexical exclusion rules
I'm fine w/ this implementation, and with just adding rules incrementally as we go, starting with these.

PR #671, opened 2024-09-25
Synonym sync: acronym case exception

joeflack4:
## Overview
Addressed an issue where we were preferring the Mondo capitalization even though it was incorrect, i.e. cases where the synonym was an acronym and capitalized in the source, but was not capitalized in Mondo.

### Documentation
- [ ] Yes

### QC
- [x] Yes
   
Build:
- #677

### New Packages
- [x] No

- [x] No

- [x] Yes

### Context
@twhetzel and I discussed this at our last 1:1. Sabrina had done some [recent curation](https://docs.google.com/spreadsheets/d/1y5jCmHN51ydBOcdqmGTBHjDeOHTPuYWd165qEYnNtIk/edit?gid=0#gid=0) and noticed that sometimes we would actually prefer the source's capitalization rather than Mondo's. I looked at the google sheet, at all of the values where `Use Source Case (Curator Review)`  == `Source`, and saw that these were all acronyms. They were all cases where the source was all caps and Mondo was not.

### Results
I ran a before/after, using DO as my test case, and looked at the differences in the outputs. There were differences in the `doid.synonyms.confirmed.robot.tsv` and `doid.synonyms.updated.robot.tsv`. I examined and the outputs are as I expected. Here are the diffs (FYI I added column headers at the top):
- [diffs.zip](https://github.com/user-attachments/files/17137387/diffs.zip)

### DIFF
--- a/src/scripts/sync_synonym.py
+++ b/src/scripts/sync_synonym.py
@@ -50,7 +50,7 @@ HEADERS_TO_ROBOT_SUBHEADERS = {
     'related_synonym_type': '>AI oboInOwl:hasSynonymType SPLIT=|',
 }
 SORT_COLS = ['case', 'mondo_id', 'source_id', 'synonym_scope_source', 'synonym_type', 'synonym_type_mondo', 'synonym']
-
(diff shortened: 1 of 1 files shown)


twhetzel (approved):
lgtm

PR #674, opened 2024-09-28
EMC: test build

joeflack4:
A build just to see if EMC in its current state will build correctly. Not meant to be merged.


joeflack4 (commented):
From skimming, LGTM.

Though:
- the main purpose of this build is to just check that `build-mondo-ingest` can run without error.
- I did not check the diffs in the EMC files

PR #628, opened 2024-08-05
Upgrade externally managed content

matentzn:
Release Update
- [x] Get Synonym Sync branch and Week 4 steps into `main`
- [ ] Update this branch to point to `main` and confirm build still runs

---
## Overview

This PR:
- Adds all missing pieces of externally managed content to Mondo Ingest (EFO, ClinGen, MedGen)
- And make goals to update them

### Documentation
Was the documentation added/updated under `docs/`?
- [x] Yes
- [ ] No, updates to the docs were not necessary after careful consideration

### QC
Was the full pipeline run before submitting this PR using `sh run.sh make build-mondo-ingest` on this branch (after 
`docker pull obolibrary/odkfull:dev`), and no errors occurred?
- [ ] Yes
- [ ] No, there are no functional (code-related) changes to the pipeline in the PR, so no re-run is necessary
   
### New Packages
Were any new *Python packages* added?
- [ ] Yes, [requirements.txt.full](https://github.com/INCATools/ontology-development-kit/blob/master/requirements.txt.full) was updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [X] No

Were any other *non-Python packages* added?
- [ ] Yes, [Dockerfile](https://github.com/INCATools/ontology-development-kit/blob/master/Dockerfile) updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [X] No

### PR Review and Conversations Resolved
Has the PR been sufficiently reviewed by at least 1 team member of the Mondo Technical team and all threads resolved?
- [ ] Yes

PR #679, opened 2024-10-13
EMC outputs update

joeflack4:
Updated outputs for:
- #628

Ran in: 13 min, 38sec

@twhetzel There's no harm I think in merging these updates, but we also don't have to. This is simply to show that I was able to successfully run this, after I made these changes:
- https://github.com/monarch-initiative/mondo/pull/8236

Though in order to test those changes, I had to make them locally, as that PR has not been merged yet.


twhetzel:
Closing since some other work is also now needed in the EMC branch.

PR #681, opened 2024-10-21
Remove GARD

joeflack4:
Resolves #399

## Overview
Removes semblance of GARD as a normal source. It's moving to EMC.

### Documentation
Was the documentation added/updated under `docs/`?
- [ ] Yes
- [x] No, updates to the docs were not necessary after careful consideration

### QC
Was the full pipeline run before submitting this PR using `sh run.sh make build-mondo-ingest` on this branch (after 
`docker pull obolibrary/odkfull:dev`), and no errors occurred?
- [x] Yes
- [ ] No, there are no functional (code-related) changes to the pipeline in the PR, so no re-run is necessary

Build PR:
- #682

### New Packages
Were any new *Python packages* added?
- [ ] Yes, [requirements.txt.full](https://github.com/INCATools/ontology-development-kit/blob/master/requirements.txt.full) was updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

Were any other *non-Python packages* added?
- [ ] Yes, [Dockerfile](https://github.com/INCATools/ontology-development-kit/blob/master/Dockerfile) updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

### PR Review and Conversations Resolved
Has the PR been sufficiently reviewed by at least 1 team member of the Mondo Technical team and all threads resolved?
- [ ] Yes

## Changes
Remove GARD
Remove GARD as non-EMC source
- ODK setup: metadata/gard.yml & mondo-ingest-odk.yaml
- Specific files: docs, outputs, and intermediaries.
- Refs within files: mondo-ingest.Makefile, gitignore, mkdocs.yml, Makefile (after running update_repo)

PR #682, opened 2024-10-21
DO NOT MERGE: Remove GARD - Build

joeflack4:
Build for:
- #681


src/ontology/lexmatch/split-mapping-set/mondo_closematch_omim.tsv
joeflack4, line 1: New file, FYI.

joeflack4 (commented):
LGTM!

Feels like a surprisingly low amount of files changed, but I suppose that is mostly due to the removal of GARD.

src/mappings/omim.sssom.tsv
```diff
@@ -217,7 +217,7 @@ MONDO:0007060		skos:exactMatch	OMIM:102530	spermatogenic failure 6	semapv:Unspec
 MONDO:0007061		skos:exactMatch	OMIM:102590		semapv:UnspecifiedMatching
 MONDO:0007062		skos:exactMatch	OMIM:102650	adactylia, unilateral	semapv:UnspecifiedMatching
 MONDO:0007064		skos:exactMatch	OMIM:102700	severe combined immunodeficiency, autosomal recessive, t cell-negative, B cell-negative, nk cell-negative, due to adenosine deaminase deficiency	semapv:UnspecifiedMatching
-MONDO:0007066		skos:exactMatch	OMIM:102800	adenosine triphosphatase deficiency, anemia due to	semapv:UnspecifiedMatching
+MONDO:0007066		skos:exactMatch	OMIM:102800		semapv:UnspecifiedMatching
```
twhetzel, line 220: Unrelated to this PR, however this OMIM entry, without a label, has been moved to a new entry, https://www.omim.org/entry/102800, so this should not show up as an exactMatch.

src/ontology/lexmatch/split-mapping-set/mondo_closematch_omim.tsv
twhetzel: Thanks for pointing this out. New or deleted files under `lexmatch/split-mapping-set` are not concerning since these `split-mapping-set` files are just subsets of the overall lexmatch files and some variance would be expected between runs. E.g. one run may have a new closematch entry that then creates a new file in `lexmatch/split-mapping-set`.

twhetzel (approved):
Thanks for running a feature build for this related code change. I think this looks ok.

PR #683, opened 2024-11-06
DO NOT MERGE: ODK 1.5.3 & Replace jinja2 with Jinjanator - Build

joeflack4:
Build for:
- #622


src/ontology/lexmatch/README.md
Looketh good to me.

Mostly just making sure this runs fine as well w/ the new version of ODK, which was also bumped in this PR.

joeflack4: @twhetzel I had in my notes it said I could merge when this passed, so merging!

PR #689, opened 2024-11-17
Synonym Sync - Build (post `develop` merge)

joeflack4:
Build for:
- #93

PR #693, opened 2024-11-19
Recreate dev branch and test synonym sync (consider merging 697 INSTEAD)

matentzn:
This PR reverts the last commit on `develop` and deconstructs it into smaller commits. 

@twhetzel @joeflack4 you can check check the commits 1 by 1 to see that I didn't make a mistake.

I have also done a complete test build which I will push to another branch.


tests/test_sync_subclassof.py
joeflack4, line 1: ## Test diffs
This file was deleted intentionally in the synonym sync, even though it is not synonym sync crelated, because I viewed synonym sync as being the new "first test PR". If you look at this old test file, it is an empty template. I would re-delete this file again, here and now or in a separate PR if you like.

src/ontology/config/mondo-excluded-values.yml
```diff
@@ -0,0 +1,2402 @@
+exclusions:
```
twhetzel, line 1: Good to see this file added back. It was removed when #93 was merged into `develop`. It can be found in the commit before that [here](https://github.com/monarch-initiative/mondo-ingest/blob/7d346079a929e72a8eecaa5117db7f5f72595b09/src/ontology/config/mondo-excluded-values.yml).

src/ontology/reports/sync-synonym/doid.synonyms.added.robot.tsv
joeflack4, line 1: ## Deletion of synonym sync outputs
I don't see a particular reason to delete these. They don't exist on `main`, so there should be no merge conflict issue to avoid.

However, perhaps you are trying to `develop` --> `merge` diff smaller and easier to read. I get the idea, but it seems mostly superfluous to me, as the review diff complexity is pushed elsewhere (i.e. here).

In any case, it made sense to review these outputs previously. But I think we've already done so, so if you wish to delete them, there's no problem either.

joeflack4: ## `.gitignore` some of these?
Actually, I feel like the `SOURCE.synonyms.added.robot.tsv`, etc, could be ignored, in favor of the combined one `sync-synonyms.added.robot.tsv`, etc, since those are all that `mondo` cares about.

It might be a good idea to do that to keep `mondo-ingest` smaller, and make for smaller easier to review diffs.

If we were to do that, it'd make sense to do so with other files too, e.g. subclass sync.

joeflack4 (commented):
## General review
I looked and outside of some outstanding issues, looks fine.

There's really not much to review here. Any file that wasn't synonym sync related, I ignored. I'll review those in #694 instead. That leaves only additions/deletions, of which there were only a handful.

I didn't observe any changes to the makefile or synonym sync Python code, which is good.

So if we follow my recommendations, the only changes of this PR will be:
1. Updating the non-synonym sync related outputs.
2. Deletion of synonym sync outputs.

Just FYI, I don't understand why there are 6 commits here. The process I would have followed would be something like:
1. `develop`: Check out the last commit before merging synonym sync. Force push overwrite this over current `develop`.
  - Can also back up `develop` before doing this.
2. Undelete `sync1-synonyms` branch (#93)
3. Check it out locally, do a `git log`, and check out the commit before merging build #689 into #93.
4. Push it as a new branch and make a new PR.
5. Sanity check / re-review that PR.
6. Merge it into `develop`.

Outstanding issues:
- Re-deletion of 1 or 2 files (see other threads)

src/ontology/lexmatch/README.md
I looked and outside of some outstanding issues, looks fine.

There's really not much to review here. Any file that wasn't synonym sync related, I ignored. I'll review those in #694 instead. That leaves only additions/deletions, of which there were only a handful.

I didn't observe any changes to the makefile or synonym sync Python code, which is good.

So if we follow my recommendations, the only changes of this PR will be:
1. Updating the non-synonym sync related outputs.
2. Deletion of synonym sync outputs.

Just FYI, I don't understand why there are 6 commits here. The process I would have followed would be something like:
1. `develop`: Check out the last commit before merging synonym sync. Force push overwrite this over current `develop`. Can also back up `develop` before doing this.
2. Undelete `sync1-synonyms` branch (#93)
3. Check it out locally, do a `git log`, and check out the commit before merging build #689 into #93.
4. Push it as a new branch and make a new PR.
5. Sanity check / re-review that PR.
6. Merge it into `develop`.

Outstanding issues:
- Re-deletion of 1 or 2 files (see other threads)

src/ontology/config/mondo-excluded-values.yml
twhetzel: If this file was re-named, what code needs to change to reflect this here?

joeflack4: Since Nico didn't make any code changes, the code should be hooked up to `mondo-exclusion-configs.yml`.

I mis-wrote earlier. I said that `mondo-exclusion-configs.yml` was the old file. I should've said: It's what `mondo-exclusion-values.yml` was renamed to a few months ago.

The code is currently hooked up to `mondo-exclusion-configs.yml`:
https://github.com/monarch-initiative/mondo-ingest/blob/eb1bf1eaf27dfb11095dfe5400101d5677599454/src/ontology/mondo-ingest.Makefile#L622-L627

twhetzel: Ok, I see [mondo-exclusion-configs.yml](https://github.com/monarch-initiative/mondo-ingest/blob/dev-new/src/ontology/config/mondo-exclusion-configs.yml) exists in `dev-new` so I agree, this file (`mondo-excluded-values.yml`) should be deleted.

src/ontology/reports/sync-synonym/doid.synonyms.added.robot.tsv
twhetzel: Interesting point, but we're just trying to get the state of `develop` back as it was so I don't think it's relevant for this PR.

src/ontology/lexmatch/README.md
twhetzel: Joe, what you proposed is also what I suggested so while I see value in how this PR was pulled apart and created, I am also not sure the build data was taken out of `sync1-synonyms` as part of the process in this PR.

twhetzel (commented):
I'm not sure I 100% follow the steps here since the build data needed to be pulled out from the `sync1-synonyms` branch as well as from `develop`. 

Other than that question, I agree with Joe that these files should be removed from this PR:
- src/ontology/config/mondo-excluded-values.yml
- tests/test_sync_subclassof.py

tests/test_sync_subclassof.py
matentzn: https://github.com/monarch-initiative/mondo-ingest/commit/2bc7a069bcf15cca37e6dc746584785d4a895e12

src/ontology/lexmatch/README.md
matentzn: > Just FYI, I don't understand why there are 6 commits here.

Why is the number of commits an issue? I have been taught that we should take better care of our commits. In fact, one of my most important collaborators only agreed to work with me if I adopted better commit practices.. I recommend reading https://incenp.org/dvlpt/docs/vcs-good-practices/index.html. 

This might _seem_ to contradict the SOP, but I think doing merge commits should probably only apply to small feature branches, not mega 6 months projects. 

That said, your process makes indeed more sense thinking of it. Thanks a ton for suggesting it!

1. Created backup branch: https://github.com/monarch-initiative/mondo-ingest/tree/develop-bk
2. Force removed the latest commit in develop: `git reset --hard HEAD~1 && git push --force` 
3. Undeleted `sync1-synonyms` branch, and checked out commit before data merge
4. Made PR: https://github.com/monarch-initiative/mondo-ingest/pull/697
5. Feel free to merge that one instead.

matentzn:
I now did what Joe suggested which is, despite the force push which we should probably use sparingly, a much better way of doing things! Thanks @joeflack4 for taking the time to writing this up, and sorry @twhetzel I did not think of this approach, I guess my mental model of git was not as strong as I thought!

src/ontology/lexmatch/README.md
twhetzel: To confirm, 
1. The `develop` branch in GitHub is now reverted back to before _any_ Synonym Sync content was pushed into this branch
2. The PR https://github.com/monarch-initiative/mondo-ingest/pull/697, once approved, can be merged into `develop` and this merge is not dependent on any other PRs being merged before this

@matentzn are both of these statements correct?

joeflack4: Thanks, Nico. Yeah I don't necessarily disagree w/ that style. When I said '6 commits' I meant just meant that I was confused why we were taking an approach unlike #697. But we changed the approach now. Thanks for all of this!

matentzn:
Closing in favour of #697

Issue #690, opened 2024-11-17
Fix tests: Synonym Sync


joeflack4:
## Overview
Due to some refactoring these tests are failing.

<details><summary>Details</summary>
<p>

```
/Users/joeflack4/virtualenvs/mondo-ingest/bin/python /Applications/PyCharm.app/Contents/plugins/python/helpers/pydev/[pydevd.py](http://pydevd.py/) --multiprocess --qt-support=auto --client 127.0.0.1 --port 55914 --file /Applications/PyCharm.app/Contents/plugins/python/helpers/pycharm/_jb_pytest_runner.py --path /Users/joeflack4/projects/mondo-ingest/tests/[test_sync_synonym.py](http://test_sync_synonym.py/) 
Testing started at 4:09 PM ...
Connected to pydev debugger (build 233.15325.20)
Launching pytest with arguments /Users/joeflack4/projects/mondo-ingest/tests/[test_sync_synonym.py](http://test_sync_synonym.py/) --no-header --no-summary -q in /Users/joeflack4/projects/mondo-ingest/tests

============================= test session starts ==============================
collecting ... collected 4 items

[test_sync_synonym.py](http://test_sync_synonym.py/)::TestSyncSynonyms::test_added 
[test_sync_synonym.py](http://test_sync_synonym.py/)::TestSyncSynonyms::test_confirmed 
[test_sync_synonym.py](http://test_sync_synonym.py/)::TestSyncSynonyms::test_deleted 
[test_sync_synonym.py](http://test_sync_synonym.py/)::TestSyncSynonyms::test_updated 

======================== 3 warnings, 4 errors in 3.44s =========================
ERROR                 [ 25%]
test setup failed
cls = <class 'test_sync_synonym.TestSyncSynonyms'>

    @classmethod
    def setUpClass(cls):
        """Set up"""
        # todo: #1 run pipeline / prereqs - dependent on fixing _run_make_goal()
        # if not os.path.exists(INPUT_SOURCE_DB):
        #     cls._run_make_goal('../../tests/input/sync_synonym/omim.db')
        # if not os.path.exists(INPUT_MONDO_DB):
        #     cls._run_make_goal('../../tests/input/sync_synonym/mondo.db')
        # if not os.path.exists(INPUT_MONDO_SYNONYMS):
        #     cls._run_make_goal('../../tests/input/sync_synonym/mondo-synonyms-scope-type-xref.tsv')
>       sync_synonyms(
[...]

Issue #687, opened 2024-11-15
Bug: Synonym Sync: `-confirmed` has non-exact entries


twhetzel:
There are two Mondo terms where the Synonym Sync pipeline adds source evidence for a confirmed synonym where the Synonym Sync pipeline should not. These Mondo terms from the qc check are: 
```
http://purl.obolibrary.org/obo/MONDO_0012817,Ewing sarcoma,NCIT:C7542,PNET of Thoracopulmonary region,MONDO:relatedTo
http://purl.obolibrary.org/obo/MONDO_0002514,hepatobiliary neoplasm,DOID:3117,hepatobiliary tumors,MONDO:mondoIsBroaderThanSource
```

These should be sorted out with high priority to resolve this issue, but it is not blocking the existing work on the Synonym Sync pipeline from being merged and getting into `main`. So #93 and #671 should be merged before working on this at this time.

joeflack4:
resolved by:
- #695

PR #696, opened 2024-11-20
Bugfix syn sync no exact mappings - outputs diff check

joeflack4:


src/ontology/reports/sync-synonym/doid.synonyms.confirmed.robot.tsv
```diff
@@ -2316,7 +2283,6 @@ MONDO:0002508	gingivitis	oio:hasExactSynonym		gingivitis			DOID:3087	gingivitis
 MONDO:0002512	papillary adenocarcinoma	oio:hasExactSynonym		papillary adenocarcinoma			DOID:3112	papillary adenocarcinoma					papillary adenocarcinoma	DOID:3112										
 MONDO:0002513	kidney benign neoplasm	oio:hasExactSynonym		kidney benign neoplasm			DOID:3116	kidney benign neoplasm					kidney benign neoplasm	DOID:3116										
 MONDO:0002513	kidney benign neoplasm	oio:hasExactSynonym		renal and ureteral tumor			DOID:3116	kidney benign neoplasm					renal and ureteral tumor	DOID:3116										
-MONDO:0002514	hepatobiliary neoplasm	oio:hasExactSynonym		hepatobiliary tumors			DOID:3117	hepatobiliary benign neoplasm					hepatobiliary tumors	DOID:3117
```
joeflack4, line 2319: ## Case 1 of 2 removed
As shown in the QC error in #687, this pair (`MONDO:0002514`, `DOID:3117`) had no exact mapping. Now it is removed.

src/ontology/reports/sync-synonym/ncit.synonyms.confirmed.robot.tsv
joeflack4, line 1: ## Case 2 of 2 removed
As shown in the QC error in #687, this pair (`MONDO: 0012817 `, `NCIT: C7542 `) had no exact mapping. Now it is removed.

The diff is too large to view, but you can examine the [raw file](https://raw.githubusercontent.com/monarch-initiative/mondo-ingest/ca0a63f4dae999ef2fd15321fd6b2b033ebe163f/src/ontology/reports/sync-synonym/ncit.synonyms.confirmed.robot.tsv) and search and see that this entry no longer appears.

joeflack4 (commented):
## General review
There are a lot more deletions than I'd hoped. Thoughts:

1. Wondering why the QC didn't find more.
2. More of these removals should be looked at further / face checked. I plan to do so.
3. Note: There are some very few `+` diffs (additions) here. I expected to only see `-` (removals). I attribute this to a slight delay in when the base synonym sync files were ran and when the new, post-code changes were ran--that there must have been a difference in inputs during that interim.
4. I suppose I didn't have a test case to check the `mondo.sssom.tsv` filtering. I could add that, but would need to fix #690 first.

Normally I keep PR comments low and tend towards discussing in detail at meetings with Trish, but I would like @matentzn's input as well on this one.

src/ontology/reports/sync-synonym/doid.synonyms.confirmed.robot.tsv
joeflack4, line 1: ## General review
There are a lot more deletions than I'd hoped. Thoughts:

1. Wondering why the QC didn't find more.
2. More of these removals should be looked at further / face checked. I plan to do so.
3. Note: There are some very few `+` diffs (additions) here. I expected to only see `-` (removals). I attribute this to a slight delay in when the base synonym sync files were ran and when the new, post-code changes were ran--that there must have been a difference in inputs during that interim.
4. I suppose I didn't have a test case to check the `mondo.sssom.tsv` filtering. I could add that, but would need to fix #690 first.

Normally I keep PR comments low and tend towards discussing in detail at meetings with Trish, but I would like @matentzn's input as well on this one.

matentzn: > Wondering why the QC didn't find more.

You say: "A lot more deletions than hoped" and then proceed to "Wondering why the QC didn't find more" - I think I need more context to help you with that.

> Note: There are some very few + diffs (additions) here. I expected to only see - (removals). I attribute this to a slight delay in when the base synonym sync files were ran and when the new, post-code changes were ran--that there must have been a difference in inputs during that interim.

Dont worry about the `+` too much, just spot checking reveals these are issues with non-deterministic ordering: https://github.com/monarch-initiative/mondo-ingest/pull/696/files#diff-0b42e55a06c0e95058d79c5f12692bcb454d84e3b1f59db2541b3ea7741aae50R42 (compare the changed line and the one above)


I think I need more context how I can help with this PR

joeflack4: ## QC results != this diff
Remaining sub-tasks:
- [ ] Determine: (a) If QC results should be different than this diff or (b) if QC is broken
- [ ] Review this diff & ensure changes are good: #695

Context: #687 has a copy/paste of the recent QC failure, where there are only 2 cases. But this diff has >4,000.

What the diff represents:
Code changes from #695. Just filtering out any cases where the Mondo & source terms do not have an exact mapping in `mondo.sssom.tsv`.

What the QC represents: 
@twhetzel @matentzn I thought the QC was checking the same thing, but I don't know where to go do look for the QC failure shown in #687. I think Nico said he ran it locally. Perhaps it is checking something fundamentally different?

Why Joe thinks #695 (& this diff) is correct:
I did an analysis ([google sheet](https://docs.google.com/spreadsheets/d/1QqhE19dk0gLVlsM0mcNJVLvkpetLOcfnFdehVFkQugQ/edit?gid=1707105518#gid=1707105518)), semi-randomly picking some of the cases removed. As expected, all cases removed do not have entries in `mondo.sssom.tsv` (`inSSSOM` column). So they shouldn't be in the synonym sync. The other columns are there to help me try and figure out why the QC might not have caught these 1000's of cases. But I couldn't figure it out. I'd have to look at the QC check itself.

joeflack4: The QC has different results because it is only checking cases in which the scope in `mondo-edit` doesn't match scope of `-confirmed`.
It's not checking that the entries have corollaries in `mondo.sssom.tsv`, which is what this PR is doing.

joeflack4:
Closing this as now redundant with:
- #707

PR #698, opened 2024-11-20
Rename OMIM properties

joeflack4:
## Overview
Update OMIM properties in properties.txt to the correct spelling / casing.

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [ ] No

- [ ] No

- [ ] Yes


src/ontology/config/properties.txt
```diff
@@ -15,8 +15,8 @@ http://www.geneontology.org/formats/oboInOwl#source
 http://www.geneontology.org/formats/oboInOwl#hasSynonymType
 http://www.geneontology.org/formats/oboInOwl#SynonymTypeProperty
 http://purl.obolibrary.org/obo/mondo#GENERATED
-http://purl.obolibrary.org/obo/mondo#omim_included
-http://purl.obolibrary.org/obo/mondo#omim_formerly
+http://purl.obolibrary.org/obo/mondo#includedEntryInOMIM
```
joeflack4, line 18: Do these need to actually appear in `properties.txt`?

During the tech meeting today, @matentzn was surprised I was calling them properties. But I feel like they belong here, alongside `http://purl.obolibrary.org/obo/mondo#GENERATED` which I think was disappearing if we didn't have it showing in this file.

I could do a before/after test to check this myself but for now I'll just pose the question.

matentzn: I dont think this needs to be here. Most of the things you see here, like "GENERATED" are synonym type properties.. That is very different from the `MONDO:includedEntryInOMIM` source annotations (see Mondo file)

joeflack4: Aren't the also supposed to be synonym type properties? Is there anything about the nature of `MONDO:GENERATED` that makes it more appropriate to say that it is a property of the synonym's type than the OMIM ones?

I remember that we have been discussing making object property declarations (I believe for these) in `omim.ttl`, but have not yet.

twhetzel: At some point things were failing without these being added, but this may not be the location that fixed things.

matentzn: Look at how `includedInOMIM` is used in Mondo:

```
xref: OMIM:184850 {source="MONDO:includedEntryInOMIM"}
```

As a `oboInOwl:source` annotation. This is not at all like:

```
synonym: "undetermined early-onset epileptic encephalopathy" EXACT GENERATED []
```

Which is how the GENERATED synonym type was used.

The confusion of why `includedEntryInOMIM` isn't used the same way, i.e.

```
synonym: "undetermined early-onset epileptic encephalopathy" EXACT includedEntryInOMIM []
```

That question is valid! I don't know the exact answer, but lets say for the sake of rapid progress now that the reason is historical.

joeflack4: Ahh I see, right.

Well I have to look at this more tomorrow to double check unless someone else wants to, but I think the way that `MONDO:omimFormerly` is being used might be different than `MONDO:includedEntryInOMIM`. I think it is being added directly onto the synonyms.

And shucks... I think maybe  `MONDO:includedEntryInOMIM` might be now as well in:
- https://github.com/monarch-initiative/omim/pull/168

I'll have to look into it tomorrow.

But for 'included', I think there were two things we were doing. One was a general thing on the OMIM class to say "this thing has included entries" (perhaps that's what `MONDO:includedEntryInOMIM` is supposed to be...), and then there are things directly added onto synonyms. Perhaps the latter was `MONDO:omim_included` and I was simply conflating the two.

matentzn: Yea maybe.. I dont think you really need the synonym type "omim included", but you may need the synonym type "omim formerly". Right now I think we should do whatever can be done right now and open issues to revise the decision if need be..

twhetzel: @joeflack4 have you looked further into this?

joeflack4: I had this on our [recent 1:1 agenda](https://docs.google.com/document/d/1zsfl5rkxdAhCk9xFX6oBDSa5UAjncBsloFuKU9Yt4Lo/edit?tab=t.0), section titled "_MONDO: omimFormerly & includedEntryInOMIM_".

Can wait until our next meeting if you like.

twhetzel: @joeflack4 can you sort out if this change is needed, try to run relevant make goals with and without the change to check and any other ways you may think of to test this.

joeflack4: @twhetzel May want to just move this to [our meeting](https://docs.google.com/document/d/1zsfl5rkxdAhCk9xFX6oBDSa5UAjncBsloFuKU9Yt4Lo/edit?tab=t.0), that might be easier. This is basically a sub issue of:
- #715 

And the answers to those questions / decisions will impact whether this is needed or not. If they will become synonym types as Nico seems to suggest might be appropriate, these will definitely be needed. If you / we decide that they should not be changed to that, then I can go ahead and run such an examination now.

joeflack4, line 18: Spoke with Nico today: [conclusions](https://docs.google.com/document/d/1zsfl5rkxdAhCk9xFX6oBDSa5UAjncBsloFuKU9Yt4Lo/edit?disco=AAABacqdWMQ).

- [ ] @joeflack4 Keep `includedEntryInOMIM`
- [ ] @joeflack4 Also keep `omimIncluded`

matentzn: No, includedEntryInOMIM has nothing to do here; it is not a property. It is a source code value. Yes. It should be kept in mondo. No, it should not be kept in this list.

joeflack4: Yes you're right. Not sure why I put that checkbox on this particular PR. But it does also need:
- [ ] @joeflack4 `omimFormerly`

joeflack4: Changes made, so re-requesting review.

matentzn (approved):

Issue #691, opened 2024-11-17
Bug: Empty `synonym_sync_combined_cases.robot.tsv`


joeflack4:
## Overview
For some reason this file is coming out blank at the end of running a build. And one time it had the text `beep` in it, which is Joe's way of playing a sound when a command has finished. Could be some strange piping issue.

## Choices
- [ ] 1. Decide whether or not to simply remove this
  - This is what becomes the google sheet; useful for curation. But not used in any pipeline.
- [ ] 2. If keep, fix.
  - Joe has a lot more notes in Workflowy. Crux is that the goal and its prereqs should be refactored. Could also change implementation to Python.

## Additional info
[Background](https://github.com/monarch-initiative/mondo-ingest/pull/689#discussion_r1845287083):
> This time it did not have the text `beep`. It is just empty, with 3 lines. I will interrogate the makefile, etc.
> 
> What's strange is that in a previous build when we got just `beep` in the file, I re-ran `sh run.sh make sync-synonyms` (I believe that's what I ran), and it created this file with no problem.
> 
> FYI, when I ran this build, I did as is always my protocol now: fresh clone + `sh run.sh make build-mondo-ingest -B`

joeflack4:
@twhetzel woops I forgot to close this and the other issues after the merge. Thanks!

PR #707, opened 2024-11-22
DO NOT MERGE: Bug fix: Synonym Sync: Entries without exact mappings - Build

joeflack4:
Build for:
- #695 

Though this mini build was probably adequate as well:
- #696


joeflack4 (commented):
LGTM!

- Have negative diffs resulting from changes in #695, and consistent with the preliminary mini-build #696.
  - Includes observed fixes (removals) for 2 cases outlined in #687
- Still seeing #691 for now
-  Large positive diff in OMIM genes files, as expected, consistent with those recent changes and other recent builds.
- Otherwise everything else looks same as usual; minimal diffs.

Issue #684, opened 2024-11-08
Bug: Synonym Sync: Duplicate rows


joeflack4:
## Overview
I found that there are duplicate rows entering into the `-confirmed` template, while examining Nico's `mondo` PR for the confirmed cases template:
- https://github.com/monarch-initiative/mondo/pull/8269

I do not know if this bug exists for the other ROBOT templates.

The bug does not have a negative consequence on processing; confirmed by checking the outputs for #8269.

The bug seems to manifest itself in that there can sometimes be multiple rows when there is a case difference between Mondo and the source. And it doesn't happen all of the time.

Example case:
|mondo_id|mondo_label|synonym_scope|synonym|source_id|synonym_case_diff_mondo|synonym_case_diff_source|
|-|-|-|-|-|-|-|
|MONDO:0000001|disease|oio:hasExactSynonym|disease|NCIT:C2991| | |
|MONDO:0000001|disease|oio:hasExactSynonym|disease|NCIT:C2991|disease|Disease|

PR #718, opened 2024-12-06
Data build files for Week 1 of Mondo Release Cycle

twhetzel:
Resolves #ISSUE(s).

## Overview

This PR:
Data build files for Week 1 of Mondo Release Cycle SOP

### Documentation
- [ ] Yes

### QC
- [x] Yes
   
### New Packages
- [x] No

- [x] No

- [ ] Yes


src/mappings/ordo.sssom.tsv
joeflack4, line 1: 2.2k lines changed is more than I'd expect.

Looking through it, it looks like they've updated the labels for a lot of terms.

src/ontology/external/mondo-medgen.robot.tsv
joeflack4, line 1: 324 lines is more than I'd expect to see here as well.

I don't have a lot to compare this too, though. I'm thought there must be another build that ran recently which must include this, but the only one I could find was #679, which had 54 lines changed.

I face checked and it doesn't look like it's a sorting issue; just a real diff.

src/ontology/reports/sync-synonym/icd11foundation.synonyms.updated.robot.tsv
joeflack4, line 1: ## `-updated` changes because of recent synonym sync (ok)
414 lines is more of a diff than I'd expect to see here as well.

It looks like the cause though is nothing concerning. It appears that a lot of Mondo evidence for synonyms changed (mostly drops) recently, as the result of us running recent synonym sync updates, and that is the cause of this diff.

Observed this as well for other sources, e.g. 2724 changes for NCIT.

src/ontology/reports/sync-synonym/ncit.synonyms.confirmed.robot.tsv
```diff
@@ -517,8 +518,8 @@ MONDO:0000929	balloon cell malignant melanoma	oio:hasExactSynonym		balloon cell
...
 MONDO:0000929	balloon cell malignant melanoma	oio:hasExactSynonym		balloon cell malignant skin melanoma	balloon cell malignant skin melanoma	Balloon Cell Malignant Skin Melanoma	NCIT:C4227	Balloon Cell Melanoma					balloon cell malignant skin melanoma	NCIT:C4227										
-MONDO:0000929	balloon cell malignant melanoma	oio:hasExactSynonym		balloon cell melanoma	balloon cell melanoma	Balloon Cell Melanoma	NCIT:C4227	Balloon Cell Melanoma					balloon cell melanoma	NCIT:C4227										
 MONDO:0000929	balloon cell malignant melanoma	oio:hasExactSynonym		balloon cell melanoma	balloon cell melanoma	Balloon cell melanoma	NCIT:C4227	Balloon Cell Melanoma					balloon cell melanoma	NCIT:C4227										
+MONDO:0000929	balloon cell malignant melanoma	oio:hasExactSynonym		balloon cell melanoma	balloon cell melanoma	Balloon Cell Melanoma	NCIT:C4227	Balloon Cell Melanoma					balloon cell melanoma	NCIT:C4227
```
joeflack4, line 522: ## Large diff because of duplicate rows
792 lines changed in the NCIT confirmed synonyms TSV. Seems to be artefact of #684.

I originally thought that the duplicate rows only appeared when there was a case difference between Mondo and the source. But this highlighted example shows a duplicate row where the rows are exactly the same.

src/ontology/lexmatch/README.md
joeflack4, line 1: ## Review summary
This build has more anomalies than I have typically noticed.

I think I've found the reasons for all of the ones I noticed, though and I don't think they are a concern.

joeflack4 (approved):

src/mappings/ordo.sssom.tsv
matentzn: INteresting, yes! Not a blocker. WE dont use ORDO mappings ver much, other than for lexmatch

src/ontology/external/mondo-medgen.robot.tsv
matentzn: No concerns, the file is large, 300 changes are ok. (Many are additions)!

matentzn (approved):
LGTM - a lot of diff, and next time (after EMC has run in Mondo) we should make sure that the diff is considerably smaller. A lot of diff though because of updated ORDO which is fine.

PR #727, opened 2024-12-18
Bug fixes: `synonym_sync_combined_cases.robot.tsv` & build hangs

### LINKED ISSUE #691: Bug: Empty `synonym_sync_combined_cases.robot.tsv`

joeflack4:
Resolves #691

## Overview
Fixes that sometimes `synonym_sync_combined_cases.robot.tsv` would be empty.
Also fixes: Sometimes builds would pause in a particular place until some input was made in the terminal.

## Additional info
**Cause:**
I was just about to start working on #691 when I asked Claude about why my builds were pausing; turns out these are related!

> ```
> for file in ; do \
>   tail -n +3 $file >> reports/sync-synonym/synonym_sync_combined_cases.robot.tsv; \
> done
> ```
> There's no list of files after for `file in -` it's empty. This means the loop will execute zero times, but it still needs to be processed by the shell.
> The `head -n 2` command appears to be dangling without an input:
> `head -n 2  > reports/sync-synonym/synonym_sync_combined_cases.robot.tsv`
> It's waiting for input from stdin, which would cause it to pause until it receives input (or you press Ctrl+D to signal EOF).

This explains why this happened to me usually, Nico sometimes, and Trish never. This doesn't happen unless you run a build on a fresh clone!

**Additional tweaks made**:
The "combined cases" file has become a permanent curation file. It made sense to formalize it as such in this PR. Formalizing it this way also made the bug fix easier to implement.

### Documentation
- [ ] Yes

### QC
- [x] Yes
   
Build:
- #730

### New Packages
- [x] No

- [x] No

- [ ] Yes

### DIFF
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -594,11 +594,9 @@ tmp/%-synonyms-scope-type-xref.tsv: $(COMPONENTSDIR)/%.owl
 ../../tests/input/sync_synonym/%-synonyms-scope-type-xref.tsv:
 	$(ROBOT) query -i ../../tests/input/sync_synonym/test_$*.owl --query ../sparql/synonyms-scope-type-xref.sparql $@
 
-# todo: we may remove this output later output for analysis during development; at the end, remove it and its usages
-INPUT_FILES := $(wildcard tmp/synonym_sync_combined_cases_*.tsv)
-$(SYN_SYNC_DIR)/synonym_sync_combined_cases.robot.tsv: $(foreach n,$(ALL_COMPONENT_IDS), $(SYN_SYNC_DIR)/$(n)-synonyms.added.robot.tsv)
-	head -n 2 $(firstword $(INPUT_FILES)) > $@
-	for file in $(INPUT_FILES); do \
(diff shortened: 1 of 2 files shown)


joeflack4:
@twhetzel @matentzn Just a reminder to review. It will be very helpful for me when this is merged.

twhetzel (approved):
Glad you sorted out why the last file was always empty.

PR #747, opened 2025-01-12
Bug fix: Synonym sync exclusions

joeflack4:
- Addresses failing QC error `qc-related-exact-synonym.sparql` in https://github.com/monarch-initiative/mondo/pull/8463

## Overview
The exclusions were only excluding on matching scope. Now, they will exclude any instance of a synonym for a given `mondo_id`, regardless of scope, so long as they are in the exclusion config.

### Documentation
- [ ] Yes

### QC
- [x] Yes #758

Mini build:
- #748

### New Packages
- [x] No

- [x] No

- [x] Yes

### DIFF
--- a/src/scripts/sync_synonym.py
+++ b/src/scripts/sync_synonym.py
@@ -171,7 +171,7 @@ def _common_operations(
     """
     # Filter exclusions
     if len(mondo_exclusions_df) > 0:
-        df = _filter_a_by_not_in_b(df, mondo_exclusions_df, ['mondo_id', 'synonym_scope', 'synonym_join'])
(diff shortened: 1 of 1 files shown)


joeflack4:
See: results from applying on `mondo-edit.obo`: 
- https://github.com/monarch-initiative/mondo/pull/8562

twhetzel (approved):
The results of this code change look good in https://github.com/monarch-initiative/mondo/pull/8562 and this is correct that any synonym in Mondo with the `EXCLUDE` annotation should be excluded in the Synonym Sync regardless of the synonym scope designation between what's in Mondo and the annotation in the external source ontology.

PR #748, opened 2025-01-12
DO NOT MERGE: Bug fix: Synonym sync exclusions - mini build

joeflack4:
Mini build for:
- #747


joeflack4:
I intend mainly to do a repeat of https://github.com/monarch-initiative/mondo/pull/8463 using this branch, rather than examining these files manually, and see if the QC error for `qc-related-exact-synonym.sparql` now goes away.

PR #751, opened 2025-01-15
Address `qc-duplicate-exact-synonym-no-abbrev` failures in `mondo`

joeflack4:
Partially addresses issues in:
- https://github.com/monarch-initiative/mondo/pull/8463

## Overview
Filters cases where there are multiple Mondo terms with the same exactSynonym, and instead puts them into a special curation TSV.

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
Mini build:
- #752 

Build:
- #755

### New Packages
- [x] No

- [x] No

- [ ] Yes

## Additional info
[Google sheet](https://docs.google.com/spreadsheets/d/14u0sEPtRcHBzvJ-ipunxdcO10tlb622i7Z80LDgCsrs/edit?gid=1161223846#gid=1161223846)

Mondo PR using this branch:
- https://github.com/monarch-initiative/mondo/pull/8584

### DIFF
--- a/src/scripts/sync_synonym_curation_filtering.py
+++ b/src/scripts/sync_synonym_curation_filtering.py
@@ -63,18 +63,6 @@ def sync_synonyms_curation_filtering(
 
     # Filter cases from -added & -updated
     df_filtered = df[~df.index.isin(df_review.index)]
-
-    # todo temp - analysis, if needed
-    ar1 = df_review[df_review['case'] == 'unconfirmed']
-    # examine
-    f_df = pd.read_csv('~/Desktop/failures.csv').rename(columns={'value': 'synonym'})
-    il = f_df.merge(df_filtered, on=['synonym'], how='left')  # 1329  todo: why some not have 'case'? makes sense if they are not from syn sync and pre-existed in Mondo but that does not appear to be case because 0 fails b4
-    ii = f_df.merge(df_filtered, on=['synonym'], how='inner')  # 358
-    # for 1 class i looked at, it's in review and not in the sync
-    ar = df_review = df_review[df_review['synonym'] == 'Zlotogora-Ogur syndrome']
-    af = df_review = df_filtered[df_filtered['synonym'] == 'Zlotogora-Ogur syndrome']
-
-    # todo temp: don't run these lines until i'm done w/ code and ready to push again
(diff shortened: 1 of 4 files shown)


src/scripts/sync_synonym_curation_filtering.py
```diff
@@ -0,0 +1,148 @@
...
+    df['case'] = case
+    return df
+
+
+def sync_synonyms_curation_filtering(
+    added_path: Union[Path, str], confirmed_path: Union[Path, str], updated_path: Union[Path, str],
+    mondo_synonyms_path: Union[Path, str], mondo_db_path: Union[Path, str], outpath: Union[Path, str]
+):
+    """Filter out cases where curation is needed"""
+    # todo temp: backup temporarily for development; remove when done
```
twhetzel, line 38: Is this `todo` comment still relevant or should be removed now?

joeflack4: Hmm, good question. I'll remove the todo as it will be obsoleted this change I have which is next on my todo list:
- Make it so that the first round of outputs from the synonym sync go to `tmp/`. Then use those as the inputs for this script, and the outputs of it go into `reports/`. This eliminates the need for a backup, as the files will no longer be mutated. I'll remove the todo.

FYI that change doesn't have to hold up this PR. I can do it after.

joeflack4: Yeah, the original idea was to mutate these files. The new idea came up later, but ideally I should have gotten to it by now.

It sounds like you are ready to approve this PR? I just wanna ask because I see no other remaining threads, but there is no approval yet.

twhetzel (approved):
While I am not entirely convinced an additional script was needed vs. incorporating this as a requirement for the synonym sync script, I am approving since the data generated in https://github.com/monarch-initiative/mondo/pull/8584 looks generally as expected and the `mondo-edit.obo` file passes all QC steps. Nice job getting these items all sorted!

joeflack4:
Thanks!

If there's an idea of how to easily refactor this, let me know. It needs to come after the normal synonym sync, because the synonym sync, like the other pipeline, runs on each source individually. But in order to resolve this issue, we need to look at all sources in aggregate, hence the separate script.

PR #752, opened 2025-01-15
DO NOT MERGE: Address `qc-duplicate-exact-synonym-no-abbrev` failures in `mondo` - mini build

joeflack4:
Mini build for:
- #751


joeflack4 (commented):
LGTM!

New review file is as I would expect. Changes as I would expected, and confirmed working as attended via QC check in `mondo`:
- https://github.com/monarch-initiative/mondo/pull/8584

PR #753, opened 2025-01-16
Update doid_exclusions.tsv

nicolevasilevsky:
Resolves #ISSUE(s).

address https://github.com/monarch-initiative/mondo/issues/8455

## Overview

This PR:
- adds DOID:0050706 to the exclusion list

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [x] No

- [x] No

- [ ] Yes


matentzn (approved):

twhetzel:
@matentzn why did you merge this into `main`?

joeflack4:
I think Nicole opened it on `main`, which is fine in this case since it's a config, but we should get this change into `develop` too. I forget if there's something in the SOP about that, like merging `main` into `develop` immediately after.

twhetzel:
@joeflack4 for this case, there was nothing specifically in the Mondo Ingest Development and Build Cycle SOP document. I just added a `TODO` for this. When other files are updated in the Mondo "Week 2" cycle of the Mondo SOP, there are requests for new builds in mondo-ingest `main`. Does an update to an exclusion file also then need a new build? 

Also, the new `TODO` is that the developer who merges this kind of PR here should also update `develop`.

PR #759, opened 2025-01-17
Remove unused synonym sync files

joeflack4:
## Overview
Removes unused synonym sync file: (i) the combined cases file, (ii) now the source-specific files so that they now go in `tmp/` instead of being committed.

### Documentation
Was the documentation added/updated under `docs/`?
- [ ] Yes
- [x] No, updates to the docs were not necessary after careful consideration

### QC
Was the full pipeline run before submitting this PR using `sh run.sh make build-mondo-ingest` on this branch (after 
`docker pull obolibrary/odkfull:dev`), and no errors occurred?
- [ ] Yes #760
- [ ] No, there are no functional (code-related) changes to the pipeline in the PR, so no re-run is necessary
   
### New Packages
Were any new *Python packages* added?
- [ ] Yes, [requirements.txt.full](https://github.com/INCATools/ontology-development-kit/blob/master/requirements.txt.full) was updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

Were any other *non-Python packages* added?
- [ ] Yes, [Dockerfile](https://github.com/INCATools/ontology-development-kit/blob/master/Dockerfile) updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

### PR Review and Conversations Resolved
Has the PR been sufficiently reviewed by at least 1 team member of the Mondo Technical team and all threads resolved?
- [ ] Yes

PR #760, opened 2025-01-17
DO NOT MERGE: Remove unused synonym sync files - build

joeflack4:
Build for:
- #759


src/ontology/reports/sync-synonym/sync-synonyms.added.robot.tsv
joeflack4, line 1: Confirming it is working in that there are no longer synonym sync files by source, and no more big "everything concat'ed" file.

I did expect to see updates to there `-updated` and `-confirmed` TSVs, but I guess there were no updates to those since I last ran this, just `-added`.

twhetzel: Why would this file have all the content removed and there not be any `-confirmed` file?

Also, why is there an empty `temp` file in this PR?

twhetzel (changes requested):
Comments inline

src/ontology/reports/sync-synonym/sync-synonyms.added.robot.tsv
joeflack4: There are is a `-confirmed` file. But it remains unchanged in this build. That's why it's not showing up in the diff.

`temp` is not supposed to be there. Deleted!

twhetzel:
I am closing this PR until issues with https://github.com/monarch-initiative/mondo-ingest/pull/759 are sorted out.

PR #764, opened 2025-01-31
Fix ORDO mappings query

matentzn:
ORDO mappings were broken for a while, this commit makes the translation of ORDO xref annotations into SKOS much simpler, and hopefully a little more resilient.

Resolves issue reported by Chris on slack, that the mappings in ORDO looked off.

## Overview

This PR:
- Reduces the size of the query by focusing only on the beginning of the source metadata string provided by ORDO. This ensures that lexical variations that inevitably trickle in are not ignored in the future.
- Replaces the wrong ECO property which has already been mapped to oio:source

One super weird thing I could not get to the bottom of is that I have to run the query in a separate `query` pipe for it to work. If I put it in the first half, it handles half the transformations, the second time the rest. I have not more time to look into this - since it works I am fine with it.

Example output:

```
<!-- http://www.orpha.net/ORDO/Orphanet_100000 -->

    <owl:Class rdf:about="http://www.orpha.net/ORDO/Orphanet_100000">
        <rdfs:subClassOf rdf:resource="http://www.orpha.net/ORDO/Orphanet_100002"/>
        <rdfs:subClassOf rdf:resource="http://www.orpha.net/ORDO/Orphanet_557494"/>
        <oboInOwl:hasDbXref>ICD-10:D36.1</oboInOwl:hasDbXref>
        <oboInOwl:hasDbXref>ICD-11:2F3Y</oboInOwl:hasDbXref>
        <oboInOwl:hasDbXref>ICD10:D36.1</oboInOwl:hasDbXref>
        <oboInOwl:hasDbXref>ICD11:2F3Y</oboInOwl:hasDbXref>
        <oboInOwl:hasDbXref>UMLS:C5681810</oboInOwl:hasDbXref>
        <oboInOwl:hasExactSynonym>Reticular perineurioma</oboInOwl:hasExactSynonym>
        <oboInOwl:inSubset rdf:resource="http://purl.obolibrary.org/obo/mondo#ordo_disorder"/>
        <oboInOwl:inSubset rdf:resource="http://purl.obolibrary.org/obo/mondo#ordo_group_of_disorders"/>
        <oboInOwl:inSubset rdf:resource="http://purl.obolibrary.org/obo/mondo#ordo_subtype_of_a_disorder"/>
        <rdfs:label>Reticular perineurioma</rdfs:label>
        <skos:broadMatch>ICD-11:2F3Y</skos:broadMatch>
        <skos:exactMatch>UMLS:C5681810</skos:exactMatch>
        <skos:narrowMatch>ICD-10:D36.1</skos:narrowMatch>
    </owl:Class>
```

### Documentation
- [ ] Yes

### QC
- [x] Yes
   
Build PRs:
- #765 
- #766

### New Packages
- [X] No

- [X] No

- [ ] Yes

### DIFF
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -92,6 +92,7 @@ $(COMPONENTSDIR)/ordo.owl: $(TMPDIR)/ordo_relevant_signature.txt config/properti
 			--update ../sparql/ordo-construct-subsets.ru \
 			--update ../sparql/ordo-construct-d2g.ru \
 		query --update ../sparql/ordo-replace-annotation-based-mappings.ru \
+		remove --term oboInOwl:hasDbXref --axioms "annotation" \
 		remove -T $(TMPDIR)/ordo_relevant_signature.txt --select complement --select "classes individuals" --trim false \
 		remove -T config/properties.txt --select complement --select properties --trim true \
 		annotate --ontology-iri $(URIBASE)/mondo/sources/ordo.owl --version-iri $(URIBASE)/mondo/sources/$(TODAY)/ordo.owl -o $@; fi
--- a/src/ontology/mondo-ingest.Makefile
+++ b/src/ontology/mondo-ingest.Makefile
@@ -91,7 +91,7 @@ $(COMPONENTSDIR)/ordo.owl: $(TMPDIR)/ordo_relevant_signature.txt config/properti
 			--update ../sparql/ordo-construct-subclass-from-part-of.ru \
 			--update ../sparql/ordo-construct-subsets.ru \
 			--update ../sparql/ordo-construct-d2g.ru \
-			--update ../sparql/ordo-replace-annotation-based-mappings.ru \
+		query --update ../sparql/ordo-replace-annotation-based-mappings.ru \
 		remove -T $(TMPDIR)/ordo_relevant_signature.txt --select complement --select "classes individuals" --trim false \
 		remove -T config/properties.txt --select complement --select properties --trim true \
 		annotate --ontology-iri $(URIBASE)/mondo/sources/ordo.owl --version-iri $(URIBASE)/mondo/sources/$(TODAY)/ordo.owl -o $@; fi
--- a/src/sparql/ordo-replace-annotation-based-mappings.ru
+++ b/src/sparql/ordo-replace-annotation-based-mappings.ru
@@ -9,7 +9,7 @@ prefix sssom: <https://w3id.org/sssom/>
 
 DELETE {
   ?xref_anno a owl:Axiom ;
-    ECO:0000218 ?mapping_precision_string .
+    oboInOwl:source ?mapping_precision_string .
 } INSERT {
   ?cls a owl:Class;
       skos:broadMatch ?value .
@@ -23,7 +23,7 @@ DELETE {
     owl:annotatedSource ?cls ;
     owl:annotatedProperty ?mapping ;
     owl:annotatedTarget ?value ;
-    ECO:0000218 ?mapping_precision_string .
+    oboInOwl:source ?mapping_precision_string .
 
   FILTER(
     STRSTARTS(STR(?mapping_precision_string), "- NTBT") || 
@@ -34,7 +34,7 @@ DELETE {
 
 DELETE {
   ?xref_anno a owl:Axiom ;
-    ECO:0000218 ?mapping_precision_string .
+    oboInOwl:source ?mapping_precision_string .
 } INSERT {
   ?cls a owl:Class;
       skos:exactMatch ?value .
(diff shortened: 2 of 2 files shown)


matentzn:
@twhetzel made a start for this issue, needs a data release to test and also just someone sanity checking. Thanks!

twhetzel:
Sure, I will review and run a feature branch data build as part of the review process to merge this into `develop`. Thanks Nico!

src/sparql/ordo-replace-annotation-based-mappings.ru
joeflack4, line 1: ## General Review
I remember pausing like most of the way through this refactor a while ago, IDR why. I imagine `ECO:0000218` worked for us in the past and was later mapped to `oio:source`, but IDR either for sure.

Needing to run the query twice is indeed strange.

I did notice that you have extra patterns for narrow/broad for the extra `-`, e.g.:

```
    STRSTARTS(STR(?mapping_precision_string), "BTNT") || 
    STRSTARTS(STR(?mapping_precision_string), "- BTNT")
```

But not for exact:
```
  FILTER(STRSTARTS(STR(?mapping_precision_string), "E (Exact "))
```

@matentzn **(1)** I think this needs to be changed to account for the `-`, yeah?

`cat tmp/component-download-ordo.owl.owl | grep "E (Exact"`:
```
        <obo:ECO_0000218 xml:lang="en">E (Exact mapping: the two concepts are equivalent)</obo:ECO_0000218>
        <obo:ECO_0000218 xml:lang="en">- E (Exact mapping: the two concepts are equivalent).
        <obo:ECO_0000218 xml:lang="en">- E (Exact mapping: the two concepts are equivalent).
        <obo:ECO_0000218 xml:lang="en">E (Exact mapping: the two concepts are equivalent)</obo:ECO_0000218>
        ...
```

**(2)** @matentzn @twhetzel Should we remove [the existing releases](https://github.com/monarch-initiative/mondo-ingest/releases) for now, assuming none of the outputs are used for anything? I saw that Chris had downloaded from there, and we're not maintaining them right now.
Alternatively it wouldn't be too much trouble as well to tweak the action to simply upload files to the release after we have already ran locally and pushed.

**(3)** I'm not reviewing these outputs because Trish is already running a data build. But I'm wondering, why are there 97 files changed if supposedly a build was not run?

**(4)** I don't think there are any outputs that would show up in the diff where we can verify the effects of this SPARQL query change, right?
For now, I ran it and checked locally and verified that `components/ordo.owl` does now have the `skos` predicates as I would expect.

joeflack4: @matentzn Holding off on approval until you check on **(1)**. I can also look into it.

joeflack4: (1) Ah, OK. But you see them now in my example above yes?

(2) I think you are right! It was probably me. I changed it back. OK. We will need to remember to merge `main` into `develop` after all this is done. Or we could change it back to `develop` again; probably better. I leave it up to you or Trish.

(4) Ah OK. I'll take a look. BTW I think Trish already did a build for this:
- #765

matentzn: (1) ohh thanks for the spot.

https://github.com/monarch-initiative/mondo-ingest/commit/ed0f1dff218f12c090d152bde5e210a2725e084e

twhetzel: Nice catch Joe! I verified the ontology content and query and all cases appear to be handled now.
```
grep '<obo:ECO_0000218 xml:lang="en">' component-download-ordo.owl.owl | awk -F'[>()]' '{print $2}' | sort -u 

- BTNT 
- E 
- ND 
- NTBT 
BTNT 
E 
ND 
NTBT 
```

twhetzel: Initially Nico asked that I do the data build, which was [here](https://github.com/monarch-initiative/mondo-ingest/pull/766), but then with new query changes he ran a [new build](https://github.com/monarch-initiative/mondo-ingest/pull/766).

src/ontology/mondo-ingest.Makefile
```diff
@@ -92,6 +92,7 @@ $(COMPONENTSDIR)/ordo.owl: $(TMPDIR)/ordo_relevant_signature.txt config/properti
 			--update ../sparql/ordo-construct-subsets.ru \
 			--update ../sparql/ordo-construct-d2g.ru \
 		query --update ../sparql/ordo-replace-annotation-based-mappings.ru \
+		remove --term oboInOwl:hasDbXref --axioms "annotation" \
```
twhetzel, line 95: @matentzn are you running a build with this update?

twhetzel: final data build [PR](https://github.com/monarch-initiative/mondo-ingest/pull/768)

twhetzel:
@matentzn since the original branch was pulled from `main` I changed the base branch back to `main` vs. `develop`.

matentzn:
Thx

twhetzel:
Once merged into `main`, can you kindly merge `main` back into `develop` to get us squared away again with the SOP?

matentzn:
Feel free to do it as well if you don't want to wait - when you approve, just merge it and update develop..

twhetzel:
Ok, I'll take it from here!

twhetzel (approved):
LGTM

PR #777, opened 2025-02-10
Bug fix: Synonym Sync: Sorting of case variation columns

joeflack4:
## Overview
Fixed an issue where the |-delimited synonym_case_mondo and synonym_case_source columns in synonym sync robot templates were not deterministically sorted.

### Documentation
Was the documentation added/updated under `docs/`?
- [ ] Yes
- [x] No, updates to the docs were not necessary after careful consideration

### QC
Was the full pipeline run before submitting this PR using `sh run.sh make build-mondo-ingest` on this branch (after 
`docker pull obolibrary/odkfull:dev`), and no errors occurred?
- [x] Yes
- [ ] No, there are no functional (code-related) changes to the pipeline in the PR, so no re-run is necessary
   
Mini build:
- #778 

Build:
- #779

### New Packages
Were any new *Python packages* added?
- [ ] Yes, [requirements.txt.full](https://github.com/INCATools/ontology-development-kit/blob/master/requirements.txt.full) was updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

Were any other *non-Python packages* added?
- [ ] Yes, [Dockerfile](https://github.com/INCATools/ontology-development-kit/blob/master/Dockerfile) updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [x] No

### PR Review and Conversations Resolved
Has the PR been sufficiently reviewed by at least 1 team member of the Mondo Technical team and all threads resolved?
- [ ] Yes

PR #778, opened 2025-02-10
DO NOT MERGE: Bug fix: Synonym Sync: Sorting of case variation columns - mini build

joeflack4:
Mini build for:
- #777


joeflack4:
@twhetzel I'm relieved to see that if you check the diff for this, you can see the results of this fix.

Issue #762, opened 2025-01-22
Synonym Sync - better handling of conflicts between OMIM formerly and DOID abbreviations


twhetzel:
Based on the Mondo `../sparql/qc/general/qc-related-exact-synonym.sparql` qc check errors [here](https://github.com/monarch-initiative/mondo/pull/8562#discussion_r1912361769) from running the Synonym Sync, the Synonym Sync (added case) should filter out any DOID synonym abbreviations with the scope exact if that same synonym matches an OMIM synonym abbreviation that is also tagged as formerly and has the scope related. In these cases, the synonym abbreviation scope of the OMIM formerly synonym abbreviation as related is correct.

joeflack4:
@twhetzel Hmm, this makes sense.

I'm going to hold off on this though, because `mondo-ingest` currently has no way of knowing which synonyms are "formerly" until we merge these:
- https://github.com/monarch-initiative/omim/pull/132
- https://github.com/monarch-initiative/omim/pull/167

twhetzel:
Sure, this issue does now raise the priority of those other OMIM ingest issues.

PR #783, opened 2025-02-18
Synonym Sync: DO `-added` filtering

joeflack4:
## Overview
Now only allowing exact synonyms which were created by Mondo as derivatives of the term's label.

### Documentation
- [ ] Yes

### QC
- [x] Yes #788
   
Mini Build:
- ~#784~
- #786

### New Packages
- [x] No

- [x] No

- [ ] Yes


src/ontology/mondo-ingest.Makefile
```diff
@@ -611,8 +611,9 @@ $(SYN_SYNC_DIR)/synonym_sync_combined_cases.robot.tsv: $(foreach n,$(ALL_COMPONE
 		tail -n +3 $$file >> $@; \
 	done
 
+# DO filtration: For now we are filtering out all DOID terms from sync-synonyms.added.robot.tsv
 $(SYN_SYNC_DIR)/sync-synonyms.added.robot.tsv: $(foreach n,$(ALL_COMPONENT_IDS), $(SYN_SYNC_DIR)/$(n)-synonyms.added.robot.tsv)
-	awk '(NR == 1) || (NR == 2) || (FNR > 2)' $(SYN_SYNC_DIR)/*.synonyms.added.robot.tsv > $@
+	awk '(NR == 1) || (NR == 2) || (FNR > 2)' $(filter-out $(SYN_SYNC_DIR)/doid.synonyms.added.robot.tsv,$(wildcard $(SYN_SYNC_DIR)/*.synonyms.added.robot.tsv)) > $@
```
matentzn, line 616: Where is the related issue for this? I am not sure we should filter out _all_ synonyms. For example exactSynonyms which are basically just the rdfs:label of the sources are very valuable content for Mondo - these are safe.

twhetzel: @matentzn I do not see this an as issue to hold up this PR. Do you?

sabrinatoro: - "exact" synonyms coming from DO are often not "exact", and include the names of more specific diseases that have been "included" in the DO term (this was discovered during alignment work). Therefore, we cannot "trust" that exact synonyms in DO are actually exact. (ie what Trish said)
- we did not discuss labels. I would agree with Nico that "DO label" should be safe and should be added as exact synonyms.

I want to clarify a few things that I mentioned in the past, and I don't know where it fits in the synchronization pipeline (and where it was documented). 
- if there is an exact synonym from another source, and DO has the same exact synonym, we should add the DO id to this synonym
- **What we do not want** is to add an exact synonym from DO when DO is the only source for this exact synonym.

twhetzel: Thanks Sabrina! Yes, that is a good point from Nico that since the Synonym Sync pipeline is treating the label of a term from the external source as an exact synonym for that term itself those synonyms are useful to maintain and should be safe to add into Mondo as exact synonyms. 

Apologies @joeflack4 as this changes the scope of the feature request. Is this updated request clear? I gather this will require some changes to the initial way DO is handled by the synonym sync pipeline vs. what was a smaller update here currently.

matentzn (changes requested):
We added an annotation on exact synonyms generated from the label:

https://github.com/monarch-initiative/mondo-ingest/issues/762

That should be used to filter out DO generated exact matches (anything which does not have the new tag/synonym type on it)

src/ontology/mondo-ingest.Makefile
joeflack4: Yep, per discussion [on Slack](https://monarchinitiative.slack.com/archives/G01JACCEDAB/p1739915259129209), this is clear.

Just wanted to respond to [Nico](https://github.com/monarch-initiative/mondo-ingest/pull/783#pullrequestreview-2627554925), I believe the annotation you're referring to is `MONDO:GENERATED`, but I don't think it has anything to do with #762.

twhetzel: @joeflack4 yes, Nico and I had a meeting and chatted about the "DOID" new requirement and concocted this solution https://github.com/monarch-initiative/mondo-ingest/pull/785. Ideally I would like to run a full build (and plan to do so), but #785 is also a small enough/discrete change that I am comfortable to merge in #785 before the full build is done. I do want to run one make goal as a sanity check first and will do that now.

src/sparql/exact_syn_from_label.ru
```diff
@@ -1,10 +1,21 @@
...
   ?cls oboInOwl:hasExactSynonym ?label .
+	
+  	[ 	rdf:type owl:Axiom ;
+				owl:annotatedSource ?cls ;
+           owl:annotatedProperty oboInOwl:hasExactSynonym ;
+           owl:annotatedTarget ?label ;
+        oboInOwl:hasSynonymType <http://purl.obolibrary.org/obo/mondo#GENERATED_FROM_LABEL> ].
 } WHERE {
   ?cls rdfs:label ?label .
   FILTER NOT EXISTS { ?cls owl:deprecated ?deprecated . }
```
twhetzel, line 21: FYI - there some minor formatting changes between what you have here in this PR and this file in it's final merged [PR here](https://github.com/monarch-initiative/mondo-ingest/pull/785/files) after I removed the "fix" for brackets in #785. Apologies for that.

src/ontology/mondo-ingest.Makefile
twhetzel: @joeflack4 with the change to now include DOID synonyms that are from us converting the term label to an exact synonym tagged with `MONDO:GENERATED_FROM_LABEL` some of these may actually be new to Mondo and be in the "added" category so now the `doid.synonyms.added.robot.tsv` file should not be excluded as I understand. Do you have plans to change or close this PR now?

joeflack4: Yes, this file should not be excluded. It should've been in draft mode. It will be updated shortly.

src/ontology/mondo-ingest.Makefile
```diff
@@ -631,7 +631,8 @@ $(SYN_SYNC_DIR)/%-synonyms.added.robot.tsv $(SYN_SYNC_DIR)/%-synonyms.confirmed.
 	--outpath-added $(SYN_SYNC_DIR)/$*.synonyms.added.robot.tsv \
 	--outpath-confirmed $(SYN_SYNC_DIR)/$*.synonyms.confirmed.robot.tsv \
 	--outpath-updated $(SYN_SYNC_DIR)/$*.synonyms.updated.robot.tsv \
-   	--outpath-combined $(TMPDIR)/synonym_sync_combined_cases_$*.tsv
+   	--outpath-combined $(TMPDIR)/synonym_sync_combined_cases_$*.tsv \
```
twhetzel, line 634: Adding this line `--outpath-combined $(TMPDIR)/synonym_sync_combined_cases_$*.tsv` back in looks like an accident.

src/scripts/sync_synonym.py
```diff
@@ -247,7 +252,7 @@ def sync_synonyms(
     mondo_exclusion_configs: Union[Path, str], onto_synonym_types_path: Union[Path, str],
     mondo_mappings_path: Union[Path, str], onto_config_path: Union[Path, str], outpath_added: Union[Path, str],
     outpath_confirmed: Union[Path, str], outpath_updated: Union[Path, str],
-    outpath_combined: Union[Path, str], outpath_deleted: Union[Path, str] = None,
+    outpath_combined: Union[Path, str], outpath_deleted: Union[Path, str] = None, doid_added_filtration=False
```
twhetzel, line 255: This line `outpath_combined: Union[Path, str], outpath_deleted: Union[Path, str] = None, doid_added_filtration=False` also looks like an accident.

twhetzel:
@joeflack4 can you add the command line you used to create the DO files for the Synonym Sync (https://github.com/monarch-initiative/mondo-ingest/pull/786)?

src/scripts/sync_synonym.py
```diff
@@ -481,6 +489,10 @@ def cli():
     parser.add_argument(
         '-b', '--outpath-combined', required=True,
         help='Path to curation file which is a concatenation of all cases.')
+    parser.add_argument(
+        '-D', '--doid-added-filtration', required=False, action='store_true', help='If this flag is '
```
twhetzel, line 493: @joeflack4 where is `doid-added-filtration` being set? I would also suggest this be `True` by default.

twhetzel (changes requested):
Some additional lines have carried into this PR that were removed in earlier PRs.

The main issue is seeing and testing how `doid_added_filtration` is being used.

src/ontology/mondo-ingest.Makefile
joeflack4: Oh no, what that is is the addition of a ` \` at the end of that line.

The combined files are being removed in the PR I'm working on next week.

src/scripts/sync_synonym.py
joeflack4: This one is also a new addition at the end of the line: `, doid_added_filtration=False`

src/scripts/sync_synonym.py
joeflack4: It's present in the make goal, appearing after [this line](https://github.com/monarch-initiative/mondo-ingest/pull/783#discussion_r1964742353). I thought that defaulting to `False` would be good so that in the future if we want to disable this, one easy way is to drop the usage of that flag in the make goal.

I don't personally like the default being True because I think that by default, if one comes to the synonym sync without prior knowledge, or if we come back to it in the future after having forgotten this detail, one would naturally expect that it is doing the same operations on every source, without special treatment of any source. That's pretty much how all the other pipelines work, and is good design I think. So the presence of the flag also serves as self-documentation.

After you have considered my argument here, if you feel the pros of changing it outweigh the cons, let me know and I'll change it.

twhetzel: That link is not pointing to a code line for me.

src/ontology/mondo-ingest.Makefile
twhetzel: ok, i did not realize that change was only in the other PR

src/ontology/mondo-ingest.Makefile
joeflack4, line 1: > @joeflack4 can you add the command line you used to create the DO files for the Synonym Sync (https://github.com/monarch-initiative/mondo-ingest/pull/786)?

The way I did it was:
1) In the PyCharm debugger, I already had a runtime config set up for running the synonym sync specifically for DO. I ran that. It generated an `-added`, `-updated`, and `-confirmed` file. I committed just the `-added.`

The config looks like this:
```
--mondo-mappings-path
tmp/mondo.sssom.tsv
--ontology-db-path
components/doid.db
--mondo-synonyms-path
tmp/mondo-synonyms-scope-type-xref.tsv
--onto-config-path
metadata/doid.yml
--outpath-added
reports/sync-synonym/doid.synonyms.added.robot.tsv
--outpath-confirmed
reports/sync-synonym/doid.synonyms.confirmed.robot.tsv
--outpath-updated
reports/sync-synonym/doid.synonyms.updated.robot.tsv
--mondo-exclusion-configs
config/mondo-exclusion-configs.yml
--onto-synonym-types-path
tmp/doid-synonyms-scope-type-xref.tsv
--outpath-combined
tmp/synonym_sync_combined_cases_doid.tsv
--doid-added-filtration
```

If you wanted to run it via the terminal, you'd have to put this in the front `python3 ../scripts/sync_synonym.py`.

You could also run `make reports/sync-synonym/doid-synonyms.added.robot.tsv`, but the prereqs trigger and it takes a long time to run those. Since I already ran the prereqs in the "mini build base", it wasn't necessary for me to run it this way.

2) Run this goal:
```makefile
$(SYN_SYNC_DIR)/sync-synonyms.added.robot.tsv: $(foreach n,$(ALL_COMPONENT_IDS), $(SYN_SYNC_DIR)/$(n)-synonyms.added.robot.tsv)
	awk '(NR == 1) || (NR == 2) || (FNR > 2)' $(SYN_SYNC_DIR)/*.synonyms.added.robot.tsv > $@
```
But before doing that, I temporarily dropped  the `$(foreach n,$(ALL_COMPONENT_IDS), $(SYN_SYNC_DIR)/$(n)-synonyms.added.robot.tsv)` prereq. Otherwise it would have run the synonym sync for all files. And then that would also trigger the prereqs for the synonym sync, which I didn't want to do a second time as mentioned above.

---

As we spoke about at the meeting, this was done in this particular way this time because running the full synonym sync results in garbled files due to a bug that's fixed in #767 but not yet ready.

@twhetzel Another way to do it simply yourself is by running just `sh run.sh make reports/sync-synonym/sync-synonyms.added.robot.tsv`, and this will run all the prereqs as well which I don't think you will mind doing.

src/scripts/sync_synonym.py
twhetzel: ok, i did not realize that removing the combined file was only in https://github.com/monarch-initiative/mondo-ingest/pull/767

src/scripts/sync_synonym.py
joeflack4: The link is pointing to my comment regarding the line that you highlighted; that's what I mean.

I see you're update. Thanks for your review! I thought you might be curious about the setup.

src/ontology/mondo-ingest.Makefile
twhetzel: Thanks, I wanted to be able to repeat how you ran it. I was able to successfully run the synonym sync pipeline itself.

twhetzel (approved):
LGTM - thanks for getting this done!

src/scripts/sync_synonym.py
```diff
@@ -404,6 +409,9 @@ def sync_synonyms(
     # - leave only synonyms that don't exist on given Mondo IDs
     added_df = _filter_a_by_not_in_b(source_df_with_mondo_ids, mondo_df, ['mondo_id', 'synonym_join'])
     added_df = added_df[added_df[['mondo_id', 'source_id']].apply(tuple, axis=1).isin(mapping_pairs_set)]
+    if doid_added_filtration and source_name.lower() == 'doid':
+        added_df = added_df[added_df['synonym_type'] == 'http://purl.obolibrary.org/obo/mondo#GENERATED_FROM_LABEL']
```
matentzn, line 413: This is exactly how I was hoping we do it!

matentzn (approved):
Exactly how I thought we should do it! Nice. 

Completely irrelevant side bar: I have come to profer option syntax `--doid-added-filtration true` over flag (or argument) -style syntax (`--doid-added-filtration`) in most complex python programs for two reasons:

1. The option stays in the makefile even if it is switched off, giving some level of documentation to the option. If it is removed, it will disappear from consciousness (i.e. no one will ever look at the python scripts)
2. There may be a nee for more fine grained control than true/false in the future, like `--doid-added-filtration exact-from-label,related,narrow`, and in option syntax, we can more easily build that without having to change dependent code.

Now this was more of a personal opinion piece, and I dont think you need to concern youself with it here as the script only has one dependent (the makefile) and the option will most likely not change.

src/scripts/sync_synonym.py
joeflack4: Glad to hear! You make good points on the alternative setup for the flag. I won't change it for now, but I'll add a `todo`.

PR #784, opened 2025-02-18
DO NOT MERGE: Filter DO from sync-synonyms.added.robot.tsv - mini build

joeflack4:
Mini build for:
- #783


twhetzel (commented):
I don't think this PR is valid now based on the changes in #783 and comment [here](https://github.com/monarch-initiative/mondo-ingest/pull/783/files#r1964100481)

PR #785, opened 2025-02-19
Update exact_syn_from_label.ru

matentzn:
Resolves #ISSUE(s).

## Overview

This PR:
- One
- Two
- Three

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [ ] No

- [ ] No

- [ ] Yes

### DIFF
--- a/src/sparql/exact_syn_from_label.ru
+++ b/src/sparql/exact_syn_from_label.ru
@@ -4,7 +4,17 @@ PREFIX owl: <http://www.w3.org/2002/07/owl#>
 PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
 
 INSERT {
+  <http://purl.obolibrary.org/obo/mondo#GENERATED_FROM_LABEL> rdf:type owl:AnnotationProperty .
+  <http://www.geneontology.org/formats/oboInOwl#hasSynonymType> rdf:type owl:AnnotationProperty .
+  <http://purl.obolibrary.org/obo/mondo#GENERATED_FROM_LABEL> rdfs:subPropertyOf <http://www.geneontology.org/formats/oboInOwl#SynonymTypeProperty>  .
+	
   ?cls oboInOwl:hasExactSynonym ?label .
+	
+  	[ 	rdf:type owl:Axiom ;
+				owl:annotatedSource ?cls ;
+           owl:annotatedProperty oboInOwl:hasExactSynonym ;
+           owl:annotatedTarget ?label ;
+        oboInOwl:hasSynonymType <http://purl.obolibrary.org/obo/mondo#GENERATED_FROM_LABEL> ].
 } WHERE {
   ?cls rdfs:label ?label .
   FILTER NOT EXISTS { ?cls owl:deprecated ?deprecated . }


twhetzel (changes requested):
a test of the components make goal, e.g. `sh run.sh make components/doid.owl -B` now works following Joe's update, but I am not seeing the `GENERATED_FROM_LABEL` annotation.

@joeflack4 let me know if you are looking into this or I should take this. thanks

src/sparql/exact_syn_from_label.ru
```diff
@@ -1,6 +1,7 @@
 # Create an exact synonym copy for all labels
 PREFIX oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
 PREFIX owl: <http://www.w3.org/2002/07/owl#>
+PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
```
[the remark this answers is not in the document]
joeflack4: @twhetzel Wow you got onto that fast. Yes, I avoid them quite often these days. I did that last one within 1 minute, sorry you encountered it.

joeflack4 (commented):
x

src/sparql/exact_syn_from_label.ru
joeflack4, line 1: ## General review
Good thinking w/ the new annotation prop, Nico.

I ran this locally and it works fine.

@twhetzel I think we should merge this first so that we can use it in #783, without running a build. Or, probably even better, I'll just effectively branch #783 off of this, and we can skip the build for this one and just do the build on #783 instead.

FYI changed the base to `develop`.

I also pushed 2 new commits to finish some issues with this (missing rdf namespace & properties.txt entry).

@matentzn How do you intend this to work alongside the `MONDO:GENERATED` created by https://github.com/monarch-initiative/mondo-ingest/blob/main/src/sparql/fix-labels-with-brackets.ru? I suppose that is a bad file name, because it doesn't fix labels. It creates synonyms from labels and other fields. And it also will create them whether or not there are brackets. It will just fix/remove brackets if they exist. Here's how the props are coming out now:


```owl
<owl:Class rdf:about="http://purl.obolibrary.org/obo/DOID_7086">
    <rdfs:label xml:lang="en">multicentric papillary thyroid carcinoma</rdfs:label>
</owl:Class>
<owl:Axiom>
    <owl:annotatedSource rdf:resource="http://purl.obolibrary.org/obo/DOID_7086"/>
    <owl:annotatedProperty rdf:resource="http://www.geneontology.org/formats/oboInOwl#hasExactSynonym"/>
    <owl:annotatedTarget xml:lang="en">multicentric papillary thyroid carcinoma</owl:annotatedTarget>
    <oboInOwl:hasSynonymType rdf:resource="http://purl.obolibrary.org/obo/mondo#GENERATED_FROM_LABEL"/>
</owl:Axiom>
<owl:Axiom>
    <owl:annotatedSource rdf:resource="http://purl.obolibrary.org/obo/DOID_7086"/>
    <owl:annotatedProperty rdf:resource="http://www.geneontology.org/formats/oboInOwl#hasExactSynonym"/>
    <owl:annotatedTarget>multifocal papillary thyroid carcinoma</owl:annotatedTarget>
    <oboInOwl:hasSynonymType rdf:resource="http://purl.obolibrary.org/obo/mondo#GENERATED"/>
</owl:Axiom>
```

joeflack4 (approved):

src/sparql/exact_syn_from_label.ru
twhetzel: Thanks for pushing the "properties.txt" file, I am now getting the `#GENERATED_FROM_LABEL` tags.

twhetzel: @joeflack4 what is your concern with the other query, `[fix-labels-with-brackets.ru](https://github.com/monarch-initiative/mondo-ingest/blob/main/src/sparql/fix-labels-with-brackets.ru)` and the query in this PR?

twhetzel: Based on your comment, one concern is that this query is not using the already "fixed" label when the original label contains brackets. For example, for http://id.who.int/icd/entity/186800184 'Metastatic nasopharyngeal carcinoma [primary nasopharyngeal carcinoma spreading elsewhere]' in ICD11 here's what happens after running `sh run.sh make components/icd11foundation.owl -B` on this branch
Screenshot 2025-02-19 at 1 58 49 PM


And in these cases it would be better that the `#GENERATED_FROM_LABEL` exact synonym is created from the `#GENERATED` exact synonym. I am running a test to check another option for the query in this PR....

joeflack4: Yeah, my concern was that adding `#GENERATED` as well feels a little redundant when there's already `#GENERATED_FROM_LABEL`. Though it's not a big gripe. At this current time, I'm not actually opposed to completing the work quickly as opposed to getting it perfect stylistically and such.

But @twhetzel you bring up a good point, that the `#GENERATED_FROM_LABEL` would be better done after the brackets fixing.

joeflack4: @twhetzel You asked what my other general gripes are with [fix-labels-with-brackets.ru](https://github.com/monarch-initiative/mondo-ingest/blob/main/src/sparql/fix-labels-with-brackets.ru). That is explained in [this comment](https://github.com/monarch-initiative/mondo-ingest/pull/785#discussion_r1962379284); not within scope of this issue.

src/sparql/exact_syn_from_label.ru
```diff
@@ -1,11 +1,29 @@
...
+    owl:annotatedSource ?cls ;
+    owl:annotatedProperty oboInOwl:hasExactSynonym ;
+    owl:annotatedTarget ?cleanLabel ;
+    oboInOwl:hasSynonymType <http://purl.obolibrary.org/obo/mondo#GENERATED_FROM_LABEL>
+  ] .
+} 
+WHERE {
   ?cls rdfs:label ?label .
+
+  BIND(REPLACE(STR(?label), " [\\[(].*[)\\]]$", "") AS ?cleanLabel)
```
twhetzel, line 26: I tested this update with ICD11 and here's an example IRI, http://id.who.int/icd/entity/1945176477, where it's using the "cleaned" label to convert to the exact synonym tagged as `GENERATED_FROM_LABEL`. Tested by running: `sh run.sh make components/icd11foundation.owl -B`

twhetzel: @matentzn are you ok with this addition to the query or do you prefer the original label be used as the `GENERATED_FROM_LABEL` exact synonym before any attempts to "fix" the label due to the presence of brackets? 

One question I have is how well this "fix" for labels was previously tested and reviewed to confirm it's working as expected in all cases so that we do not add badly formatted synonyms into Mondo.

matentzn: I am not sure. This fix was needed for ICD10, and was not really tested with anything else. I am not sure this is a good idea to make default for _all_ ontologies. I would do this only for ICD11 and ICD10. There are some legitimate reasons of having brackets in class names..

matentzn: Maybe I would just leave the unchanged label for now.. for safety reasons

twhetzel: Sure, while I would like to have more meaningful synonyms, ie without brackets etc., if this bracket "fix" was not more deeply tested then we run the risk of adding even worse formatted synonyms into Mondo which is a big negative. I will revert this change, but this small though exercise has been useful.

src/sparql/exact_syn_from_label.ru
twhetzel: FYI - the `#GENERATED_FROM_LABEL` is used since the "fix" brackets query works on both labels and synonyms so this new tag makes it explicit where this new synonym is from as well as the parallel discussion in this PR about that the "fix" for brackets should not be applied to labels being used as synonyms at this time.

twhetzel (approved):
lgtm

twhetzel:
For this specific PR, I am ok with merging this into `develop` before doing a full data build so that Joe can move forward with  the other issues to work on. I will do a feature branch data build, however as a sanity check.

PR #798, opened 2025-02-25
DO NOT MERGE: Synonym Sync: `qc-duplicate-exact-synonym-no-abbrev` related updates && Remove unused files - mini build 5

joeflack4:
Mini build for:
- #767


src/scripts/sync_synonym_curation_filtering.py
```diff
@@ -150,9 +150,6 @@ def cli():
     parser.add_argument(
         '-A', '--added-outpath', required=True,
         help='Path to filtered ROBOT template TSV containing synonyms that aren\'t yet integrated into Mondo.')
-    parser.add_argument(
-        '-C', '--confirmed-outpath', required=True,
```
joeflack4, line 154: Just FYI this change is erroneously showing up in this diff.

The change hadn't been pushed to the base branch when I opened this PR. But I pushed it now, yet this still shows up in the diff when it should not.

Context: I removed this param from `sync_synonym_curation_filtering.py` since `-confirmed` is not mutated by this file. It was only here as a way to save files to `reports/sync-synonym/` when they were previously saved to `tmp/`, but I changed the way that this is set up in the makefile to make more sense.

src/ontology/reports/sync-synonym/review-qc-duplicate-exact-synonym-no-abbrev.tsv
joeflack4, line 1: Good that there is a negative diff. Less things are being filtered here now.

joeflack4, line 1: ## General review
`git diff --numstat qc-duplicate-exact-synonym-no-abbrev--updates`

|added|removed|filename                                                           |
|-----|-------|-------------------------------------------------------------------|
|352  |3358   |src/ontology/reports/sync-synonym/review-qc-duplicate-exact-synonym-no-abbrev.tsv|
|33962|29727  |src/ontology/reports/sync-synonym/sync-synonyms.added.robot.tsv    |
|5646 |5646   |src/ontology/reports/sync-synonym/sync-synonyms.confirmed.robot.tsv|
|20783|20779  |src/ontology/reports/sync-synonym/sync-synonyms.updated.robot.tsv  |

Number of rows for `-confirmed` unchanged. The `-updated` difference is nominal. Likely just because sources have changed in the past 1-2 weeks since these files were last committed.

~3,000 rows removed from the review file. This is good, reflecting the fact that less collisions are now being filtered. Namely, `-added` is no longer being impacted by collisions with things coming in from `-updated`. This is reflected by ~3,000 more rows in added, as they are no longer being filtered.

I won't get hung up on the exact numbers here. The indeterminism of these outputs makes the diffs a bit hard to review. And the filtering from the review file was previously bugged, based on indexes rather than common primary keys.

The most important thing is to run the QC check in Mondo off of this branch and make sure that it passes.

joeflack4: Confirming some more things were fixed today.

More detail is in the [OP of the main PR](https://github.com/monarch-initiative/mondo-ingest/pull/767#issue-2837086226), as a series of checkboxes at the bottom, which link to the commit which addressed the issue, as well as link to comment where it was brought up. Some of the updates were:

- `MONDO:GENERATED_FROM_LABEL`: Confirmed that this is showing up appropriately in the new `synonym_type_internal` column, and not in the `synonym_type` column.
- Some things erroneously marked unconfirmed when they were really confirmed.
- Filtering `-updated` from the review template: The `-updated` file itself was not being filtered. But some cases were still appearing in the review file. They were only cases where the synonym was the same as the term's label.

twhetzel: I am re-running the synonym sync from https://github.com/monarch-initiative/mondo-ingest/pull/767 to confirm whether the files still have the issue with `GENERATED_FROM_LABEL` still in the ROBOT column to determine if this is a code issue or somehow the wrong files were pushed here.

twhetzel: In both the "-confirmed" and "-added" files the value `http://purl.obolibrary.org/obo/mondo#GENERATED_FROM_LABEL` is still in the ROBOT column, e.g. exact_synonym_type >AI oboInOwl:hasSynonymType SPLIT=|.

joeflack4: Fixed:
Remove MONDO:GENERATED_FROM_LABEL from ALL synonym_type columns, not just the one plainly called 'synonym_type'. This type will now correctly only appear in the synonym_type_internal column. (4c1ed57e0b3ed81032ef53f9514cb8093e366ec9)

twhetzel: Looks good!

twhetzel (approved):
The "added" and "confirmed" data files look good. I did not review the "review" file.

There are still a handful of various Mondo QC checks, but they are within manageable levels for a curator to fix vs. the previous case of >1600 instances of errors from the `qc-duplicate-exact-synonym-no-abbrev` check.

NOTE: There are still some data issues that will need to be addressed before these files can be used to update Mondo for a release. These will be in a new ticket.

PR #813, opened 2025-03-27
SubClass Sync: Exclusions

joeflack4:
Resolves #ISSUE(s).

## Overview

This PR:
- One
- Two
- Three

### Documentation
Was the documentation added/updated under `docs/`?
- [ ] Yes
- [ ] No, updates to the docs were not necessary after careful consideration

### QC
Was the full pipeline run before submitting this PR using `sh run.sh make build-mondo-ingest` on this branch (after 
`docker pull obolibrary/odkfull:dev`), and no errors occurred?
- [ ] Yes
- [ ] No, there are no functional (code-related) changes to the pipeline in the PR, so no re-run is necessary
   
### New Packages
Were any new *Python packages* added?
- [ ] Yes, [requirements.txt.full](https://github.com/INCATools/ontology-development-kit/blob/master/requirements.txt.full) was updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [ ] No

Were any other *non-Python packages* added?
- [ ] Yes, [Dockerfile](https://github.com/INCATools/ontology-development-kit/blob/master/Dockerfile) updated and an [ODK issue](https://github.com/INCATools/ontology-development-kit/issues/new) submitted
- [ ] No

### PR Review and Conversations Resolved
Has the PR been sufficiently reviewed by at least 1 team member of the Mondo Technical team and all threads resolved?
- [ ] Yes

PR #816, opened 2025-03-28
DO NOT MERGE: SubClass Sync: Exclusions - build

joeflack4:
Build for:
- #813


src/ontology/lexmatch/README.md
The main thing to check in this PR is that these changes worked as intended:
1. Filters excluded subclass axioms
2. Filters axioms that include any susceptibility terms

However, due to time, I'll opt to defer to others to check for this unless otherwise requested.

So instead is a standard face checking of the outputs. And they LGTM!

twhetzel: Can you point out some examples where the excluded subsclasses are now excluded?

src/ontology/reports/doid.subclass.added.robot.tsv
```diff
@@ -27,23 +27,18 @@ MONDO:0000297	baylisascariasis	MONDO:0005560	DOID:0050259	DOID:936	brain disorde
 MONDO:0000298	dioctophymiasis	MONDO:0005093	DOID:0050260	DOID:37	skin disorder
 MONDO:0000298	dioctophymiasis	MONDO:0005240	DOID:0050260	DOID:557	kidney disorder
 MONDO:0000301	ophthalmomyiasis	MONDO:0005328	DOID:0050268	DOID:5614	eye disorder
-MONDO:0000302	basidiobolomycosis	MONDO:0000255	DOID:0050278	DOID:0050135	subcutaneous mycosis
```
joeflack4, line 30: # Example: 1. Filters excluded subclass axioms
This subclass relationship (SCR) is now being filtered out. If we check `mondo-edit.obo`, we can see that the SCR has `excluded_subClassOf`.
```
id: MONDO:0000302
name: basidiobolomycosis
relationship: excluded_subClassOf MONDO:0000255 {source="DOID:0050278", source="https://orcid.org/0000-0001-5208-3432"} ! subcutaneous mycosis
```

src/ontology/lexmatch/README.md
joeflack4: @twhetzel [Here's one](https://github.com/monarch-initiative/mondo-ingest/pull/816#discussion_r2038302209). I'll see if I can point out some more.

It looks like you are not asking for me to show an example of (2), which is good for me, given my available time.
I was going to show an example of that anyway, but it looks like that is much more difficult to find. Too look for such cases where those are now excluded, method 1 is that I could look through the diff, removal by removal, and check if it was removed because of susceptibility, rather than `excluded_subClassOf`. But it could be that I have to go through dozens or hundreds of cases before I find one. Method 2, I could go through `mondo-susceptibility-terms.tsv` (generated from the new SPARQL query), and for each Mondo term shown, look at `mondo-edit.obo`, then make sure that, for a given source ontology, it has an exact match, as well as an SCR (and is not marked `excluded_subClassOf`), where both the objects of the match and SCR are of that same source ontology, then check if that SCR appears as a removal in the diff. I just checked a few cases, but I didn't find a qualifying case. It could also be that I would have to go through dozens or hundreds of entries in `mondo-susceptibility-terms.tsv` before I found a qualifying case. So the efficient way to show that this is indeed having an effect is to write some queries, but I'm going to opt not to do that.

src/ontology/reports/icd10cm.subclass.added.robot.tsv
```diff
@@ -15,11 +15,8 @@ MONDO:0006932	pulmonary edema	MONDO:0015925	ICD10CM:J81	ICD10CM:J80-J84	intersti
 MONDO:0009010	aortic arch interruption	MONDO:0001863	ICD10CM:Q25.21	ICD10CM:Q25.2	aorta atresia
 MONDO:0009061	cystic fibrosis	MONDO:0005066	ICD10CM:E84	ICD10CM:E70-E88	metabolic disease
 MONDO:0009169	endocardial fibroelastosis	MONDO:0004994	ICD10CM:I42.4	ICD10CM:I42	cardiomyopathy
-MONDO:0010888	adenomyosis	MONDO:0005133	ICD10CM:N80.0	ICD10CM:N80	endometriosis
```
joeflack4, line 18: # Example of "1. Filters excluded subclass axioms"
This subclass relationship (SCR) is now being filtered out. If we check `mondo-edit.obo`, we can see that the SCR has `excluded_subClassOf`.
```
id: MONDO:0010888
name: adenomyosis
xref: ICD10CM:N80.0 {source="MONDO:equivalentTo", source="DOID:288"}
relationship: excluded_subClassOf MONDO:0005133 {source="DOID:288", source="ICD10CM:N80.0", source="https://orcid.org/0000-0001-5208-3432"} ! endometriosis
```

src/ontology/reports/ordo.subclass.added.robot.tsv
```diff
@@ -3,51 +3,33 @@ ID		AI obo:mondo#excluded_subClassOf	>A oboInOwl:source
 MONDO:0000044	hereditary hypophosphatemic rickets	MONDO:0015962	Orphanet:437	Orphanet:183592	inherited renal tubular disease
 MONDO:0000044	hereditary hypophosphatemic rickets	MONDO:0017322	Orphanet:437	Orphanet:289098	disorders of vitamin D metabolism
 MONDO:0000087	polymicrogyria	MONDO:0015159	Orphanet:35981	Orphanet:102283	multiple congenital anomalies/dysmorphic syndrome-intellectual disability
-MONDO:0000141	mosaic variegated aneuploidy syndrome	MONDO:0015159	Orphanet:1052	Orphanet:102283	multiple congenital anomalies/dysmorphic syndrome-intellectual disability
-MONDO:0000171	muscular dystrophy-dystroglycanopathy, type A	MONDO:0016156	Orphanet:899	Orphanet:207119	qualitative or quantitative defects of FKRP
-MONDO:0000171	muscular dystrophy-dystroglycanopathy, type A	MONDO:0016184	Orphanet:899	Orphanet:209030	qualitative or quantitative defects of protein O-mannosyltransferase 1
-MONDO:0000171	muscular dystrophy-dystroglycanopathy, type A	MONDO:0016185	Orphanet:899	Orphanet:209033	qualitative or quantitative defects of protein O-mannosyltransferase 2
```
joeflack4, line 9: # Example of "1. Filters excluded subclass axioms"
In this example, multiple SCRs were removed.
```
id: MONDO:0000171
name: muscular dystrophy-dystroglycanopathy, type A
xref: Orphanet:899 {source="MONDO:equivalentTo"}
relationship: excluded_subClassOf MONDO:0016156 {source="Orphanet:899", source="https://orcid.org/0000-0001-5208-3432"} ! qualitative or quantitative defects of FKRP
relationship: excluded_subClassOf MONDO:0016184 {source="Orphanet:899", source="https://orcid.org/0000-0001-5208-3432"} ! qualitative or quantitative defects of protein O-mannosyltransferase 1
relationship: excluded_subClassOf MONDO:0016185 {source="Orphanet:899", source="https://orcid.org/0000-0001-5208-3432"} ! qualitative or quantitative defects of protein O-mannosyltransferase 2
```

Interesting case also. This is a case where Mondo is more granular, as multiple Mondo terms here are all mapping to `Orphanet:899`.

src/ontology/reports/doid.subclass.added.robot.tsv
twhetzel: @joeflack4 since MONDO:0000302 basidiobolomycosis has two excluded subclassOf annotations, why is only one removed from this file?

twhetzel: Since there is another example you commented on for ICD10CM where multiple cases of exclude subclassOf are removed for a Mondo class from the file, perhaps there was only one exclude subclassOf annotation for this Mondo class in the mondo-edit.obo version used.

twhetzel (approved):
Thanks for pointing out the examples. Looks good.

twhetzel:
For the Subclass Sync, is there not a combined file for "added"?

src/ontology/reports/doid.subclass.added.robot.tsv
joeflack4: I know you resolved this but:

I was thinking that this might have something to do with one of the SCRs not previously in `-added` or `-confirmed`. But I looked, and the reason looks like it is because only 1 of the SCRs is in `mondo.sssom.tsv` Mondo terms in these SCRs being mapped to the same DO term.

`mondo.sssom.tsv`:
```
subject_id	subject_label	predicate_id	object_id	object_label
MONDO:0000302	basidiobolomycosis	skos:exactMatch	DOID:0050278	basidiobolomycosis
```

`mondo-edit.obo`:
```
id: MONDO:0000302
name: basidiobolomycosis
xref: DOID:0050278 {source="MONDO:equivalentTo"}
relationship: excluded_subClassOf MONDO:0000255 {source="DOID:0050278", source="https://orcid.org/0000-0001-5208-3432"} ! subcutaneous mycosis
relationship: excluded_subClassOf MONDO:0005093 {source="DOID:0050278", source="https://orcid.org/0000-0001-5208-3432"} ! skin disorder
```

joeflack4:
@twhetzel 
> For the Subclass Sync, is there not a combined file for "added"?

Strangely, there's not! Nor for `added-obsolete`. And I don't remember why!

```make
sync-subclassof: $(REPORTDIR)/sync-subClassOf.confirmed.tsv $(REPORTDIR)/sync-subClassOf.confirmed-direct-source-indirect-mondo.tsv $(REPORTDIR)/sync-subClassOf.direct-in-mondo-only.tsv $(TMPDIR)/sync-subClassOf.added.self-parentage.tsv
```

My guess is that it's this way because of a some myopia in development. I see that `mondo` has goals for processing confirmed subclass sync, but apparently none (yet) for added subclass sync. Possibly Nico asked explicitly for the current concatenated files, but never asked for a version for plain "added", so I never created it. IDK it's kind of weird. If you want, I can add this "-added" concatenation file as a build output.

src/ontology/lexmatch/README.md
joeflack4: @twhetzel I wasn't thinking creatively enough yesterday. There is a straightforward way to test this.

I can make a new branch, comment out the code for processing both `excluded_subClassOf` and exclusions because of "susceptibility" descendancy. Then I can run a mini build and commit. Then I can uncomment the susceptibility part and run another mini build. Then we can diff the before/after.

Let me know if that's something you want me to do.

joeflack4:
@twhetzel Closing this because I'm merging the approved PR, but I added some more comments for you. If we need to make any tweaks / do any checks, I can do it outside of this branch.

twhetzel:
Thanks for reporting back those findings!

PR #828, opened 2025-04-26
DO NOT MERGE: Temp branch: mapping docs investigation

joeflack4:


twhetzel:
Why is ordo mappable 13,713? That does match the Mapping Progress Report table https://monarch-initiative.github.io/mondo-ingest/reports/unmapped/

Originally I did think that the "mappable" content would be the component minus obsolete terms in the ingest source and minus any terms with an exclusionReason, but Nico suggested to start with the mirror signature instead.

matentzn:
I think in the end, starting from "relevant signature" yields more relevant percentages then mirror - that's what we did for the paper as well.

joeflack4:
@twhetzel I'm going to close this, but feel free to re-open if you want.

PR #830, opened 2025-04-30
DO NOT MERGE - add notebook to review mapping alignment numbers

twhetzel:
Resolves #ISSUE(s).

## Overview

This PR:
- One
- Two
- Three

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [ ] No

- [ ] No

- [ ] Yes


joeflack4 (commented):
I'm surprised, but these numbers actually match quite closely with https://github.com/monarch-initiative/mondo-ingest/blob/main/docs/reports/unmapped.md.

The reason I'm surprised, and what _could_ potentially cause incorrect counts here, is that subtraction of unique non-inclusions aren't calculated.

Unique non-inclusions would be: Set of non-inclusions - Set of exclusions

Non-inclusions being defined as the set of terms that are filtered out when converting the mirror to the component.

@twhetzel My guess is that the reason why these numbers are so close to the mapping progress report is that it may be that 100% (or nearly) of all non-inclusions are also captured by the extensional exclusions.

twhetzel:
Thanks for reviewing this @joeflack4. Yes, it's not entirely clear whether the list of terms in `<ONTOLOGY_NAME>_term_exclusions` also contains non-inclusions based on how the term exclusions file is created and defined.

joeflack4 (commented):
Very helpful! To see:
1.   Mirror Signature [ALL]
2.   Initial Inclusions [IN-SCOPE]
3.   Excluded Terms [EXCLUDED]
4.   Included Terms [DISEASE]

I had assumed that all of (2) was also (4)!

PR #850, opened 2025-07-14
modify nord template to use new preferred_label_for_community

twhetzel:
Resolves #ISSUE(s).

## Overview

This PR:
- Modifies the NORD template to use the new `preferred_label_for_community` property. This PR is intended to be merged into `main`.

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [ ] No

- [x] No

- [ ] Yes


matentzn (approved):
My only slight concern if the source changes and the exact string you are replacing do not exist, the process will not fail; but this is mitigated by the fact that

1. we will get them to fix it at source
2. the table is created automatically so no formatting changes expected.

PR #873, opened 2025-09-08
Main data build 3 sep2025

twhetzel:
Resolves #ISSUE(s).

## Overview

This PR:
- Full data build following the Mondo release (this was run on 3-Sep). However, the new source-versions files were not originally in the newly built files so this PR also includes an update to `.gitignore` to track these files, the source-version files, and a minor code change to swap the file suffix from `.txt` to `.tsv`.

### Documentation
- [ ] Yes

### QC
- [ ] Yes
   
### New Packages
- [ ] No

- [ ] No

- [ ] Yes


src/ontology/external/mondo-omim-genes.robot.tsv
matentzn, line 1: A bit weird there are only 474 additions since we protected more than 600

twhetzel, line 1: Large addition due to recent addition of entries into the "protected" file in the omim repo. NOTE: That PR does have some duplicates, e.g see [here](https://github.com/monarch-initiative/omim/pull/276/files#diff-261066b07106a3db627bd3a89f5a2a068e3f5af7542e72d04c20bdc978c2b60eR98-R99), so the PR title of 600+ new protected associations may actually be less and therefore why only 474 new associations are found in this mondo-ingest file.

PR #904, opened 2026-06-09
2026-06-09 data release

ptgolden:
A number of new terms from DOID.

Some churn in OMIM-MONDO mappings, but they all match up with curatorial work done in Mondo.

I don't notice anything else much out of the ordinary.


matentzn (approved):
I think this looks good but please follow up on the NORD EMC stuff.. This is important

PR #906, opened 2026-06-26
Icd10who external release

Reasat:
Partially addresses external-release pattern for ICD10WHO (component prep moved to source repo; mondo-ingest consumes releases).

## Overview

This PR switches ICD10WHO from an in-repo ROBOT component build to **downloaded release artifacts** from the external icd10who source repo, while keeping mirror-based QC in mondo-ingest.

This PR:
- Downloads `components/icd10who.owl` from the icd10who GitHub release (`ICD10WHO_RELEASE`) instead of building it in mondo-ingest; removes `icd10who-relevant-signature.sparql` (no longer needed)
- Downloads `icd10who.raw.ttl` for mirror input (`ICD10WHO_UPSTREAM`), then builds `tmp/mirror-icd10who.owl` for exclusions and signature reports
- Converts release OWL from LinkML functional syntax to RDF/XML after `wget` so `semsql` / `rdftab` can build `components/icd10who.db` (fixes CI failure on fresh release downloads)
- Adds a D2 high-level pipeline diagram under `docs/` (`icd10who-high-level-flow.d2`, `.png`)
- Adds **ICD10WHO integration check** GitHub Actions workflow: runs `make slurp/icd10who.tsv` in `odkfull:v1.6` and uploads report TSVs as artifacts

**Release URLs** currently point at the external icd10who fork releases (`icd10who.owl`, `icd10who.raw.ttl`). Happy to switch to `monarch-initiative/icd10who` before merge once those assets are published there.

**Integration test (CI):** [ICD10WHO integration check — success](https://github.com/Reasat/mondo-ingest/actions/runs/28216988896)

Download **Artifacts** → `icd10who-integration-<sha>` for: `icd10who_mapping_status.tsv`, `icd10who_unmapped_terms.tsv`, `icd10who_term_exclusions.txt`, `icd10who_exclusion_reasons.robot.template.tsv`, component/mirror signatures, `slurp/icd10who.tsv`.

**Design note:** Mirror remains in mondo-ingest intentionally for exclusions/signatures (component vs upstream diff), consistent with other sources. Mirror is built from released raw TTL today; source repo also publishes `icd10who.mirror.ttl` if we want to drop the redundant normalize step later.

### Documentation

- [x] Yes — D2 pipeline diagram (`docs/icd10who-high-level-flow.d2`, `.png`)

*(No new `make` goals; `docs/developer/workflows.md` unchanged.)*

### QC

Was the full pipeline run before submitting this PR using `sh run.sh make build-mondo-ingest` on this branch (after `docker pull obolibrary/odkfull:dev`), and no errors occurred?
- [ ] Yes
- [x] No — full multi-source `build-mondo-ingest` was **not** run (hours-long). **ICD10WHO integration check** CI completed successfully on this branch ([run link above](https://github.com/Reasat/mondo-ingest/actions/runs/28216988896)); artifact zip contains mapping/unmapped/exclusion/slurp outputs for reviewer inspection.

### New Packages

- [x] No

- [x] No

- [ ] Yes


src/ontology/mondo-ingest.Makefile
```diff
@@ -18,7 +18,8 @@
 DOID=				http://purl.obolibrary.org/obo/doid.owl
 ICD10_BP_CODE=		27
 ICD10CM=			https://data.bioontology.org/ontologies/ICD10CM/submissions/$(ICD10_BP_CODE)/download?apikey=8b5b7825-538d-40e0-9e9e-5ab9274a9aeb
-ICD10WHO=			https://github.com/monarch-initiative/icd10who/releases/latest/download/icd10who.ttl
+ICD10WHO_UPSTREAM=	https://github.com/Reasat/icd10who/releases/latest/download/icd10who.raw.ttl
```
matentzn: The fact this is necessary is a bit of a stop gap - I would prefer `ICD10WHO_UPSTREAM` was not needed in Mondo Ingest at all. But this can be done as a separate step

.github/workflows/icd10who-integration.yml
matentzn, line 1: This action we have not discussed it yet - I am a bit concerned that because some pipeline steps are dependent on ALL ingests at once, that isolating alignment processes this way might not be really possible.. Did you test this?

matentzn (commented):
@Reasat this is amazing, thanks! I would suggest we dont worry about the integration action just yet - for now just delete as much code from mondo-ingest.Makefile and files from the repo as you can.

🎉 🎉

Reasat:
@matentzn Following up on your review, I have this updated target design for the source repo and the mondo-ingest. 

[design diagram](https://github.com/Reasat/mondo-ingest/blob/icd10who-external-release/docs/mondo-pipeline-target.png)

You might need to download the image for zoom in/ zoom out etc.

This PR is revised to match the updated target design.

### Changes since `d01b13da`

1. **Removed** `.github/workflows/icd10who-integration.yml` (Not useful at this stage).
2. **Removed `ICD10WHO_UPSTREAM`** and all in-repo WHO preprocessing (mirror from raw TTL, local semsql, SPARQL signatures, sssom parse).
3. **mondo-ingest now wget's the full icd10who release bundle** from `ICD10WHO_RELEASE_BASE`:
   - `icd10who.owl` → `components/icd10who.owl` (+ ROBOT convert for semsql)
   - `icd10who.mirror.owl`, `icd10who.db`
   - `mirror_signature.tsv`, `component_signature.tsv`, `icd10who.sssom.tsv`, `icd10who-metrics.json`
   - GitHub Releases flattens paths (`reports/foo.tsv` → `foo.tsv` at download URL)
   - **Note: ICD10WHO_RELEASE_BASE currently points to [Reasat/icd10who](https://github.com/Reasat/icd10who). We need to move it under the appropriate org when convenient.**
4. **Added** `docs/icd10who-artifacts.md`, updated current and target flow diagram. but these are kind of optional tutorial/reference for a new contributor (like me), can be omitted for now.

### Partial integration test 

To avoid a full multi-hour `build-mondo-ingest`, I ran a **partial rebuild**:

- **6 components + mappings** frozen from official mondo-ingest release [`v2026-01-24`](https://github.com/monarch-initiative/mondo-ingest/releases/tag/v2026-01-24)
- **WHO only** from icd10who release [`v2026-07-05`](https://github.com/Reasat/icd10who/releases/tag/v2026-07-05)
- Then: `make exclusions-icd10who slurp/icd10who.tsv prepare_release_fast` in `odkfull:v1.6`

| Metric | Result |
|--------|--------|
| Runtime | ~35 min |
|WHO class count | 12,597 (official component: 12,543 — +54 classes) |
|WHO axiom count| 37,769 (official: 63,010), each term carries fewer redundant annotation triples? |
| `slurp/icd10who.tsv` | 984 proposals |
| Mapping status | 12,597 terms — 209 mapped (unchanged), 12,388 unmapped (not disease related probabaly)|
| `mondo-ingest.owl` | 573.9 MB vs official 575.0 MB (~1 MB smaller) |
| `icd10who.owl` | ~10.5 MB vs official ~12 MB (1.5 MB smaller) |

Code to replicate test

```
cd /path/to/mondo-ingest

docker run --rm \
  -v "$PWD":/work \
  -w /work/src/ontology \
  obolibrary/odkfull:v1.6 \
  bash -lc '
set -e
MI=https://github.com/monarch-initiative/mondo-ingest/releases/download/v2026-01-24
WHO=https://github.com/Reasat/icd10who/releases/download/v2026-07-05
mkdir -p components ../mappings reports tmp metadata

echo "=== 6 official components (frozen) ==="
for s in doid omim ordo ncit icd10cm icd11foundation; do
  wget -nv "$MI/$s.owl" -O "components/$s.owl"
  wget -nv "$MI/$s.sssom.tsv" -O "../mappings/$s.sssom.tsv"
done

echo "=== WHO release bundle (new source repo) ==="
wget -nv "$WHO/icd10who.mirror.owl" -O tmp/mirror-icd10who.owl
wget -nv "$WHO/icd10who.db" -O components/icd10who.db
wget -nv "$WHO/mirror_signature.tsv" -O reports/mirror_signature-icd10who.tsv
wget -nv "$WHO/component_signature.tsv" -O reports/component_signature-icd10who.tsv
wget -nv "$WHO/icd10who.sssom.tsv" -O ../mappings/icd10who.sssom.tsv
wget -nv "$WHO/icd10who-metrics.json" -O metadata/icd10who-metrics.json
wget -nv "$WHO/icd10who.owl" -O tmp/icd10who-release.owl.tmp
robot convert -i tmp/icd10who-release.owl.tmp -o components/icd10who.owl
rm -f tmp/icd10who-release.owl.tmp

echo "=== WHO alignment + release merge ==="
make exclusions-icd10who COMP=true MIR=false
make slurp/icd10who.tsv COMP=true MIR=false
make prepare_release_fast

echo "=== results ==="
ls -lh ../../mondo-ingest.owl ../../mondo-ingest.obo
wc -l slurp/icd10who.tsv reports/icd10who_mapping_status.tsv
```
Full `build-mondo-ingest` not run (i.e. build 6 individual sources from scratch using the mondo-ingest current method and wget the icd10who from the new source repo). Happy to run before merge if required.

### Questions:
- Any feedback on the current pipeline design? 
- How would you like to do further testing?

src/ontology/mondo-ingest.Makefile
```diff
@@ -207,18 +207,37 @@ $(COMPONENTSDIR)/icd10cm.owl: $(TMPDIR)/icd10cm_relevant_signature.txt $(TMPDIR)
...
-			--update ../sparql/fix-labels-with-brackets.ru \
-			--update ../sparql/exact_syn_from_label.ru \
-		remove -T config/properties.txt --select complement --select properties --trim true \
-		annotate --ontology-iri $(URIBASE)/mondo/sources/icd10who.owl --version-iri $(URIBASE)/mondo/sources/$(TODAY)/icd10who.owl -o $@; fi
+####################################
+### ICD10WHO external release ######
+####################################
+# Preprocessing lives in the icd10who source repo; mondo-ingest wget's the release bundle.
+
+# linkml-owl releases functional-syntax OWL; convert to RDF/XML for semsql rdftab.
```
ptgolden: If we want to move as much processing as possible into a self-contained ICD10WHO release, why not just run this conversion there?

src/ontology/mondo-ingest.Makefile
```diff
@@ -18,7 +18,12 @@
 DOID=				http://purl.obolibrary.org/obo/doid.owl
 ICD10_BP_CODE=		27
 ICD10CM=			https://data.bioontology.org/ontologies/ICD10CM/submissions/$(ICD10_BP_CODE)/download?apikey=8b5b7825-538d-40e0-9e9e-5ab9274a9aeb
-ICD10WHO=			https://github.com/monarch-initiative/icd10who/releases/latest/download/icd10who.ttl
+ICD10WHO_RELEASE_BASE=	https://github.com/Reasat/icd10who/releases/latest/download
```
ptgolden, line 21: I think it would be better to just define one variable up here (`ICDWHO_RELEASE_BASE` is fine) and then refer to the files later on explicitly rather than with variables. Like this:

```
reports/mirror_signature-icd10who.tsv: | $(REPORTDIR)
	wget -nv $(ICD10WHO_RELEASE_BASE)/mirror_signature.tsv -O $@
```

Just defining the base up here makes it easier to follow later on what is actually being downloaded, rather than having to track what variable names represent.

ptgolden (changes requested):
One thing first. All the documentation you've put in here is fantastic, but I think it should be left out of this PR. It would put in a lot of documentation of a particular style that would only exist for ICD10WHO. I would save all this, and we can integrate it in a further PR where we can holistically think out how external sources should be documented

---

Second, I think the changes to your Makefile-- which just downloads files from an external repo's release-- could be extremely simple. If you were to update the external repo so that its release's filenames were exactly what mondo-ingest expects, the recipe could just be this:

```
ICD10WHO_DOWNLOADS := $(REPORTDIR)/mirror_signature-icd10who.tsv \
		      $(REPORTDIR)/component_signature-icd10who.tsv \
		      $(MAPPINGSDIR)/icd10who.sssom.tsv \
		      metadata/icd10who-metrics.json

ifeq ($(COMP),true)
ICD10WHO_DOWNLOADS += $(COMPONENTSDIR)/icd10who.owl \
		      $(COMPONENTSDIR)/icd10who.db \
		      $(TMPDIR)/mirror-icd10who.owl
endif

$(ICD10WHO_DOWNLOADS): | $(REPORTDIR) $(COMPONENTSDIR) $(TMPDIR) $(MAPPINGSDIR)/
	wget -nv $(ICD10WHO_RELEASE_BASE)/$(@F) -O $@.tmp && mv $@.tmp $@
.PRECIOUS: $(TMPDIR)/mirror-icd10who.owl
```

(`$(@F)` is "The file-within-directory part of the file name of the target. If the value of ‘$@’ is dir/foo.o then ‘$(@F)’ is foo.o. ‘$(@F)’ is equivalent to ‘$(notdir $@)’. " <https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html>)

(I also added the `wget` .tmp dance to prevent partial downloads that might break things.)

This also provides a template for other sources to externalize in the same manner-- just publish files with those names, and download all those files to the place where `mondo-ingest` expects them. (This could even be turned into a `define`/`eval` line, like:

```
$(eval $(call add_external_source,icd10who,$(ICD10WHO_RELEASE_BASE)))
$(eval $(call add_external_source,icd11foundation,$(ICD11FOUNDATION_RELEASE_BASE)))
$(eval $(call add_external_source,omim,$(OMIM_RELEASE_BASE)))
```

...etc.

---

There are two issues with this approach that I want to point out.

1. Using `releases/latest` introduces a race condition: a new release might be published in the middle of `mondo-ingest` running if it downloads multiple files in a row. The chances of this are slim, but I wanted to point it out. The alternative would be to explicitly set release tags e.g. `v2026-07-30`. This could be annoying, but it could also be automated.

2. Not building things from source and instead `wget`ing them prevents incremental updates. If I were to run `make reports/icd10who_mapping_status.tsv` from a fresh clone, everything would work as planned. If I were to run it again a month later, as things stand now, `make` would tell me everything is up to date because all of the dependencies are already up to date. To actually make it run, you would need to use `make --always-make` (aka `-B`).

Again, this is fine in the vision we have where all of this is just run in a fresh checkout via GitHub Actions, but it's worth noting.

src/ontology/mondo-ingest.Makefile
Reasat: Sounds good, I updated the relevant lines below so that they read files relative to the base

src/ontology/mondo-ingest.Makefile
Reasat: I agree. I have pushed this to the upstream repo. The released icd10who.owl is now RDF/XML (ROBOT convert runs in the icd10who source repo before publish). We can do a wget below to get processed files.

Reasat:
**Docs**
Agreed they shouldn’t live only for ICD10WHO in this PR. Removed from this branch; parked on docs/icd10who-external-source-docs until this PR gets mature.

**Makefile / matching release filenames**

Done. The icd10who release now ships basenames that match what mondo-ingest expects (mirror-icd10who.owl, mirror_signature-icd10who.tsv, component_signature-icd10who.tsv, etc.), and this PR uses your 
ICD10WHO_DOWNLOADS + $(@F) + .tmp && mv pattern:
```
ICD10WHO_DOWNLOADS := $(REPORTDIR)/mirror_signature-icd10who.tsv \
		      $(REPORTDIR)/component_signature-icd10who.tsv \
		      $(MAPPINGSDIR)/icd10who.sssom.tsv \
		      metadata/icd10who-metrics.json
ifeq ($(COMP),true)
ICD10WHO_DOWNLOADS += $(COMPONENTSDIR)/icd10who.owl \
		      $(COMPONENTSDIR)/icd10who.db \
		      $(TMPDIR)/mirror-icd10who.owl
endif
$(ICD10WHO_DOWNLOADS): | $(REPORTDIR) $(COMPONENTSDIR) $(TMPDIR) $(MAPPINGSDIR)/
	wget -nv $(ICD10WHO_RELEASE_BASE)/$(@F) -O $@.tmp && mv $@.tmp $@
```
Latest WHO bundle with those names: [icd10who v2026-08-07.1](https://github.com/Reasat/icd10who/releases/tag/v2026-08-07.1).

Didn’t add add_external_source / eval yet. We can address that as the templates gets a bit mature through this ICD10WHO exploration.

**releases/latest race**
Noted. Leaving …/releases/latest/download for now. If we want to close the race without manual pin bumps, a simple approach is to resolve the latest tag once at the start of a CI/make run and reuse that base URL for every wget. I could implement that is that sounds good.

**No incremental refresh after wget**
Also noted. With wget’d artifacts, make won’t re-fetch when a new release appears upstream unless you use a fresh checkout, make -B, or delete the local files. That matches the “fresh checkout via GitHub Actions” workflow; we can document a local refresh target later if useful.

(I have made some replies to the related inline comments: OWL convert now runs in the icd10who source repo; released icd10who.owl is RDF/XML so mondo-ingest only wgets.)

ptgolden (approved):
This looks great. Last question-- did you test it out yourself?

Also, this is maybe a question for @matentzn, but should we move the external repo into the `monarch-initiative` organization before merging this? (I could see arguments either way).

I think the documentation additions you made are great. If and when we generalize this pattern to other external sources, we can remove that note, and list the expectations for any external source repo.

ptgolden:
>  Noted. Leaving …/releases/latest/download for now. If we want to close the race without manual pin bumps, a simple approach is resolve the latest tag once at the start of a CI/make run and reuse that base URL for every wget. I could implement that is that sounds good.

That's what I was thinking too. Let's leave out resolving tags for now, and it's something else we'll explore as we expand this pattern to other sources.

What is the release cadence for your repository? I didn't knock on wood when I said that it would be a longshot to have that race condition, so Murphy's Law says that it will happen the next time I run a data release.

Reasat:
@ptgolden Thanks for the review!
I ran the test `make slurp-icd10who` in `odkfull:v1.6` (made sure the wget artifacts and target outputs were deleted before running)
```
cd <path>/mondo-ingest/src/ontology
docker run --rm -i \
  -v /workspace/Projects/mondo-ingest:/work \
  -w /work/src/ontology \
  -e ROBOT_JAVA_ARGS="-Xmx20G" \
  -e JAVA_OPTS="-Xmx20G" \
  -e ODK_USER_ID="$(id -u)" \
  -e ODK_GROUP_ID="$(id -g)" \
  obolibrary/odkfull:v1.6 \
  make slurp-icd10who
```
slurp/icd10who.tsv has 984 proposals
reports/icd10who_mapping_status.tsv has 12,597 terms (209 mapped, 0 excluded)
reports/icd10who_unmapped_terms.tsv has 12,388 rows

Let me know if these number look good or should we be a bit more rigorous and check against some known ground truth. (Whatever you guys feel comfortable with)

For the who source repo, the who data seems to be frozen since 2019? So, I am not expecting new releases for this particular source. But the other sources might have some regular intervals. Looking into it.

I'll leave the repo moving question for @matentzn !

Reasat:
Additionally, I was wondering, how do you guys like to organize docs? 

https://github.com/monarch-initiative/mondo-ingest/tree/main/docs
I saw this, I understood them partially, but some of the folders/files were empty.

matentzn:
Additionally, I was wondering, how do you guys like to organize docs?

https://github.com/monarch-initiative/mondo-ingest/tree/main/docs

Stick them in there for now, and add doc generation targets to mondo-ingest.Makefile for dynamic content. I will make a concerted issue for the docs later this year.

ptgolden:
@matentzn this looks good to me to merge, but I want your opinion on whether we should move the external repo into `monarch-initiative` first.

matentzn:
> we should move the external repo into monarch-initiative first.

Yes @Reasat can you move any repos to this org? I will give you access back immediately.

Reasat:
monarch-initiative/icd10who already exists. 
- Should I push my repo as a branch there, then we can make it default and archive other branches?
- Other option is to rename my repo and try the transfer; (or rename monarch-initiative/icd10who --> monarch-initiative/icd10who-legacy)
@matentzn

matentzn:
Renamed: https://github.com/monarch-initiative/icd10who-legacy

Reasat:
Thanks, but the new error is saying I can't create private/public repos in the org. I transferred ownership to you (@matentzn). Maybe you can place it in the org? (Let me know if you didn't receive a transfer request email)

PR #909, opened 2026-07-14
EMC post process script fixes

ptgolden:
A few small fixes to the EMC post-processing script. Two bugs were causing the pipeline to error out (fixed by 65de60f and c92b889), one addresses a Pandas deprecation warning, and one adds a new QC entry (reporting the literal value `MONDO:`).


matentzn (approved):
- There's still an `applymap` in `src/scripts/utils.py:319`. If the point of the deprecation fix is to get ahead of pandas removing it (3.0) then...you know
- The `^MONDO:$` (and the existing `^MONDO:MONDO:` ) check runs `.map` across *every* column and drops the row if *any* cell matches. In practice only the first/ID column should ever hold a bare `MONDO:` - just FYI

PR #910, opened 2026-07-20
2026-07-20 data release

ptgolden:


ptgolden:
@matentzn, I've been trying to figure it out but can't-- what do you think accounts for the order-switching of capitalizations in the synced synonyms? They don't seem to be semantically meaningful differences, so I don't think it should prohibit merging. Just wondering.

ptgolden:
I think I've nailed it down to some non-determinism for the order of multi-valued fields. I'm going to fix them tomorrow. But-- there's no semantic difference, so this is fine to merge.

matentzn:
Yep, not concerned, you go ahead

matentzn (approved):

PR #915, opened 2026-08-18
In synonym sync scripts, give multi-valued cells a deterministic order

ptgolden:
There was a good deal of churn in data releases caused solely by non-meaningful re-ordering of multi-valued cells in TSV that were generated from SPARQL and Pandas groupings, neither of which are guaranteed to be identical across runs. This commit explicitly sorts those values.

Fixes the issue raised in this comment: <https://github.com/monarch-initiative/mondo-ingest/pull/910#issuecomment-5028486391>


matentzn (approved):
Nice!

PR #918, opened 2026-09-15
2026-09-15 data release

ptgolden:
This is the first data release since #906.

@matentzn there's a huge change in `unmapped/icd10who-unmapped.owl`, which I imagine is a product of it now being produced by the LinkML pipeline. All of the classes there now no longer contain:

* `rdfs:seeAlso` assertions
* `<oboInOwl:hasSynonymType rdf:resource="http://purl.obolibrary.org/obo/mondo#GENERATED_FROM_LABEL"/>` annotations on synonyms

Additionally, the `<owl:Ontology>` header no longer has an associated URI.

Should these be produced upstream? (Besides those issues, the data produced looks good. Thanks, @Reasat!)


matentzn (approved):
Fine with this will file some follow up issues.

